PR #203, opened 2015-02-24
Support packed tests

dffischer:
Some developers may write their tests in the docstrip format, for example to unify common setup code for multiple tests. ([You may look at a simple exemplary use case here.](https://github.com/dffischer/l3build-testingtests)) 

`l3build.lua` does currently support this quite poorly. `testfiledir` can be set to `unoackdir` to make generated `lvt` files be found, but then `build.lua save` stores the `tlg`s into `unpackdir` where they are soon to be deleted.

This patch looks for `lvt`s in `unpackdir`, too, but still stores `tlg`s into `testfiledir`, where they persist.

I read that you do not apply patches directly at github. If you favor to receive this patch in any other way, please leave a comment.

### DIFF
--- a/l3build/l3build.dtx
+++ b/l3build/l3build.dtx
@@ -604,6 +604,10 @@
 %   \item Correction of a \LuaTeX{} error message typo (|I''m going to assume|).
 % \end{itemize}
 %
+% \subsection{Generating test files with docstrip}
+%
+% It is also possible to pack tests and goals in \texttt{.dtx} files. If a test specification (\texttt{.lvt}) or goal (\texttt{.ltg}) shows up during the usual unpacking process that starts most of the command, it will be considered with the \texttt{check} and \texttt{save} commands as if it was found in the \texttt{testfiledir}. Be aware that this means that a test goal there may shadow a generated one of the same name. When performing test related tasks, the \texttt{testfiledir} will also be searched for \texttt{sourcefiles} to unpack. These will not be relevant to other commands, especially they will not show up in the archives created for the \texttt{ctan}.
+%
 % \section{Writing test files}
 % \label{sec:writing-tests}
 %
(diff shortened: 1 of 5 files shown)


josephwright:
I'd imagine someone packing their `.lvt` files into a source might also pack the `.tlg` files. Should that be supported too, at least from the 'where to look for results' part? (Clearly the 'save the results' part is going to be limited in what in can do under such circumstances.)

josephwright:
Aside:  Pull requests don't really 'work' here because the Git repo is only a mirror of the 'real' one (Subversion). That means it's more-or-less read-only.

dffischer:
As pull requests cannot be merged directly, can you direct me to a more appropriate way to submit patches?

Packing `.tlg` files is indeed a sensible idea, so I incorporated it. The "save" command can still be used, but it shows a warning when masking a generated file.

dffischer:
The recent commits also look for `sourcefiles` to unpack in `testdir`, but only for the commands dealing with tests. This keeps things neatly assorted. I also added documentation and a simple example to the `testfiles` directory showing how to use packed tests.  The latter have dashes in their name and will work with patch #205.

josephwright:
I'm working through the various changes: one or two seem to cause issues at my end and I want to be sure I understand them all! Will check something in over the next day or so.

josephwright:
I've integrated the idea so will close (hope this the right way to handle pull requests!).

josephwright:
@dffischer BTW, thanks very much for this and other suggestions :-)

dffischer:
> hope this the right way to handle pull requests

The ideal way would be to mark it as _merged_ instead of _closed_, but I don't know whether this is possible in a mirror.

josephwright:
@dffischer  As I say, we can't merge (it would alter the GitHub version of the mirror but thus be out-of-sync with the SVN)

PR #204, opened 2015-02-26
remove line numbers when \showing boxes

dffischer:
The manual suggests checking typeset output by placing it into a box and then `\showbox` its content. The output will then include a number denoting at which line of the `lvt` source the command was issued. This could mess up comparison and should be removed analogous to `on line ...` is. The attached patch fixes this by replacing the line number with `...` like it is already done with other numbers.


josephwright:
There was some reason that in the current `expl3` tests we've gone with `\tl_show:n`, _etc._, rather than a 'raw' `\show`. I'll double-check before making this change: may just have been to do with word-wrapping.

dffischer:
I don't think line numbers are the reason to use expl3 commands. They also include the line numbers. Pages 14 and 15 of the l3build documentation show this. But I updated the branch to catch them as well.

PR #205, opened 2015-02-28
handle hyphenated test file names

dffischer:
When trying to execute the tests for l3build, I noticed that test file names with dashes cause errors. This is because they have to pass through lua's `string.gsub` as a pattern where the dash is a special symbol. Escaping it solves this problem.


dffischer:

PR #206, opened 2015-03-11
support expectation tests

dffischer:
l3build is focused around regression tests, i.e. evaluating test output by comparing with a previous output defined as correct. Especially when testing typeset output, boxes or other complex matter, the test goals are not intuitive to read. It can often be easier to write and comprehend expectation tests, that is specifying test content and expected result in the same format. The patches proposed herein make it possible to alternatively specify expected results in the same format as tests. If such an expectation specification (`.lve`) is present, the `.tlg` is generated from it. The test passes when both `.lvt` and `.lve` lead to the same output, after processing by the usual log formatting.

You can look at an example in a [further branch](https://github.com/dffischer/svn-mirror/blob/expect-packed/l3build/testfiles/01-expect.dtx) showing test and expectation side-by-side in a `dtx` file, which provides a clean format to use thanks to b6b38b0412f0d96283db83e78203da77aa18640d.

### DIFF
--- a/l3build/l3build.dtx
+++ b/l3build/l3build.dtx
@@ -112,6 +112,7 @@
 \luavarset{lvtext} {".lvt"} {Extension of test files.}
 \luavarset{tlgext} {".tlg"} {Extension of test file output.}
 \luavarset{logext} {".log"} {Extension of checking output, before processing it into a \texttt{.tlg}.}
+\luavarset{lveext} {".lve"} {Extension of expextation files. Missing \texttt{.tlg} files will be generated from them.}
 \luavarseparator
 \luavarset{checkdeps}   {\{~\}} {List of build unpack dependencies for checking.}
 \luavarset{typesetdeps} {\{~\}} {\dots for typesetting docs.}
@@ -806,6 +807,10 @@
 %   \caption{Output from displaying the contents of a simple box to the log file, using plain \TeX{} (left) and \pkg{expl3} (right). Some blank lines have been added to the plain \TeX{} version to help with the comparison.}
 %   \label{fig:box-log}
 % \end{figure}
+%
+% \subsection{Specifying expectations}
+%
+% Regression tests check whether changes introduced in the code modify the test output. Especially while developing a complex package there is not yet a baseline to save a test goal with. It might then be easier to formulate the expected effects and outputs of tests directly. To achieve this, you may create an \texttt{.lve} instead of a \texttt{.tlg} file.\footnote{Mnemonic: \texttt{lv\textbf{t}}: \textbf{t}est, \texttt{lv\textbf{e}}: \textbf{e}xpectation} It is processed exactly like the \texttt{.lvt} to generate the expected outcome. The test fails when both differ.
 %
 %
 %
--- a/l3build/l3build.dtx
+++ b/l3build/l3build.dtx
@@ -605,14 +605,6 @@
 %   \item Correction of a \LuaTeX{} error message typo (|I''m going to assume|).
 % \end{itemize}
 %
-% \subsection{Generating test files with \pkkg{DocStrip}}
-%
-% It is  possible to pack tests inside source files. Tests generated during the
-% unpacking process will be available to the \texttt{check} and \texttt{save}
-% commands as if they were stored in the \texttt{testfiledir}. Any explicit
-% test files inside \texttt{testfiledir} take priority over generated ones
-% with the same names.
-%
 % \section{Writing test files}
 % \label{sec:writing-tests}
 %
@@ -807,6 +799,15 @@
 %   \caption{Output from displaying the contents of a simple box to the log file, using plain \TeX{} (left) and \pkg{expl3} (right). Some blank lines have been added to the plain \TeX{} version to help with the comparison.}
 %   \label{fig:box-log}
 % \end{figure}
+%
+% \section{Alternative test formats}
+% \subsection{Generating test files with \pkkg{DocStrip}}
+%
+% It is  possible to pack tests inside source files. Tests generated during the
+% unpacking process will be available to the \texttt{check} and \texttt{save}
+% commands as if they were stored in the \texttt{testfiledir}. Any explicit
+% test files inside \texttt{testfiledir} take priority over generated ones
+% with the same names.
 %
 % \subsection{Specifying expectations}
 %
--- a/l3build/l3build.dtx
+++ b/l3build/l3build.dtx
@@ -813,6 +813,39 @@
 %
 % Regression tests check whether changes introduced in the code modify the test output. Especially while developing a complex package there is not yet a baseline to save a test goal with. It might then be easier to formulate the expected effects and outputs of tests directly. To achieve this, you may create an \texttt{.lve} instead of a \texttt{.tlg} file.\footnote{Mnemonic: \texttt{lv\textbf{t}}: \textbf{t}est, \texttt{lv\textbf{e}}: \textbf{e}xpectation} It is processed exactly like the \texttt{.lvt} to generate the expected outcome. The test fails when both differ.
 %
+% Combining both features enables contrasting the test with its expected outcome in a compact format. Listing \ref{fig:expect-dtx} exemplary tests \TeX{}s counters. Listing \ref{fig:expect-ins} shows the relevant part of an \texttt{.ins} file to generate it.
+%
+% \begin{figure}
+%   \begin{lstlisting}[frame=single,language={TeX},gobble = 6]
+%     \input regression-test.tex\relax
+%     \START
+%     \TEST{counter-math}{
+%     %<*test>
+%       \OMIT
+%       \newcounter{numbers}
+%       \setcounter{numbers}{2}
+%       \addtocounter{numbers}{2}
+%       \stepcounter{numbers}
+%       \TIMO
+%       \typeout{\arabic{numbers}}
+%     %</test>
+%     %<expect>  \typeout{5}
+%     }
+%     \END
+%   \end{lstlisting}
+%   \caption{Test and expectation can be specified side-by-side in a single \texttt{.dtx} file.}
+%   \label{fig:expect-dtx}
+% \end{figure}
+%
+%\begin{figure}
+%   \begin{lstlisting}[frame=single,language={TeX},gobble = 6]
+%      \generate{\file{\jobname.lvt}{\from{\jobname.dtx}{test}}
+%                \file{\jobname.lve}{\from{\jobname.dtx}{expect}}}
+%   \end{lstlisting}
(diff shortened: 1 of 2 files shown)


josephwright:
I must be missing something fundamental here: the `\tlg` file _is_ what you expect to get from running a test.

josephwright:
BTW, for testing typesetting the workflow can only really be to run a test by hand, check the visual appearance of the PDF then save the resulting log, as you can't be sure otherwise
 that things 'look right'/

josephwright:
I'm going to raise this with the team as a whole: it's a relatively significant change and I'd like to get a broader feeling before doing anything.

dffischer:
I am aware that this change introduces a whole new variant of testing and thus is quite extensive, so I will try to provide you some more information to aid the decision. Here are the needs that led me to implementing this proposal.

### Compactness
l3build is the only system I found to be sufficiently capable of automating tests for typeset output. PDFs cannot even be `diff`ed directly. Graphical comparisons like ImageMagicks `compare` may introduce errors through rasterization. l3build does quite a good job enabling such tests by utilizing `\tracingoutput=1` or `\showbox`, as denoted in the manual.
But storing the output of such tests lead to large `.tlg`, impossible to write by hand. Storing a simple `.tex` / `.lve` file that creates them not only takes up less space, but also makes it easier to understand the test intention.


### Model of thought
Often, it is more intuitive to define an expected outcome in terms of `TeX` code.

Imagine for example, I want to test the basic LaTeX `\include` mechanism and ensure that a document composed from multiple files produces the exact same result that a single document containing all the content would, leaving sections intact. It would be convenient to specify a test that directly compares

```LaTeX
\begin{document}
\include{file1}
\include{file2}
\section{main file content}
\end{content}
```

to 

```LaTeX
\section{contents of file1}
\section{contents of file2}
\section{main file content}
```

With the current state of l3build, one would have to resort to:
1. Write the expectation with all content in an `.lvt` file.
2. Run it through `texlua build.lua save`.
3. Modify the `.lvt` to use `\include`.
4. Compare this and future versions using `texlua build.lua check`.

Especially when the source code is under version control, the question arises whether to put the expectation specification (Step 1) into the history just to be overwritten right after.

Also, Step 1 can be done before the implementation of the tested feature, which can aid sort of a test-driven development process.


### Portability.
This is more of a side effect. Now suppose that the aforementioned tests are run on another machine that has some weird version of LaTeX that typesets `\section` two points larger. When specified as a `.tlg`, the tests fail. A test specified as `.lve` and `.lvt` will still succeed, which I think is totally fine, as `\section` is expected to be left untouched. A change to `\include` on the other hand would make the test fail as it is intended.


I hope the example clarifies things a bit.

wspr:
Interesting idea. I can see some minor benefits as well for cases where you might want to write, say, a plain TeX or vanilla LaTeX2e implementation and check against that. That’s particularly relevant, say, when you’re not entirely sure what the box log output is supposed to look like. I’m not sure how many times this situation would actually be relevant, though.

On the other hand, I do think we want quite a compelling use case before adding new features, and as Joseph says there is a fair bit you can do with “just” using the `tlg` file. I’d be interested to hear Frank’s views, especially.

davidcarlisle:
On 14 March 2015 at 07:08, Will Robertson <notifications@github.com> wrote:

> Interesting idea. I can see some minor benefits as well for cases where
> you might want to write, say, a plain TeX or vanilla LaTeX2e implementation
> and check against that. That’s particularly relevant, say, when you’re not
> entirely sure what the box log output is supposed to look like. I’m not
> sure how many times this situation would actually be relevant, though.
>
> On the other hand, I do think we want quite a compelling use case before
> adding new features, and as Joseph says there is a fair bit you can do with
> “just” using the `tlg` file. I’d be interested to hear Frank’s views,
> especially.
>
> —
>


As I commented previously to Joseph in sx.chat, this is very similar to the
"reftest" system
used to test javascript/css/html in browsers

https://developer.mozilla.org/en/docs/Creating_reftest-based_unit_tests

A difference here though is in the browser there is no analogue of
\showoutput to give
a representation of the rendering tree, so comparing the transient rendered
results of
two input files is often the only way.

Here, since the result comparison is still going via the normalised tlg log
it's not
so different from the existing tests. I think in this context it would be
more natural
to consider (and implement) as an extension of the "documented tlg"
mechanism just added
rather than generate the tlg from a dtx file essentially you are generating
it by running
the tlg generator over a different input file, but otherwise the test
mechanism is the same.


So on balance I'm not totally against the idea....

David

dffischer:
First and foremost, thanks for your consideration.

> consider (and implement) as an extension of the "documented tlg" mechanism just added

> I do think we want quite a compelling use case before adding new features

I assume that "documented tlg" means a `.tlg` generated from `.dtx` format. The two features actually play together quite nice. I will try to clarfy this with the use case I am currently struggling with. It isn't fetched too far from the example in my last post. I am currently developing a package that adjusts the sectioning levels. It would for example provide an environment `sectup` that turns `\subsection` into `\section`, `\subsubsection` into `\subsection`, `\paragraph` into `\subsubsection`, and so on. So I would like to test whether

```LaTeX
\begin{sectup}
  \begin{sectup}
    \subsubsection{SubSubSection is two levels below Section}
  \end{sectup}
\end{sectup}
```

is equivalent to

```LaTeX
\section{SubSubSection is two levels below Section}
```

I see no possibility to compare both commands directly with something like `\ifx` because the environment redefines all sectioning commands at once, where they become macros possibly relying on conditionals and counters themselves.

Utilizing a `.tlg`, even when packed into a `.dtx`, the complete output of the sectioning command has to be included, which would be.

```LaTeX
Completed box being shipped out [1]
\vbox(655.63496+0.0)x435.61176
.\glue -25.16531
.\vbox(680.80028+0.0)x418.25555, shifted 17.3562
..\vbox(17.0+0.0)x418.25555, glue set 17.0fil
...\glue 0.0 plus 1.0fil
...\hbox(0.0+0.0)x418.25555
..\glue 20.40001
..\glue(\lineskip) 0.0
..\vbox(595.80026+0.0)x418.25555, glue set 573.8954fil
...\write-{}
...\glue(\topskip) 1.00002
...\hbox(9.99998+0.0)x418.25555, glue set 61.1252fil
....\hbox(9.99998+0.0)x15.84006
.....\glue 0.0
.....\OT1/cmss/bx/n/14.4 1
.....\glue 7.92003
....\OT1/cmss/bx/n/14.4 S
....\OT1/cmss/bx/n/14.4 u
....\OT1/cmss/bx/n/14.4 b
....\OT1/cmss/bx/n/14.4 S
....\OT1/cmss/bx/n/14.4 u
....\OT1/cmss/bx/n/14.4 b
....\OT1/cmss/bx/n/14.4 S
....\OT1/cmss/bx/n/14.4 e
....\OT1/cmss/bx/n/14.4 c
....\OT1/cmss/bx/n/14.4 t
....\OT1/cmss/bx/n/14.4 i
....\OT1/cmss/bx/n/14.4 o
....\OT1/cmss/bx/n/14.4 n
....\glue 5.28001 plus 2.64 minus 1.76
....\OT1/cmss/bx/n/14.4 i
....\OT1/cmss/bx/n/14.4 s
....\glue 5.28001 plus 2.64 minus 1.76
....\OT1/cmss/bx/n/14.4 t
....\kern-0.44
....\OT1/cmss/bx/n/14.4 w
....\kern-0.44
....\OT1/cmss/bx/n/14.4 o
....\glue 5.28001 plus 2.64 minus 1.76
....\OT1/cmss/bx/n/14.4 l
....\OT1/cmss/bx/n/14.4 e
....\OT1/cmss/bx/n/14.4 v
....\OT1/cmss/bx/n/14.4 e
....\OT1/cmss/bx/n/14.4 l
....\OT1/cmss/bx/n/14.4 s
....\glue 5.28001 plus 2.64 minus 1.76
....\OT1/cmss/bx/n/14.4 b
....\kern0.43999
....\OT1/cmss/bx/n/14.4 e
....\OT1/cmss/bx/n/14.4 l
....\OT1/cmss/bx/n/14.4 o
....\kern-0.44
....\OT1/cmss/bx/n/14.4 w
....\glue 5.28001 plus 2.64 minus 1.76
....\OT1/cmss/bx/n/14.4 S
....\OT1/cmss/bx/n/14.4 e
....\OT1/cmss/bx/n/14.4 c
....\OT1/cmss/bx/n/14.4 t
....\OT1/cmss/bx/n/14.4 i
....\OT1/cmss/bx/n/14.4 o
....\OT1/cmss/bx/n/14.4 n
....\penalty 10000
....\glue(\parfillskip) 0.0 plus 1.0fil
....\glue(\rightskip) 0.0 plus 1.0fil
...\write1{\@writefile{toc}{\protect \contentsline {section}{\protect \numberline \ETC.}
...\penalty 10000
...\glue 10.84352 plus 0.94289
...\glue 0.0 plus 1.0fil
...\glue 0.0
...\glue 0.0 plus 0.0001fil
..\glue(\baselineskip) 40.54335
..\hbox(7.05666+0.0)x418.25555
...\hbox(7.05666+0.0)x418.25555
....\vbox(7.05666+0.0)x418.25555
.....\hbox(7.05666+0.0)x418.25555, glue set 206.3902fil
......\glue 0.0 plus 1.0fil
......\OT1/cmr/m/n/10.95 1
......\glue 0.0 plus 1.0fil
...\glue 0.0 plus 1.0fil minus 1.0fil
```

This is quite large and hard to understand. Utilizing the proposed patches together with the "documented .tlg" feature, this could be condensed into a single `.dtx` file

```LaTex
\include{regression-test.tex}
\documentclass{article}
\begin{document}
\START
%<*test>
\begin{sectup}
  \begin{sectup}
    \subsubsection{SubSubSection is two levels below Section}
  \end{sectup}
\end{sectup}
%</test>
%<expect>\section{SubSubSection is two levels below Section}
\newpage
\END
```

where flag `test` generates the `.lvt` and flag `expect` generates the `.lve`. In this format, the intention of the test is clearly visible. Or did I overlook a way to achieve something like this with `.lvt` and `.tlg` only?

josephwright:
I'm not sure your `\section` example is that good a demo: wouldn't you test that your input yielded `\section` rather than testing the typesetting?

dffischer:
> test that your input yielded \section rather

Yes, I indeed think testing the commands directly instead of their output would be more optimal. But,

> I see no possibility to compare both commands directly with something like `\ifx` because the environment redefines all sectioning commands at once, where they become macros possibly relying on conditionals and counters themselves.

So a comparison would have to figure out how to expand and evaluate the redefined `\subsubsection` until it becomes the original `\section`.

This gets even more complicated when a later environment should be able to figure out the topmost sectioning command automatically and adjust, working in combination with `\include`, quite similar to the example in my [post before](https://github.com/latex3/svn-mirror/pull/206#issuecomment-78397751).

PR #207, opened 2015-03-19
support biblatex

dffischer:
Some authors may want to put citations into their documentation, but `l3build` currently has no support for bibliographies in manuals.

This patch automatically runs `biber` in between typeset runs when needed.

I stuck to `biblatex` with the `biber` backend only, because their need is clearly indicated whenever a `.bcf` file is present. Checking for the need to `bibtex` is more difficult, as its directives hide in the the `.aux` file.

### DIFF
--- a/l3build/l3build.lua
+++ b/l3build/l3build.lua
@@ -1021,8 +1021,7 @@ function doc ()
       end
     end
     local function typeset (file)
-      local errorlevel =
-        os.execute (
+      return os.execute (
             os_setenv .. " TEXINPUTS=" .. typesetdir .. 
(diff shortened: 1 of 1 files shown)


josephwright:
I'll look at this later, but just to say that BibTeX/Biber support is in my mind as I have packages needing exactly that! (I'd avoid using BibTeX or Biber for references _in_ a `.dtx` or similar, but for a 'live demo' it's obviously required.)

josephwright:
I've added more-or-less this code in SVN 5557 (as noted earlier, I can't merge using Git). In the commit message there I've mentioned other outstanding questions in this area: I'll see what I can come up with, but input most welcome!

PR #209, opened 2015-03-28
ignore all line numbers at the start of log lines

dffischer:
Recently, I had to test error messages with l3build and found that they also include line numbers that interfere with the comparison of `.tlg` files. Commit 7abddccc8569af0140bda59585637dbc10e8a9fb does not suffice to remove them, so I modified the regular expression introduced therein to universally catch line numbers at the start of lines.

### DIFF
--- a/l3build/l3build.lua
+++ b/l3build/l3build.lua
@@ -554,7 +554,7 @@ function formatlog (logfile, newfile)
     -- Zap "on line <num>" and replace with "on line ..."
     line = string.gsub (line, "on line %d*", "on line ...")
     -- Zap line numbers from \show, \showbox, \box_show and the like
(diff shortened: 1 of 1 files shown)


josephwright:
I have a feeling we discussed something akin to this and for some reason didn't do it. However, I can't spot a problem so will make the change. As there's a knock-on on LaTeX2e and we are in the middle of a CTAN update for that, I'll hold off for a few days.

josephwright:
An aside: Perhaps consider opening an issue asking the team to us Git 'natively' if you are expecting to make a lot of pull requests! We've got some internal discussion ongoing on this, and it would be useful to see if others would really make use of this possibility.

PR #262, opened 2015-08-26
actually utilize example test from documentation

dffischer:
Adding the example test from the manual as an actual test does not only ensure that the example is actually valid, it also helps the user to understand the feature as he can observe it in action right within the sources of latex3 itself.

Doing so, I recognized that l3build should be able to find tests in the `testdir` also when they are packed.

### DIFF
--- a/l3build/l3build.lua
+++ b/l3build/l3build.lua
@@ -583,7 +583,7 @@ function checkinit()
   for _,i in ipairs(filelist(localdir)) do
     cp(i, localdir, testdir)
   end
-  bundleunpack()
+  bundleunpack({".", testfiledir})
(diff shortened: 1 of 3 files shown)


josephwright:
All looks fine module very minor comments above: just need to discuss with the team whether we directly use pull requests with the updated backend set-up.

josephwright:
BTW, if you are happy I'll do the integration using `rebase` as it keeps our history cleaner: particularly for SVN that's useful.

dffischer:
Yes, please, go ahead. I highly value a clean history and therefore utilize `rebase` very often myself.

dffischer:

PR #265, opened 2015-09-01
 copy recursively

dffischer:
Some users may want to specify whole directories in `docfiles`, `supportfiles` or the like. Imagine for example a complex documentation that should be neatly hierarchically assorted by chapters and sections.

This patch teaches l3build's `cp` function to copy around directories and everything contained within.

### DIFF
--- a/l3build/l3build.lua
+++ b/l3build/l3build.lua
@@ -412,11 +412,11 @@ function cp(glob, source, dest)
     local source = source .. "/" .. i
     if os_windows then
       os.execute(
-          "copy /y " .. unix_to_win(source) .. " "
-            .. unix_to_win(dest) .. " > nul"
+          "xcopy /y /e /i" .. unix_to_win(source)
+            .. " " .. unix_to_win(dest) .. " > nul"
         )
     else
-      os.execute("cp -f " .. source .. " " .. dest)
+      os.execute("cp -rf " .. source .. " " .. dest)
     end
   end
 end


josephwright:
Merged (rebased) as https://github.com/latex3/latex3/commit/3b95dbf5f603eb8624389c17a4079e4843f7f9dc.

dffischer:
Thanks for the quick processing, as always.

josephwright:
I've had to back this out: initial testing looked OK but when I tried to build `siunitx` on Windows I got an error about cyclic copying. I'll need to test more carefully: will do later today or tomorrow.

josephwright:
The exact error message is covered in https://technet.microsoft.com/en-us/library/cc959231.aspx. Tracing with `siunitx` as a model and replacing `os.execute` with `print` I get
```
xcopy /y /e /i .\siunitx.dtx .\build\unpacked > nul
xcopy /y /e /i .\siunitx.dtx .\build\doc > nul
xcopy /y /e /i .\siunitx.dtx .\build\doc > nul
xcopy /y /e /i .\siunitx.dtx .\build\unpacked > nul
```
The issue then is that the way we currently have the copy system set up can be from 'various places'. That then leads to the addition of `.\` in the above, but `xcopy` then sees the destination as a subset of the source (both start `.\`). It might be possible to adjust that: need to think about the copying.

dffischer:
I am no Windows user nor do I have regular access to a Windows machine. But I had the chance to experiment with these a little bit. It seems the error does not originate from the `.\`, but occurs whenever a regular file residing in the local directory is to be copied with the new switches. So using `/e /i` only when the source file is a directory could remedy the problem.

josephwright:
@dffischer Luckily I am and I do :-) I'll work on this later in the week (first up is (u)pTeX support).

josephwright:
At the moment I'm going to close this: I think we'll wait for some real use cases before I try to get it right!

dffischer:
Meanwhile, I got my hands on a Windows machine and tested the assumption stated in [my last comment](https://github.com/latex3/latex3/pull/265#issuecomment-138341388). It seems to be working this way, so I rebased [my branch](https://github.com/dffischer/latex3/tree/copy) onto master and pushed [a fix there](https://github.com/dffischer/latex3/commit/fc7f12522514a9bad9b751c6ad7cd2d24c540cd2).

I also tried to test the changes with [`siunitx`](https://github.com/josephwright/siunitx), but it does not seem to compile even without this patch. Compiling `l3build` itself with the patch works fine.

josephwright:
I've opened a new pull request for this as GitHub seems a bit confused :-)

dffischer:
Committing forth and back and forth again, it has all reason to be. I hope it won't have to deal with this change again. Thank you for the rapid merge.

PR #271, opened 2015-09-25
Two typos in XeTeX primitive renaming

qinglee:
`\xetex_charglyph:D` is required in `xeCJK`. So I would be very grateful for a new CTAN release.


josephwright:
Merged into the SVN

josephwright:
I've just fixed one thing on CTAN: I'll do another fix tomorrow for this (we've got other changes as well so it's non-trivial: I'll have to check we are OK to go).

PR #276, opened 2015-10-01
Create test case for the CTeX package

LiamHuang0205:
1. Create `contrib/testfiles/CTeX001.lvt` as the test example.
2. Create `contrib/testfiles/CTeX001*.tlg` files for engines.
3. Modify `texlive.sh` for the deps of the CTeX package

@qinglee @leo-liu


qinglee:
Please add `xtemplate` to the variable `checkdeps` in `build.lua` as it is required by `xeCJK`.

LiamHuang0205:
Okay, I'll check all of these, and make a new PR later.

josephwright:
I think you're supposed to be able to simply add more commits. (BTW, this works best if you make a separate branch, as all new commits to the branch you use automatically go into the pull request.)

LiamHuang0205:
Well, I'll take your advice and have a try. `: )`

josephwright:
BTW, we've released [from this test :-)] that there's an issue with the LaTeX2e format update and CTeX. @davidcarlisle  and I will fix that probably today (I have a fix available already, just need to work out if its OK). So for testing here, don't update just yet!

josephwright:
@qinglee  I'd spotted the `xtemplate` business: I'll do that once the pull request gets added (or it could be part of it).

josephwright:
@qinglee, @LiamHuang0205 Does using `gbsn` rather than `zhmetrics` gain anything? As noted, the former will need `arphic` and it's quite big (it contains fonts) whereas `zhmetrics` is pretty small.

leo-liu:
@josephwright `zhmetrics` needs Windows platform to use the actual TrueType fonts preinstalled in the OS.

josephwright:
@leo-liw Only if you actually want to make a PDF ;-) We only need the metrics to get the log to work (and we can if required force DVI mode for pdfTeX to avoid any attempt to actually load fonts).

josephwright:
@leo-liu On my test VM (Ubuntu 12.04TLF) the pdfTeX tests other than LuaTeX work fine without adding any fonts: the log is OK even if the resulting PDF is unusable. We only have to know that the test file gives the correct visual result on some system where the fonts are available.

josephwright:
@LiamHuang0205 BTW, I hope my comments are OK: as I say, this pull request business is a new one to me!

I suspect there's still going to be an issue with the LuaTeX test on the VM. I'm trying doing a full install of TL there and will report back, but the minimal system gives a failure and I'm not sure why!

qinglee:
@josephwright `fandol` is a CID font, so `luaotfload` needs some mapping files to handle it correctly. `fandol` uses `Adobe-GB1-4.cidmap` which is included in `context`. But we don't install it in our test.

LiamHuang0205:
@josephwright 
`xCJK2uni` is used when `CTeX` is working with the `GBK` encoding, and in our test case, it's useless.

josephwright:
As I've commented on the test by @qinglee, both look quite good. For the present, could we make this test `ctex002`? Also, until I pin down the issue with the VM the best plan with the LuaTeX test  is to bail out: the missing font business is pretty hard to mask. (I've no idea why we get that issue!)

qinglee:
@josephwright Do you mean that  `fandol` is reported by `luaotfload` as having no chars? As I mentioned in the previous comment, `fandol` is a CID font, `luaotfload` needs some mapping files to handle it correctly. But those files are included in `context`. We don't install `context`, so `luaotfload` gives a wrong report.

josephwright:
@qinglee  You see the same, though, if you do install `context`, if you add that file to the local tree or if you do a complex TL install on the Ubuntu VM I have.

josephwright:
@qinglee  Actually, things do work now: I'll track down the exact pattern required.

qinglee:
@josephwright is a bit strange. Do you run `luaotfload-tool --cache=erase` to remove the old cache located at `texmf-var\luatex-cache\generic\fonts\otf` before testing?

josephwright:
@qinglee  Like I say, let me work on it. I want first to alter some parts of the script to make it easier to add deps. I've got to go out in a bit, so it might be this evening (European time) before I can sort.

josephwright:
@qinglee  OK, I can now get this all working. I'm going to alter the scripts a bit, then will ask for another pull update!

qinglee:
@josephwright Fine! `fandol` is a big package (`21089k`). Do you think it is acceptable to install it? If it is OK, I will use it in `ctex001.lvt`.

josephwright:
@qinglee Keeping size down is good, even with caching, but working tests are more important! I think we are fine to use `fandol`. Hopefully I've now sorted the other issues.

josephwright:
I think I've now got the test system sorted such that it will run here barring the `xtemplate` business, which I'll adjust shortly. Could you make the test files here `ctex002`, rebase following the changes to the script, _etc._, and I think we'll be good to go.

josephwright:
BTW, if we could avoid `arphic` that would be good :-)

josephwright:
This has been covered by #280 so I'll close here.

PR #277, opened 2015-10-02
Add a basic test for ctex

qinglee:
This test is only to indicate running without any errors.

@LiamHuang0205 @leo-liu

### DIFF
--- a/contrib/build.lua
+++ b/contrib/build.lua
@@ -15,6 +15,7 @@ checkdeps   =
   {
     maindir .. "/l3build",
     maindir .. "/l3packages/xparse",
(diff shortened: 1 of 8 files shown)


qinglee:
@josephwright I think we had better copy ` texlive-fontconfig.conf` to `~/.fonts.conf`

josephwright:
@qinglee  No need to copy any files: we have a hard-coded location for TL so can just `cat` the necessary data to the correct place [or do what I do and stick to fonts by name :-)].

BTW, I plan to revise the script further to make it easier to add dependencies, but I'll wait until the current pull requests are sorted.

PR #284, opened 2015-11-05
Copy recursively

josephwright:
New pull request following #265.


josephwright:
Looks fine for me with `siunitx` on my Windows system so I've merged.

PR #310, opened 2016-08-24
l3doc: Make sure hyperlink destination always exists.

qinglee:


josephwright:
This got commited as fa79d9591e413a0fb46f7d0aac30486c28c741fc: GitHub seems not to have picked that up ...

PR #355, opened 2017-04-10
Also typeset unpacked files

dffischer:
Usually, documentation is generated by simply typesetting a `dtx` file. In more complex cases, the file to typeset may first have to be unpacked from this `dtx`. The commits proposed herein allow for this scenario by searching for `typesetfiles` among the unpacked content, too.


dffischer:
Thank you for the quick processing.

PR #358, opened 2017-04-15
establish a typesetsourcefiles variable

dffischer:
This table essentially does for the *doc* command what the `checkfiles` do for *check*: Files specified in there are treated like `sourcefiles`, which means they are pulled into the unpacking process, only when typesetting the documentation.

Moving files needed in the documentation from the `sourcefiles` to this variable keeps them from interfering with other processes and speeds them up.

### DIFF
--- a/l3build/l3build.dtx
+++ b/l3build/l3build.dtx
@@ -89,6 +89,7 @@
 \luavarset{typesetdemofiles}{\{~\}}       {Files to typeset before the documentation (as demos), but where the PDF results are not included in the release.}
 \luavarset{typesetfiles}    {\{"*.dtx"\}} {Files to typeset for documentation.}
 \luavarset{typesetsuppfiles}{\{~\}}       {Files needed to support typesetting when \enquote{sandboxed}.}
+\luavarset{typesetsourcefiles}{\{~\}}     {Files to copy to unpacking when typesetting.}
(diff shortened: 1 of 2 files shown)


josephwright:
Can you suggest a scenario where this is required? The idea of `sourcefiles`/`typesetfiles` was that they are relatively clear and between them cover a wide variety of cases.

josephwright:
To add that performance isn't the key requirement here. In particularly, most of the time for building is going to be taken up by running tests and/or documentation (lots of TeX runs): the time for looping in Lua never seems to be significant. (File copying does take a bit of time, but again disk operation related.)

dffischer:
Most problems one can think of stem from unpacked or compiled `*.lvt`. `*.lve`, `*.lvg` or even `*.pdf` files mixing up the `check` command. An example for this can be found [here](https://github.com/dffischer/typesetsourcefiles-example). The [README](https://github.com/dffischer/typesetsourcefiles-example/blob/master/README.md) there has more information on how the new variable helps.

In combination with #359, it provides for further comfort. When it only applies to files in the `docfiledir` introduced in that pull request, it would be possible to specify rules like

    sourcefiles = { "important.*" }
    docfiledir = { "doc" }
    typesetsourcefiles = { "*" }

meaning: "For typesetting, copy in all important files from the base directory as well as everything found in _doc_/"

Issue #27, opened 2011-05-27
\iow_wrap broken when arg contains \bgroup


blefloch:
I only have a high-level example:

```
\RequirePackage{xparse}
\DeclareDocumentCommand{\foo}{O{\bgroup}}{#1}
```

breaks when trying to write the info message for the signature. What happens is that parsing with `\futurelet` is not sufficient to distinguish explicit `{` from `\bgroup`.  Then `\iow_wrap_group:n` thinks that it has an explicit `{` in front of it, and grabs an undelimited argument, puts a brace and continues parsing, hitting the same character.

Two solutions: simply detokenize before wrapping (maybe only in some situations?), or do some more elaborate parsing.

PR #359, opened 2017-04-15
introduce a docfiledir

dffischer:
This table essentially does for the *doc* command what the `testfiledir` dos for *check*: This directory is searched for matching files to copy into any process only when typesetting the documentation. This allows to move more complex documentation sources into their own directory, improving project structure.

### DIFF
--- a/l3build/l3build.lua
+++ b/l3build/l3build.lua
@@ -101,6 +101,7 @@ sourcefiles      = sourcefiles      or {"*.dtx", "*.ins"}
 textfiles        = textfiles        or {"*.md", "*.txt"}
 typesetdemofiles = typesetdemofiles or { }
 typesetfiles     = typesetfiles     or {"*.dtx"}
+docfiledir       = docfiledir       or maindir .. "/doc"
 typesetsuppfiles = typesetsuppfiles or { }
 unpackfiles      = unpackfiles      or {"*.ins"}
 unpacksuppfiles  = unpacksuppfiles  or { }
@@ -1859,7 +1860,7 @@ function doc(files)
     cp(i, supportdir, typesetdir)
   end
   depinstall(typesetdeps)
-  unpack()
+  unpack({".", docfiledir})
   -- Main loop for doc creation
   for _, typesetfiles in ipairs({typesetdemofiles, typesetfiles}) do
     for _,i in ipairs(typesetfiles) do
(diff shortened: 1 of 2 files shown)


l3build/l3build.lua
```diff
@@ -1862,7 +1863,7 @@ function doc(files)
     cp(i, supportdir, typesetdir)
   end
   depinstall(typesetdeps)
-  unpack()
+  unpack({".", docfiledir})
```
josephwright: I'm not clear on this: `unpack()` is for extracting package code from sources.

l3build/l3build.lua
```diff
@@ -2069,12 +2070,12 @@ end
 
 -- Unpack the package files using an 'isolated' system: this requires
 -- a copy of the 'basic' DocStrip program, which is used then removed
-function unpack()
+function unpack(sourcedir)
```
josephwright: I'm currently working on properly documenting the internal functions. My feeling at the moment is that the top-level ones should have arguments which are more-or-less restricted to just the command line `<files>` (where appropriate). (There are a few issues with that which is why I'm working through the code!) Here, I'm not sure having an argument makes sense: the 'main' source dir is 'here' and at least in this pull request the argument here is not actually used.

l3build/l3build.lua
dffischer: The `unpack` command extracts package and related code, e.g. a `*.sty` itself, tests or documentation. The new parameter specifies directories to look into when gathering files that are to be unpacked.

l3build/l3build.lua
dffischer: The parameter is solemnly used to pass on to `bundleunpack`. For both, it is optional. Nonetheless, I agree that user-facing functions should indeed have easy understandable parameters.

For `unpack`, a user may expect a parameter to specify a file to unpack from or a file to generate. (The latter is difficult already as it is unknown in advance.) So I agree that "directory to collect files for unpacking" is a rather unintuitive parameter.

The parameter could be hidden by moving all `unpack` code to an internal function, turning the user-facing `unpack` into a simple wrapper.

Alternatively, it could be avoided completely by writing a dedicated initialization function for the `doc` function, similar to how `check` already has `checkinit`. On the other hand, this would largely duplicate code from the `unpack` and `checkinit` functions. Would this be preferable?

dffischer:
For all my other pull requests, I compiled some example project. As this one just has the sole purpose to move all the documentation-related files to a common directory, I will simply put an example structure here.

    .
    \-- package.sty
    \-- doc
        \-- documentation.tex
        \-- image-for-documentation.png
        \-- graph-for-documentation.csv
    \-- testdir
        \-- 1-testone.lvt
        \-- 1-testone.lve
        \-- 2-testtwo.lvt
        \-- 2-testtwo.lve

josephwright:
This one is currently conflicting and looks a bit tricky. We're about to do some rearrangement of where `l3build` lives. I'm going to close the pull request, do the rearrangements then log the idea as an issue: I will look to provide something of this type but it may not be immediate.

dffischer:
I ported the changes over to the new l3build repository and rebased them to the current master. The resulting [pull request #27](https://github.com/latex3/l3build/pull/27) should now cleanly apply.

PR #360, opened 2017-04-15
Glob subdirectories

dffischer:
The changes herein provide the facilities to perform multi-directory globbing, i.e. work with patterns that include slashes `/`. They can enable l3build to work with complex project hierarchies, processing files not only from a single directory but also looking into paths below a starting point.

For the moment, I employed this features in the typesetting and unpacking process only, for as I currently deal with just these areas. However, I modeled the subdirectory-aware file-listing function `tree` as close as possible to its single-directory counterpart `filelist`, which should make it possible to apply it in more areas in the future.


l3build/l3build.lua
```diff
@@ -1643,8 +1629,16 @@ function cmdcheck()
   for _,i in ipairs(typesetsuppfiles) do
     cp(i, supportdir, testdir)
   end
+<<<<<<< HEAD
```
josephwright: Presumably here and elsewhere something is up!

l3build/l3build.lua
```diff
@@ -565,27 +565,13 @@ function mkdir(dir)
...
-    target = gsub(target, trimpattern, "")
-    source = gsub(source, trimpattern, "")
-  end
-  -- Go up from the source
-  for i = 0, select(2, gsub(target, "/", "")) do
-    resultdir = resultdir .. "../"
-  end
-  -- Return the relative part plus the unique part of the target
-  return resultdir .. target
+-- Return an absolute path from a relative one
```
josephwright: This seems very risky to me:  we end up with debug data, etc., that depends on the system in use (rather than being relative to the working dir).

l3build/l3build.lua
```diff
@@ -552,6 +552,28 @@ function filelist(path, glob)
   return files
 end
 
+-- Does what filelist does, but can also glob subdirectories. In the returned
+-- table, the keys are paths relative to the given starting path, the values
+-- are their counterparts relative to the current working directory.
+function tree(path, glob)
+  function cropdots(path)
```
josephwright: Should be local?

l3build/l3build.lua
```diff
@@ -2125,20 +2125,21 @@ bundleunpack = bundleunpack or function(sourcedir)
...
       -- TeX if \askforoverwrite is true
       -- That is all done using a file as it's the only way on Windows and
       -- on Unix the "yes" command can't be used inside execute (it never
       -- stops, which confuses Lua)
       execute(os_yes .. ">>" .. localdir .. "/yes")
+      local path, name = splitpath(j)
       local localdir = abspath(localdir)
       errorlevel = run(
-        unpackdir,
+        unpackdir .. "/" .. path,
```
josephwright: Won't this add a trailing `/` in all cases (so incorrect if the `path` is empty)?

josephwright: I'm not sure about the logic here: _should_ unpacked files retain the subdivisions present in the _sources_?

l3build/l3build.lua
```diff
@@ -1383,29 +1383,31 @@ end
 --
 
 -- An auxiliary used to set up the environmental variables
-function runtool(envvar, command)
+function runtool(subdir, dir, envvar, command)
   return(
     run(
-      typesetdir,
+      typesetdir .. "/" .. subdir,
```
josephwright: As with a previous commit, don't we end up with an incorrect trailing `/` here? (I wonder about dividing up the typesetting into directories: couldn't things be done by working through the directories and copying back the results as required without needing to copy the files in the structure?)

l3build/l3build.lua
josephwright: I guess what I'm trying to get at is that provided the two locations are on the same disk there is a relative path from one to the other, so one should be able to work out the way to traverse it.

josephwright:
Hmm, that review was meant to be about a single commit: not quite got the hang of this GitHub-based business!

l3build/l3build.lua
dffischer: That was a rebase gone wrong. It is now fixed and properly rebased on the current master.

l3build/l3build.lua
dffischer: As noted in the first commit comment, the previous `relpath` was broken and I needed a working base to implement globbing on. Unfortunately, lua's limited filesystem functions lack the tools to properly implement a more elegant version, so I quickly hacked together `abspath` sans a better alternative.

As it temporarily changes the current working directory of the whole program, it should only be used with caution. I strongly suggest to replace it with something better in the future. Teaching the old `relpath` how to handle `.` and `..` could be the far superior alternative. A conversion function for MakeIndex is then easy to write.

l3build/l3build.lua
dffischer: I did not know enough Lua to suspect that this might be a global declaration.

    local cropdots = function(path)

Is this what you want to indicate? Or should `tree` be local?

l3build/l3build.lua
dffischer: Yes, the core intention is that unpacked files should retain the same directory structure as their sources are assorted in the project layout. I came to this conclusion by thinking about the following structure.

    .
    \-- document.tex
    \-- dir
        \-- include.tex

In `document.tex`, `include.tex` is used in compiled form. Then, in `document.tex`, I suspect it to be more intuitive to write

    \includegraphics{dir/include.pdf}

than

    \includegraphics{include.pdf}

It also avoids name clashes when files in different directories have the same names.

l3build/l3build.lua
dffischer: As is is also the case with the previous comment about a trailing `/`, nearly all paths are of the form `./path/to/a/file`. So `path` should never end up empty to begin with, but at least be `.`. The parameter is then passed to `cd`, which also is not confused by a trailing `/`.  But a check for an empty `path` could well be added.

dffischer:
As noted above, the commit picked is to be used carefully. I do not know whether anything in l3build uses threads or other, possibly hidden, means of concurrent processing, but especially these could severely break due to the working directory change in `abspath`.

I also composed [an example project](https://github.com/dffischer/recursive-example) to show where large documentations may benefit from handling subdirectories.

l3build/l3build.lua
josephwright: OK, I guess users have a tendency to do this, even thought the team position is names should be unique (and flat).

josephwright:
I've rebased/merged this manually, with the exception of 69a1c95cd: the standard settings should reflect the fact that  the team's recommendation is to use a flat structure.

dffischer:
Thanks for the - admittedly complex - merge.

PR #361, opened 2017-04-15
provide table for files typeset but not shipped

dffischer:
The `typesetfiles` are all meant to be shipped out. The commit herein introduces a new variable collecting files that are only compiled, but not processed by l3build any further. They are instead meant to be utilized by other `typesetfiles`.

This is often the case, when the documentation of a package showcases examples of its output, which are usually included as images. These images have to be compiled from a separate *`.tex` before, which can be specified in the new `typesetsilent` table.


josephwright:
Good idea but I'm not quite sure about the set up. First, all of the file tables have names ending `...files`, which I'd like to stick to. Second, I'm not sure if this is best viewed as 'additional to `typesetfiles`' or 'subsetted out of `typesetfiles`'.  I wonder about `typesetsubfile` or `typesetdemofiles`: 'Files which will be typeset but not copied back to the main directory; useful for compling examples', with (probably) a statement that they are compiled _before_ the main runs.

(BTW, the same effect could currently be achieved using shell escape and appropriate settings, but I can see that this is suboptimal.)

josephwright:
Something like `typesetdemofiles` I guess is best 'in addition to `typesetfiles` rather than `Ignored when copying `typesetfiles` output back to the main directory'.

josephwright:
I've rebased, _etc._, this commit by hand: as I've mentioned above I've gone for a slightly different variable name in the end.

dffischer:
Thank you. The name is more telling, indeed.

PR #366, opened 2017-05-22
l3candidates: Add preliminary shell-escape to l3sys

### LINKED ISSUE #288: Add the possibility for shell-escaping
There is `\c_log_iow` and `\c_term_iow`.  Please add `\c_shell_iow`.

```
\tex_chardef:D \c_eighteen = 18 ~
\cs_new_eq:NN \c_shell_iow \c_eighteen
```

hmenke:
Fixes #288


l3kernel/l3candidates.dtx
```diff
@@ -2006,6 +2006,113 @@
...
+    \sys_shell_now:n { #1 }
+  }
+  {
+    \iow_shipout:Nn \c__sys_eighteen { #1 }
+  }
+}
+%    \end{macrocode}
+%   \emph{Note:} There is no method for delay in LuaTeX (perhaps through
+%   a callback).  This should be discussed.
+% \end{macro}
```
davidcarlisle: I don't think you need a callback here just use `\latelua` rather than `\directlua` (cf `\ShellEscape` and `\DelayedShellescape` in tools/shellesc.)

hmenke: Thanks for the hint!  I'll upvote one of your questions on TeX.SX.

hmenke:
@josephwright How do I get the CI to pass?

josephwright:
@hmenke You need to update the test files :) In this case, adding or removing _any_ function/variable from the `expl3` core always shows up in `m3expl001` and `m3expl003`. So you want
```
texlua build.lua save -epdftex,xetex,luatex,ptex,uptex m3expl001 mexpl003
```
from inside `l3kernel` to update the `.tlg` files. (You also need to be using the same TL version as Travis-CI, so TL'16 at present.)

josephwright:
We've got a team meeting this week so I'll likely raise this there. Quickly looking over the  diffs, I don't see any documentation part to the commits: could you add?

josephwright:
@hmenke  I see I had a typo in my comment which might be why you didn't get a test pass: try with the edited version!

hmenke:
@josephwright Is the inline documentation not sufficient?

josephwright:
@hmenke  You'll notice that `expl3` has formal API documentation for _everything_: you could probably move most of the code comments to the documentation part to create it here. (Notice that `interface3` doesn't have any code-part in it, so at present the functions added here won't show up in the docs.)

josephwright:
@hmenke  To be clear, we'll almost certainly take this pull request: I can always sort out such minor issues later. (For example, I suspect `\__sys_shell_escape:` should be `\g__sys_shell_escape_bool`, but I can change that later.)

hmenke:
@josephwright Actually it shouldn't be a bool but and int.  Shell esape has three possible states (off, restricted, unrestricted).  I edited my code accordingly.  Please let me know before you merge, so I can squash my changes into a single commit.  I do not want to litter the commit history with my failed attempts to run the CI.

josephwright:
@hmenke  I'll probably not do anything until after the team meeting (Thursday) (although this is for 'candidates' so is even more 'down to me' than usual ...). Don't worry overly about the history: we have _lots_ of commits of the 'Drat, forgot XXX' type. (Remember the core repo is SVN, so there is no hiding these things.)

l3kernel/l3candidates.dtx
```diff
@@ -506,6 +506,61 @@
...
+% \begin{function}[added = 2017-05-22, EXP, pTF]{\sys_if_shell_unrestricted:}
+%   \begin{syntax}
+%     \cs{sys_if_shell_unrestricted_p:}
+%     \cs{sys_if_shell_unrestricted:TF} \Arg{true code} \Arg{false code}
+%   \end{syntax}
+%   Performs a check for whether \emph{unrestricted} shell escape is
+%   enabled.
+% \end{function}
+%
+% \begin{function}[added = 2017-05-22, EXP, pTF]{\sys_if_shell_unrestricted:}
```
blefloch: Typo: `\sys_if_shell_unrestricted:` => `\sys_if_shell_restricted:`

l3kernel/l3candidates.dtx
```diff
@@ -2006,6 +2061,117 @@
...
+        \pdftex_shellescape:D
+      }
+  }
+%    \end{macrocode}
+% \end{variable}
+%
+% \begin{macro}[EXP, pTF]{\sys_if_shell:}
+%   Performs a check for whether shell escape is enabled.  This will
+%   return true if either of restricted or unrestircted shell escape
+%   is enabled.
```
blefloch: Typo unrestircted.

l3kernel/l3candidates.dtx
```diff
@@ -2006,6 +2061,117 @@
...
+      \prg_return_true:
+    \else:
+      \prg_return_false:
+    \fi:
+  }
+%    \end{macrocode}
+% \end{macro}
+%
+%    \begin{macrocode}
+\tex_chardef:D \c_eighteen = 18 ~
```
blefloch: We're slowly moving towards fewer named integers, so it will be better renamed as `\c__sys_` something.

l3kernel/l3candidates.dtx
```diff
@@ -2006,6 +2061,117 @@
...
+%    \end{macrocode}
+%
+% \begin{macro}{\sys_shell_now:n}
```
blefloch: In cases with engine-dependent definitions we've so far preferred to do the engine test outside the definition to avoid repeating the test multiple times.

hmenke:
@blefloch Done, using `Npn` now.  Also I squashed all commits together into one.

Issue #373, opened 2017-06-09
xparse: Number and dimension arguments


hmenke:
I was thinking that it would be cool if the parser could check whether an argument is a number or a dimension, so the user does not have to check it themselves.  The new shortcuts `n` for number and `l` for length seem appropriate. (EDIT: unfortunately, both `l` and `d` are taken)
````latex
\NewDocumentCommand \concept { n l }
{
  \looseness=#1
  \setlength\tabcolsep{#2}
}
````
This seems very user-friendly to me because the error is caught immediately when trying to expand the first stage.  Otherwise the user would get
````
Missing number, treated as zero.
<to be read again> 
                   .
l.3 \looseness=.
                ..
````
without knowing where it comes from.

**Comment by eg9** [comment]
I'd not add it.

**Comment by josephwright** [comment]
The nature of `xparse` argument types is that they relate to grabbing the argument. WE have got 'processors' which could be used here
```latex
\documentclass{article}
\usepackage{xparse}

\NewDocumentCommand \concept { >{\CheckInteger}mm }
{
  \looseness=#1
  \setlength\tabcolsep{#2}
}
\ExplSyntaxOn
\cs_new_protected:Npn \CheckInteger #1
  {
    \regex_match:nnF { ^[\+\-]*[\d]+$ } {#1}
      {
        \ERROR % Some real message!
      }
  }
\ExplSyntaxOff
\begin{document}

\concept{10}{10ml}

\end{document}
```
although in general we've not tried to cover validation.

hmenke:
The processors are a good feature.  I like the idea of adding a set of validators.  Is `l3regex` part of the kernel?

josephwright:
We moved `l3regex` to the kernel in the most recent release.

josephwright:
I think at present this is a 'no' for the reasons outlined above.

PR #380, opened 2017-07-14
[xparse] Define a set of validators

hmenke:
Alternative approach at #373.

The regexes for floats and dimensions might look a little weird, but in fact `.pt` is a valid dimension in TeX (and thus `.` is a valid floating point literal as well).
```latex
\dimen0=.pt
\the\dimen0 % 0.0pt
```

### DIFF
--- a/l3packages/xparse/xparse.dtx
+++ b/l3packages/xparse/xparse.dtx
@@ -3833,6 +3833,7 @@
         \__msg_kernel_error:nnxxx { xparse } { type-mismatch }
           { \tl_to_str:n {#1} } { int } { \iow_char:N \\ \l_@@_function_tl }
       }
+    \tl_set:Nn \ProcessedArgument { #1 }
   }
 %    \end{macrocode}
 % \end{macro}
@@ -3847,6 +3848,7 @@
         \__msg_kernel_error:nnxxx { xparse } { type-mismatch }
           { \tl_to_str:n {#1} } { fp } { \iow_char:N \\ \l_@@_function_tl }
       }
+    \tl_set:Nn \ProcessedArgument { #1 }
   }
 %    \end{macrocode}
 % \end{macro}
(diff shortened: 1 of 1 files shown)


eg9:
An integer in TeX can also be in octal, hexadecimal or alphabetic format.

josephwright:
@eg9  True, but at the interface level one is free to define whatever one wants. I might be tempted to consider `\afterassignment` here and check for 'residue', if the aim is to see 'Will TeX parse fully as an integer'.

blefloch:
While the code probably works, the processors ought to set `\ProcessedArgument` I believe.  Perhaps it would be better to use `\regex_extract_once:nnN {#1} { ^([+-\ ]*\d+\ *)(.*)$` }` into some seq, then check the number of items and store just the integer in `\ProcessedArgument`.  Hopefully I got the regex right, allowing more spaces and an arbitrary number of signs.

If people think it's useful I could probably adapt the `l3fp` parser to provide a function to "parse until failure" and return the split token list, the result, and true/false.  It could have an `int`, `dim`, `skip`, `fp` variants.

hmenke:
@eg9 My intended purpose of these macros is not to check whether the argument is a valid TeX integer, but a valid integer literal, i.e. an optional sign and one or more digits.  This brings me to the suggestions by @blefloch.  The regex  `^([+-\ ]*\d+\ *)(.*)$` seems weird to me.  Why would you allow multiple signs?  Furthermore, since this is not a real processor but a validator, it leaves the argument unaltered if it is valid and throws an error otherwise.  Is setting `\ProcessedArgument` required in that case as well?

blefloch:
Yes, while xparse's implementation details currently let you not set the `\ProcessedArgument` when it's not changed, that may very well change in the future (since it's not documented behaviour) so it's safer to stick to the documented interface namely set the `\ProcessedArgument` in all cases.

About signs, it really depends on what applications you have in mind, but I wouldn't be surprised if such double signs show up after a few indirections (user defines a command in terms of the package's command) where at various stages one may want to put a sign.

josephwright:
I do plan to take this: we have some internal changes to make first ...

josephwright:
At the moment I don't think this is going to progress, but we should look again at the idea.

PR #386, opened 2017-08-01
l3str: fix a typo [ci skip]

Stone-Zeng:
Someone use `’` (U+2019) here. It should be `'` (U+0027).

So in the documentation, you will find something strange:
default


josephwright:
There seem to be a lot of commits here for the one line that should change: could you tidy that up?

Stone-Zeng:
Sorry, I'm not very familiar with git. Only the last commit (dee5115) is concerning this issue.

josephwright:
@Stone-Zeng  I'll cherry-pick that one in a bit.

josephwright:
I've cherry-picked the 'real' commit here: closing the PR.

PR #391, opened 2017-09-26
Corrected the definition of 'explicit space'

EvanAad:
By deleting the requirement that it should have character code 32.

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -866,7 +866,7 @@
 %     \cs{tl_tail:n} \Arg{token list}
 %   \end{syntax}
 %   Discards all leading explicit space characters
-%   (explicit tokens with character code~$32$ and category code~$10$)
+%   (tokens with category code~$10$)
 %   and the first \meta{item} in the \meta{token list}, and leaves the
 %   remaining tokens in the input stream.  Thus for example
 %   \begin{verbatim}
@@ -943,7 +943,7 @@
 %   Tests if the first \meta{token} in the \meta{token list}
 %   is a normal \texttt{N}-type argument. In other words,
 %   it is neither an explicit space character
-%   (explicit token with character code~$32$ and category code~$10$)
+%   (token with category code~$10$)
 %   nor an explicit begin-group character
 %   (with category code~1 and any character code). An empty
 %   argument yields \texttt{false}, as it does not have a \enquote{normal}
@@ -959,7 +959,7 @@
 %   \end{syntax}
 %   Tests if the first \meta{token} in the \meta{token list}
 %   is an explicit space character
-%   (explicit token with character code~$12$ and category code~$10$).
+%   (token with category code~$10$).
 %   In particular, the test is \texttt{false} if the \meta{token list}
 %   starts with an implicit token such as \cs{c_space_token}, or if it
 %   is empty.


josephwright:
I'll probably accept this (just need to sort an issue with my public key system), but will then alter anyway :) I think we need to slightly rephrase these parts.

davidcarlisle:
To be honest I think this should be reverted, all character tokens with catcode 10 have character code 32 so it was not an extra requirement just a clarifying statement that is now lost.

blefloch:
Should definitely be reverted; it's wrong.
```
\lccode32=`*
\def\grabnshow#1{\showtokens{|#1|}}
\lowercase\expandafter{\expandafter\grabnshow\space\relax}
\uppercase\expandafter{\expandafter\grabnshow\space\relax}
\bye
```
shows that a catcode-10 charcode-different-from-32 character is grabbed as a normal item.

EvanAad:
David, I stand corrected. Indeed, according to the TeXbook (fourth
doubly dangerous sign on p. 47) a character whose category code is 10
will be converted to a token whose character code is 32.

However, if this is indeed the case, then I don't understand how Bruno
was able to concoct an example in which a token has catcode 10 but a
character code that is different to 32.

eg9:
> On 26 Sep 2017, at 17:46, EvanAad <notifications@github.com> wrote:
> 
> David, I stand corrected. Indeed, according to the TeXbook (fourth
> doubly dangerous sign on p. 47) a character whose category code is 10
> will be converted to a token whose character code is 32.
> 
> However, if this is indeed the case, then I don't understand how Bruno
> was able to concoct an example in which a token has catcode 10 but a
> character code that is different to 32.

Page 47 refers to the process of tokenization. A space token resulting
from tokenization always has character code 32. However this can be
changed into a “funny space” using `\lowercase` or `\uppercase`:

\lccode32=`\z \lowercase{ }

will result in a funny space token having category code 10 and
character code 122.

According to the rules of page 47, also

\catcode`?=10 \lccode32=`z \lowercase{?}

will result in the same token (catcode 10 and charcode 122).

Example:

\catcode`?=10
\let\$\empty %to avoid space eating
\lccode32=`z

\lowercase{\if\$ \$?\message{equal}\else\message{not equal}\fi}

\bye

This will show “equal”.

Ciao
Enrico

EvanAad:
I don't understand Bruno's example. If it does what Bruno claims,
namely creating a catcode-10 token whose character code is different
from 32, then indeed my bug report is wrong, and should be reverted.

However, based solely on David's remark, I don't think that the change
should be reverted, since if indeed category 10 implies character code
32, then the change did not turn a correct statement into an incorrect
statement, and by eliminating redundancy it clarified the statement
(to wit, I would not have reported a bug had it been written this
way).

davidcarlisle:
it is possible to generate catcode 10 characters with other character codes but only by using \uppercase to change the character code of an existing space token. The character scanner (tex's eyes) never makes such a token.  (see "funny spaces" in the TeXBook.)

EvanAad:
Thanks for the clarification, Enrico and David.

So why does the following manuscript typesets 'z'? Based on our
discussion above, after the category code of the character 'z' is set
to 10, it is converted to a token whose character code is 32, i.e. to
an "explicit space character" (to use the terminology of The LaTeX3
Interfaces document). Therefore, it should be discarded by \tl_head:n,
and the manuscript should typset 'a'.

\documentclass{article}
\usepackage{expl3}
\begin{document}
\ExplSyntaxOn
    \catcode`z=10
    \tl_head:n {zabc}
\ExplSyntaxOff
\end{document}

josephwright:
@EvanAad You've forgotten to terminate your assignment :) As we have `\ExplSyntaxOn`, a space is ignored and a line-end is a space, so ``\catcode`z=10`` doesn't stop at the end-of-line. We then have expandable `\tl_head:n`, which expands searching for the end of the number. That tokenizes the argument _before_ the change of catcode so you have a normal `z`. If you change the example to terminate using `\relax` or `~`, or use ``\char_set_catcode_space:n { `z }``, then you will get the expected result.

EvanAad:
OK, so I agree with Bruno that my original bug report was wrong, and
needs to be reversed. However, I also think that the documentation
needs to elaborate on this issue and clarify it.

Here is another example which demonstrates that if a catcode-10 token
has a character code different from 32, the token won't be disregarded
by \tl_head:n. The code creates a macro \funny_space that expands to a
token that has character code 122 (like ASCII z) and category code 10.
It then uses this macro to insert this token in the beginning of a
token list that is passed to the \tl_head:n function. The \tl_head:n
function is invoked between two x characters, which are used as visual
delimiters. The manuscript typesets x␣x, i.e. x, space, x. This shows
that the \tl_head:n function returned the "funny space".

\documentclass{article}
\usepackage{expl3}
\begin{document}
\ExplSyntaxOn
    \catcode`z=10\relax
    \lccode`z=122\relax
    \lowercase{\def\funny_space{z}}
    x\expandafter\tl_head:n\expandafter{\funny_space abc}x
\ExplSyntaxOff
\end{document}

josephwright:
@EvanAad Please see the latest checkins. Depending on the feeling of the rest of the team, I may yet 'back out' all of today's stuff :)

EvanAad:
Joseph, why did you remove the Space tokens section in the last
commit? I thought it was helpful.

Also, what's the difference between explicit and implicit space tokens?

davidcarlisle:
> Also, what's the difference between explicit and implicit space tokens?

the terminology comes from the texbook.  Command tokens let to character
tokens are implicit character tokens (\lbrace being an implicit { token
being the most notable, but latex2e also has \@sptoken as an implicit space
token.

EvanAad:
I see. Thanks, David.

Joseph wrote back in the end of 2011:

"another key aim for LaTeX3 is to provide a complete set of
documentation and support without having to say ‘read The TeXbook’ as
a requirement to make progress."

(source: http://www.texdev.net/2011/12/07/programming-latex3-background/)

I don't know if this is still considered to be a goal of the LaTeX3
project, but if so, all these details ought to make it, eventually, to
the LaTeX3 documentation.

josephwright:
@EvanAad I've backed out the changes as the team need to agree first on how to cover this: probably they will 're-appear' but I need to let everyone have a chance to contribute.

blefloch:
@EvanAad Your code snippet
```
\documentclass{article}
\usepackage{expl3}
\begin{document}
\ExplSyntaxOn
\catcode`z=10\relax
\lccode`z=122\relax
\lowercase{\def\funny_space{z}}
x\expandafter\tl_head:n\expandafter{\funny_space abc}x
\ExplSyntaxOff
\end{document}
```
actually does something slightly different from what you probably expect.  <code>\catcode`z=10\relax</code> changes the catcode of `z`, then <code>\lccode`z=122\relax</code> is tokenized as <code>\lccode` =122\relax</code> and sets the catcode of space (`\lccode32=122\relax`).  Then the argument of `\lowercase` is tokenized as `\def\funny_space{ }`, lowercased to something with a funny space, and the rest of your explanation is then correct.

PR #393, opened 2017-09-28
Correct a typo: district -> distinct.

EvanAad:


josephwright:
I'll merge then sort out that odd test failure (probably a dependency thing)

PR #399, opened 2017-09-29
[l3quarks.dtx / 3. Quark tests] Typo: later -> latter.

EvanAad:


josephwright:
I'll cherry-pick the 'real' commit here :)

EvanAad:
What can I do so that instead of you having to cherry-pick the 'real'
commit, you can simply go ahead and merge my suggested changes?

josephwright:
@EvanAad As I said earlier, make a branch for each pull request (so all typos together would be fine), and don't pull changes to `master` back into that once you've made changes. (With experience you can sort that out, but to start with you are best avoiding it.)

EvanAad:
But this is what I did with my last pull request and yet you cherry-picked it.

josephwright:
@EvanAad If you look at the pull request you'll see you've started from a rather 'messy' position: see the various merge commits (presumably on your `master`). That's why I've cherry-picked. That approach is fine but doesn't work in GitHub: I just have to manually grab your changes.

PR #400, opened 2017-09-29
state \clist_map_variable:NNn assigns locally

EvanAad:
I modeled this fix after commit 375c341


blefloch:
The same applies to `\tl_map_variable:NNn` and `\tl_map_variable:nNn` (documented separately for some reason).  Can you also do that please?

EvanAad:
Sure. Will do.

PR #409, opened 2017-10-18
Fix documentations

wtsnjp:
Using the information which can be corrected from `<arg-spec>` of each functions, we can detect the form of the arguments (e.g. number of arguments, `\Arg` or `\meta`). Based on this concept, I wrote a simple Python3 script to analyze all `syntax` environments in `interface3.pdf`:

- <https://gist.github.com/wtsnjp/c5eac5ce080716818f00238443024362>

Then I fixed every mistakes in the document which I found with the script.

### DIFF
--- a/l3kernel/expl3.dtx
+++ b/l3kernel/expl3.dtx
@@ -436,7 +436,7 @@
 %
 % \subsubsection{Variables: guidance}
 %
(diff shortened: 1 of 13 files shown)


josephwright:
All merged: thanks :)

wspr:
Wow, that's fantastic!! Thanks so much :)

PR #414, opened 2017-11-09
Preserve spaces in key names (fixes #413)

### LINKED ISSUE #413: l3keys: trim spaces instead of remove spaces
The following code demonstrates how l3keys removes all spaces from input keys.

```
\documentclass{article}
\usepackage{expl3}
\begin{document}
\ExplSyntaxOn
\keys_define:nn {bar} { unknown .code:n = { \typeout{`\l_keys_key_tl'=#1} } }
\def\KVB#1{\keys_set:nn {bar}{#1}}
\ExplSyntaxOff
\KVB {key c=book}
\end{document}
```

While this was added for flexibility in accepting input, on balance I think we agree that it is better to be more strict here. It might make sense to have another option to enable whether keys in a family should be normalised. (Or a shorthand to simply define a pair of keys at setup time, one with and one without spaces removed.)

The main motivation for bringing this up is that I needed to use the unknown key in its original form, and it would be cumbersome to say the least to keep track of the unnormalised input while keeping the current behaviour.

josephwright:
The idea of zapping spaces *inside* key names sounds great, but it
becomes problematic when passing data through. It also allows clear
typos to be ignored: not great.


josephwright:
I've set this up as a PR as it is a change in behaviour: before changing on `master`, do we want to do this?

josephwright:
@wspr Could you perhaps try this with 'all' of `fontspec`: the most obvious risk

wspr:
Sorry for the slow feedback. I've run the fontspec and unicode-math test suites, and compiled their documentation, all which isn't as comprehensive as I'd like — but returns no errors...

wspr:
I'll allow you to delete the branch if we're sure we're done here :)

PR #447, opened 2018-03-24
Add variants of `\tl_if_eq:nn`

Stone-Zeng:
I think these functions maybe useful, for example, see https://tex.stackexchange.com/q/55600/136923.

By the way, why there is no `\tl_if_eq_p:nn`?

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -402,7 +402,12 @@
 %   yields \texttt{false}.
 % \end{function}
 %
-% \begin{function}[TF]{\tl_if_eq:nn}
+% \begin{function}[TF, added = 2018-03-24]
+%   {
+%     \tl_if_eq:nn, \tl_if_eq:Vn, \tl_if_eq:on, \tl_if_eq:xn,
+%     \tl_if_eq:nV, \tl_if_eq:no, \tl_if_eq:nx, \tl_if_eq:VV,
+%     \tl_if_eq:xx
+%   }
 %   \begin{syntax}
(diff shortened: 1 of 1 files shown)


josephwright:
I've no objection here, but most of the time the `str` comparison is more useful. Most notably, the string version is expandable: the `tl` version is not, hence the lack of a predicate version.

eg9:
> On 24 Mar 2018, at 14:44 , Zeng Xiangdong <notifications@github.com> wrote:
> 
> I think these functions maybe useful, for example, see https://tex.stackexchange.com/q/55600/136923.

Variants can be added by the programmer as soon as need arises; the kernel provides several
prebuilt ones that appeared to be needed for applications. Others can be added when use cases
turn out to be interesting.

By using \cs_generate_variant:Nn one ensures that already existing variants are not clobbered
by user defined ones. Thus code using it will not do anything wrong if the generated variant is
later included in the kernel.

> By the way, why there is no \tl_if_eq_p:nn?

Because `\tl_if_eq:nnTF` cannot be expandable.

Ciao
Enrico

wspr:
No complaints from me as long as we're consistent with the other variants we provide.

FrankMittelbach:
The policy about variants (though not formally stated anywhere I think (maybe we should do that in interface3 somewhere)) is roughly like this:

 - there are a number of variants that we provide across the board with data structures to have a base set of commads available that are commonly used. For example we typically provide "c" and place of "N"

 - for anything else the idea is to provide a variant only if  there are at least a handful of independent use cases, otherwise the advice is to use \cs_generate_variant:Nn in the package or class file as there is nearly no overhead if at a later stage the variant is already provided by the kernel.

blefloch:
It seems the consensus at least for now is to let programmers add variants themselves.

PR #464, opened 2018-06-14
Update l3prefixes

Stone-Zeng:
There are some missing comma in recent commit. Update for better rendering on GitHub.


Stone-Zeng:
By the way, please consider to add my packages (they have been uploaded to CTAN) to the prefixes list:

- [fduthesis](https://github.com/Stone-Zeng/fduthesis): prefixes `fdu`, `fdudoc` and `fdulogo`
- [zhlipsum](https://github.com/Stone-Zeng/zhlipsum): prefix `zhlipsum`

Of course, I can add them by myself in this PR if necessary.

josephwright:
Changes already made for the typos. I'll add your packages: we don't just 'hover up' packages, it's strictly 'opt in'.

josephwright:
On the 'should we record them', that's the aim: if you look at something like CPAN for Perl, that's more-or-less exactly what happens. The only issue is that of course there it's a requirement of submission to the 'wider world'.

Stone-Zeng:
Some typos are still there...

- algobox
- cellprops
- fontsizes
- rivmath

josephwright:
Should be sorted now.

josephwright:
The aim of `l3prefixes` is to be a starting point for tracking this stuff: I've imagined that over time the data might be moved to another form, I guess in the end perhaps some web-based system that is easy to query. But that requires more effort than a simple text file!

PR #480, opened 2018-09-10
l3drivers: remove extra \c_space_tl

minux:
`\c_space_tl` is passed to a function (`\@@_color_select:n`) which does not
expand its argument, and results in `xxx "color push \c_space_tl Black"`
with the dvipdfmx backend. `\@@_color_select:n` already inserts appropriate
~, so `\c_space_tl` is unnecessary anyway.

To reproduce this bug, build source3 with the (x)dvipdfmx backend using
lualatex/xelatex. (x)dvipdfmx should complain about the incorrect color
specials.


minux:
the failed l3benchmark test is unrelated to this change.

PR #494, opened 2018-11-01
Move \<thing>_rand_item:N to stable

josephwright:
Also \int_rand:n, which is a trivial extension of int_rand:nn.


josephwright:
No objections raised: I'm going to action this.

PR #495, opened 2018-11-01
Move\ tl_range:nnn to stable

josephwright:
At the moment, leave the more tricky braced/unbraced cases as-is.


josephwright:
@blefloch  Hmmm, that is more tricky then, but at the same time it is documented. As we have regexes for really complex stuff, I'd be tempted to accept the limitation here (honestly, how often will people want a `tl` starting with a space!).

@eg9 You are likely best-placed to comment here: thoughts?

josephwright:
@eg9  So you are happy with the behaviour, and it's just a documentation change?

eg9:
@josephwright I guess so. The alternative would be zapping all spaces.

josephwright:
OK, I'll give it a few days then make the change and merge (baring objections).

blefloch:
The equivalences that @egreg stated are incorrect when l+1+m (or l+1+n)
is zero or negative, because by that rule one should repeatedly add l+1
until getting a positive number, which the code most definitely doesn't do.

The proposed documentation also doesn't say what happens when bounds are
not in the correct order, which the old documentation said I think.

Not sure how to best document things.

eg9:
@blefloch 
OK, let me retry.

Let $m$ and $n$ be the start and end indices. If either is $0$, the result is empty. A positive index means ``start counting at the left end'', a negative index means ``start counting at the right end''. Let $l$ be the count of the token list.

The actual start point is determined as $M=m$ if $m>0$ and as $M=l+m+1$ if $m<0$. Similarly for obtaining the actual end point $N=n$ if $n>0$ and $N=l+n+1$ if $n<0$. If $M>N$, the result is empty. Otherwise it consists of all items from position $M$ to position $N$ inclusive; for the purpose of this rule, we imagine that the token list extends at infinity on either side, with empty items at positions $s$ for $s\le0$ or $s>l$.

Spaces in between items in the actual range are preserved; spaces at either end are instead removed.

Thus, with $l=7$ as in the examples below, all of the following are equivalent and result in the whole token list
```latex
\tl_range:nnn { abcd~{e{}}fg } { 1 } { 7 }
\tl_range:nnn { abcd~{e{}}fg } { 1 } { 12 }
\tl_range:nnn { abcd~{e{}}fg } { -7 } { 7 }
\tl_range:nnn { abcd~{e{}}fg } { -12 } { 7 }
```
Here are some more interesting examples. The calls
```latex
\iow_term:x { \tl_range:nnn { abcd{e{}}fg } { 2 } { 5 } }
\iow_term:x { \tl_range:nnn { abcd{e{}}fg } { 2 } { -3 } }
\iow_term:x { \tl_range:nnn { abcd{e{}}fg } { -6 } { 5 } }
\iow_term:x { \tl_range:nnn { abcd{e{}}fg } { -6 } { -3 } }
```
are all equivalent and will print `bcd{e{}}` on the terminal; similarly
```latex
\iow_term:x { \tl_range:nnn { abcd~{e{}}fg } { 2 } { 5 } }
\iow_term:x { \tl_range:nnn { abcd~{e{}}fg } { 2 } { -3 } }
\iow_term:x { \tl_range:nnn { abcd~{e{}}fg } { -6 } { 5 } }
\iow_term:x { \tl_range:nnn { abcd~{e{}}fg } { -6 } { -3 } }
```
are all equivalent and will print `bcd {e{}}` on the
terminal (note the space in the middle). To the contrary, 
`\tl_range:nnn { abcd~{e{}}f } { 2 } { 4 }`
will discard the space after `d`.

If we want to get the items from the third to the last, the call
is `\tl_range:nnn { <tl> } { 3 } { -1 }`. Similarly, for discarding
the last item, we can do `\tl_range:nnn { <tl> } { 1 } { -2 }`.

blefloch:
As far as I can tell reading quickly this looks correct.

Thinking some more it is really strange to drop leading/trailing spaces
in something like

\tl_range:nnn { ~ a ~ b ~ } { -10000 } { 10000 }

because as you describe one should think of it as picking up all items
in some token list ...{}{}{}{}~a~b~{}{}{}{}{}... (without keeping these
braces).

eg9:
@blefloch I made a few fixes to cope with spaces at either end.

PR #501, opened 2018-12-02
Move \prg_generate_conditional_variant:Nnn to stable

josephwright:


wspr:
This provides consistency with what we have and rounds out `\cs_generate_variant` and `\prg_set_conditional`.

The only minor point I'd raise in objection is that if we ever wanted to change the name of `\cs_generate_variant` (which was discussed at one point, maybe?) then it would good to do it now :)  On balance, though, I'd probably vote against any name change there.

blefloch:
> - Move `\prg_generate_conditional_variant:Nnn` to stable

This CHANGELOG entry is in the wrong place since the pull request was delayed a lot.

josephwright:
@blefloch That's not an issue: I expect a bit of 'tidy up' with these PRs (as they tend to sit for a while). It's mainly the concepts I'm hoping to pin down.

blefloch:
Ok.
Another minor issue with this pull request is that it would be nice to
improve the doc of \prg_generate_conditional:Nnn.  I wrote it but I find
it somewhat cryptic since it refers to language defined in several other
commands.

PR #502, opened 2018-12-02
Move \tl_set_from_file:Nnn to stable

josephwright:
Also the shell function: a pseudo-file ...


wspr:
I have no concerns with this one. It's proven to be useful in a few places (I believe) and "deprecates" the catchfile package.

josephwright:
Possibly these are better as `\file_get:nnN` or similar: that matches other variable types. Look at re-naming/re-ordering arguments and moving to stable that way.

Issue #518, opened 2019-01-11
\box_set_eq_clear:NN, \box_gset_eq_clear:NN falsely claimed to clear box globally


gabor-braun:
Version 2018-12-12 of the LaTeX3 interfaces document claims (emphasis are my addition):

> \box_set_eq_clear:NN \<box1> \<box2>
> Sets the content of \<box1> within the current TEX group equal to that of \<box2>, then
> *clears \<box2> globally*.
>
> \box_gset_eq_clear:NN \<box1> \<box2>
> Sets the content of \<box1> equal to that of \<box2>, then clears \<box2>. *These assignments
> are global*.

I have tested this with the following example under plain TeX with pdftex (a slight variation fo the example at `\box_use_drop:N`):
```tex
\input expl3-generic.tex
\ExplSyntaxOn
\hbox_set:Nn \l_tmpa_box { a }
\group_begin:
  \hbox_set:Nn \l_tmpa_box { b }
  \group_begin:
    \box_set_eq_clear:NN \l_tmpb_box \l_tmpa_box
  \group_end:
  1:~ \box_use:N \l_tmpa_box \par
\group_end:
2:~ \box_use:N \l_tmpa_box

\newpage

\hbox_set:Nn \l_tmpa_box { a }
\group_begin:
  \hbox_set:Nn \l_tmpa_box { b }
  \group_begin:
    \box_gset_eq_clear:NN \g_tmpb_box \l_tmpa_box
  \group_end:
  1:~ \box_use:N \l_tmpa_box \par
\group_end:
2:~ \box_use:N \l_tmpa_box

\ExplSyntaxOff
\end
```
The created PDF looks like (omtting page number):
```
1:
2: a
3:
4: a
```
If `\l_tmpa_box` were cleared globally, the `a` shouldn't be there.

A formulation like that of `\box_use_drop:N` is a better description of the observed behaviour:
*\<box2> is cleared at the group level it was set*.

Log file:
```log
This is pdfTeX, Version 3.14159265-2.6-1.40.19 (TeX Live 2019/dev/Debian) (preloaded format=pdftex 2018.12.21)  11 JAN 2019 13:58
entering extended mode
 restricted \write18 enabled.
 file:line:error style messages enabled.
 %&-line parsing enabled.
**\input box-set-eq-clear.tex
(./box-set-eq-clear.tex
(/usr/share/texlive/texmf-dist/tex/latex/l3kernel/expl3-generic.tex
[...]

PR #522, opened 2019-01-17
Joseph's "clear drop" changes

wspr:
I've created this pull request not because I necessarily think it's ready to merge (I haven't reviewed yet) but because it makes it much easier to discuss!


gabor-braun:
Very good, this seems to solve #518, I've reported.
I like the new descriptions.

The following in the initial notes on drop vs clear can be easily misunderstood as "the dropping functions are only relevant for local boxes."
(Bold is my addition.)

>it is also useful to have functions which \emph{drop} box
>contents on use. **This consideration *only applies to local boxes*.**
>When such a box is dropped, the box becomes empty at the group level

What about writing something like this instead? (Bold is only to highlight changes.)

>it is also useful to have functions which \emph{drop} box
>contents on use. **For global boxes, dropping is the same as clearing.**
>When a **local** box is dropped, the box becomes empty at the group level

josephwright:
@gabor-braun  Didn't I cover that in later commits?

blefloch:
> @gabor-braun Didn't I cover that in later commits?

I just checked because some commits had confused me.  You addressed this
properly I think.

gabor-braun:
Sorry for the noise, I was not up to date, my mistake.  The latest version looks good.

Issue #521, opened 2019-01-17
Change the mode of \<module>_use:N and friends


wspr:
Recent (upcoming) changes will be changing the way checking of variables is done for some of the various `_use:` functions — the idea being that in production code we usually want speed and we shouldn't be presenting the user with internal error messages at all, so it doesn't matter if they're pretty expl3 ones or low-level TeX ones. So package development can be done in debugging mode, to help the programmer, and for most users the checking will be disabled and the code will be slightly more efficient.

This has led to some comments:

* `\tl_use:N` is something of an exception in that most other `_use:` functions use `\exp_not:n` internally to safely allow their use inside expandable contexts. 

* `\int_use:N` and other wrappers around `\the` are implicitly documented to expand in a single step, and this means we can't add any sort of checking to them. 

* Checking is only done sparingly and could be implemented on many more functions to help clean up after errors (e.g., Bruno points out that `\int_add:Nn \undefined {3}` will typeset `3` if compilation proceeds past the error message.)

* `\box_use:N` has rather a different operation than all of the other `use` functions; is it correctly named?

Which has led to some suggestions:

1. Add `\exp_not:n` to `\tl_use:N`. This doesn't stop people from using ‘naked’ tl variables to expand freely if they wish and provide more of a raison d'etre for `\tl_use:N` in the first place.

2. Provide a more low level register-expansion function which is a renamed `\the`, and remove the constraint that `\int_use:N` and others return in one expansion.

3. When checking is enabled (only), try harder to exit cleanly from what are currently fairly low-level operations.

* * *

There are mild compatibility concerns with 1 and 2 above. There are diminishing returns for spending effort on 3. I'm inclined not to change `\box_use:N`.

What are people's thoughts?

blefloch:
Thanks for the very clear summary!

I think some users have diligently used \tl_use:N \l_..._tl, so they
will be quite unhappy if suddenly \tl_use:N becomes equivalent to
\exp_not:o.

Instead of \exp_args:No \func:n { \int_use:N #1 }, which is a typical
situation where the precise expansion behaviour is needed, one can use
\exp_args:NV \func:n #1, somewhat slower.  Same for dim, skip, muskip.
For integers specifically we have \int_value:w, which is the primitive
\number and expands in one step.  So this suggests we may be able to get
rid of the constraint.

\box_use:N is not a great name but maybe worth a separate discussion?
We'd need to find a word (perhaps "typeset" or "print") for "insert in
[...]

PR #523, opened 2019-01-18
Adding \cs_use:N and a change to how "checking" works

wspr:
Not quite yet ready to merge but getting close. Also see #521 

I would expect this to be a squash merge.

### DIFF
--- a/l3kernel/l3basics.dtx
+++ b/l3kernel/l3basics.dtx
@@ -739,6 +739,20 @@
 %   \end{verbatim}
 %   after two expansions of \cs{use:c}.
 %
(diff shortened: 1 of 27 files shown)


wspr:
Apologies this has dragged on so long. Trying to come back to it. I'm about to make the change that `\int_use:N` and so on are not guaranteed to expand in a single step (when checking is on). Previous discussion suggested this might be fine as we have other commands like `\int_value:w` that can stand in.

josephwright:
On reflection, I think a change of this nature is likely a bit too much at this stage. I'm also not keen on lots of debug code buried inside the 'routine' source files. So I think we will leave this alone.

PR #524, opened 2019-01-20
Support for global coffins

### LINKED ISSUE #520: Implement global coffins
Mainly a question of fiddling with internals

josephwright:
This generalises all of the coffin mechanisms to allow them to be used in global scope. This fixes #520.

As we need property lists to do this, scope for coffins is tracked. That can only be set up once some kind of 'set' operation happens, so there are a number of places we have fall-back.

We may want to think about join/attach: I wonder in particular about "gjoin".


josephwright:
The test failure points out that I'd forgotten affine transformations, which are currently in candidates not in the core. Currently, such 'manipulations' are only local in the box code too. Probably we need to think first about whether we should have _e.g._ `\box_grotate:Nn` before I try to fix the coffin code. As such, I'm going to adjust the test file to get it to pass (the failure arises as the rotated coffin wasn't really set up correctly anyway).

josephwright:
Looking quickly over the box module, it's clear that only 'core' operations are potentially-global. One might argue the same for coffins: keep `g...` for `set`/`clear`/`set_pole(?)` but not for `join`/`attach` or affine stuff or .... Worth talking about first before I go any further.

josephwright:
Also for consideration: should coffins have scope declared explicitly? That would make life a lot easier (all props declared in advance), but would need `\coffin_new_global:N` or similar. Might also impact on current `\coffin_new:N`  - we might have to deprecate. 

If this sounds useful, I'll re-work the code (probably starting from here and stripping out the 'auto-magic').

blefloch:
About the question of auto-global: couldn't the coffin variable name
suffice to decide whether it's local or global?  We could decide that
variable names that don't start with l_ or g_ give local coffins.  There
us some logic in the "debug" code in l3basics.


About the unavoidably tedious (and slow) code that gets the scope to
build the poles prop name:

+          {
+            \tl_use:c { g_@@_scope_ \@@_to_value:N #1 _tl }
+            _@@_poles_ \@@_to_value:N #1 _prop
+          }

we should maybe decide to get away with badly-named prop variables, a
bit like messages are stored in the control sequence "msg~text~>~#1/#2".
 Maybe "coffin~\@@_to_value:N #1~poles"?

josephwright:
@blefloch We can't just use the coffin name: that only works if you restrict them _purely_ to formally-named `expl3` constructs. See for example `xbox`, where I use coffins as the 'back end' for document-level boxes. Or indeed see the test files ...

On the naming, yes it might be easier to call them something non-standard.

blefloch:
How hard would it be to provide global versions of the box affine
transformations?

josephwright:
I can happily generalise all box operations: more a question of if we want to.

josephwright:
I'm going to try another tack (to be opened shortly): I'll delete the current branch and start from some of @blefloch's ideas. (It's going to be easier to start from scratch for that.)

PR #525, opened 2019-01-21
Support for global coffins

josephwright:
As for https://github.com/latex3/latex3/pull/524, the aim is to support global coffins. Here, the plan is to start by making the coffin storage a bit easier to work with in scope terms. That then allows support for global coffins in a rather clearer way.


josephwright:
Everything should now be in place: are we happy for this to go ahead?

FrankMittelbach:
Am 23.01.19 um 11:54 schrieb Joseph Wright:
> Everything should now be in place: are we happy for this to go ahead?

I guess so ... let's test what it blows up in the wild :-)

PR #532, opened 2019-02-10
Fix malformed l3prefixes.csv [ci skip]

yudai-nkt:
As of 74cbdc7336da5c3ea9d2770daecbb6138ab72498, GitHub claims that `l3kernal/l3prefixes.csv` has an error:

> We can make this file [beautiful and searchable](https://help.github.com/articles/rendering-csv-and-tsv-data/) if this error is corrected: It looks like row 34 should actually have 10 columns, instead of 9. 

I wrote [a naive Lua script](https://gist.github.com/yudai-nkt/789ce02de2b0927a1f59106e3310d87c) to validate the CSV, and fixed the errors found.


blefloch:
Thanks!  Merged.

Issue #530, opened 2019-02-08
Enhancement: \bool_case:n and \bool_case:nTF in expl3


shmuelmetz:
I would like a case statement similar to the if/elseif/else/endif constructs of various other languages, e.g., SELECT/WHEN/OTHERWISE/END in REXX, as discussed at https://tex.stackexchange.com/questions/473979/is-there-latex3-work-being-done-on-a-boolean-case-statement. The preferred syntax would be something like
<pre>
\bool_case:n
  {
    {test 1} {action 1}
    ...
    {test n} {action n}
  }
</pre>
and
<pre>
\bool_case:nTF
  {
    {test 1} {action 1}
    ...
    {test n} {action n}
  }
  {
  {match-any-action}
  {no-match-action}
</pre>

tail-reversion:
I had wondered about requesting this, so I did some looking and discovered that it had been discussed on the LaTeX-L mailing list. The reasoning given there for not providing a `\bool_case:n` was that there are already specific case functions for integers, token lists, strings, and dimensions, which covers most use cases. Since expl3 is oriented around typesetting needs, these specific functions will tend to be more common than complex boolean statements anyways.

That being said, I think that adding `\bool_case:n` could be useful… Though I’m curious to see an example of a use case, because I realized I couldn’t come up with one that couldn’t be easily restated in terms of the existing functions.

wspr:
Here's an example implementation. Criticism welcome. I mildly support adding it to the kernel as it fits the pattern of the other modules and has occasional use cases.

I also think adding a construct for "jumping ahead" would be useful in cases where you have multiple booleans that are NOT expandable.

```
\documentclass{article}
\usepackage{expl3}
\begin{document}
\ExplSyntaxOn
\quark_new:N \q_bool_case_end
\cs_new_protected:Npn \__quark_bool_case_end: #1 \q_bool_case_end {}
\cs_new_protected:Npn \bool_case:nTF #1 #2 #3
  {
    \tl_if_empty:nT {#1} { \ERRORRR \__quark_bool_case_end: }
    \int_if_odd:nT { \tl_count:n {#1} } { \ERROR \__quark_bool_case_end: }
    \__bool_case:nnnn {#2} {#3} #1 \q_nil \q_nil
    #3
    \__quark_bool_case_end:
    \q_bool_case_end
  }
\cs_new_protected:Npn \__bool_case:nnnn #1 #2 #3 #4
  {
[...]

PR #533, opened 2019-02-10
Add \bool_case:nTF in a branch

blefloch:
This family of four functions is requested by #530 (maybe I should have included that in the commit message).

But it adds one more candidate function when we are trying to get fewer of these.


wspr:
Looks good to me. [ And no worries about discarding my old implementation :) ]

eg9:
Seems good!

wspr:
I forgot to mention also that although alternative names have been proposed, I’m happy with `bool_case` since there is a pretty clear similarity in behaviour / usage to the other `_case` functions.

blefloch:
The main objector to this name was Frank, so maybe he has an opinion now?

FrankMittelbach:
> The main objector to this name was Frank, so maybe he has an opinion now?

I guess I'm fine with it too, still think it is different from the other case functions but as @wspr said probably not that much to use a different name and anyway we never had a good suggestion for a name. So let's use this one.

tail-reversion:
This is probably a useless note… The Lisp languages have a macro that works just like `\bool_case:n`, called `cond`.

car222222:
I guess that a usage something like 
      \bool_case  \c_true_bool 
would make it into a faithful analogue of the other _case cases.

Which makes me think that, since booleans are both symmetric and binary in nature, we need these two cases:

   \bool_case_true   and     \bool_case_false 

The second will select the first false case: I agree that in this case the meaning of the T and F branches gets very very confusing!

An extension of this idea would be to allow a full analogy to the other cases by providing the more general

   \bool_case:nn  {<boolean expr>}  { list of  (boolean expressions ,  code) }

Which reminds me that it is important that all these case statements only use the code from the first match.  All subsequent matches are ignored.

blefloch:
On 2/16/19 07:11, Chris Rowley wrote:
> I guess that a usage something like
> \bool_case \c_true_bool
> would make it into a faithful analogue of the other _case cases.
> 
> Which makes me think that, since booleans are both symmetric and binary
> in nature, we need these two cases:
> 
> \bool_case_true and \bool_case_false
> 
> The second will select the first false case: I agree that in this case
> the meaning of the T and F branches gets very very confusing!

I really like Chris' suggested name so I've implemented that, still in
the bool_case branch.  If we decide to use the names
\bool_case_true/false then the two commits on that branch should
probably be squashed to make all traces of \bool_case:n(TF) disappear.
For now I've kept both commits as we may still decide we prefer the name
\bool_case:nTF.

josephwright:
Overall balance here seems to be positive: should I fix the conflicts and merge?

wspr:
Yes, thanks — good for me.

FrankMittelbach:
fine by me

PR #536, opened 2019-02-13
Move coffin affine transformations to stable

josephwright:


wspr:
I don't see anything controversial here... okay with me!

FrankMittelbach:
Are there test files executing the new global coffin functions and the interaction between local and global?

josephwright:
@FrankMittelbach  Not at present. That's separate (I think) from this PR, but I'll add some. What sort of tests are you thinking of?

FrankMittelbach:
@josephwright don't think this is separate. We should make it a point to only add new stuff to the core if we already supply tests for it (doing it later is like doing proper documentation later .. seldom happens :-) ). 

anyway what tests: basic functionality (probably a copy of those for local coffins) plus some level of weird mixture to ensure that there are no dependencies between local and global anywhere in the data structures.

josephwright:
@FrankMittelbach  Test file update on its way

josephwright:
Recent fixes for scope issues mean I'll integrate this by a new commit: closing here.

PR #537, opened 2019-02-15
Drop \tl_count_tokens:n/\tl_reverse_tokens:n

josephwright:
The ideas are likly good but these have been in candidates for some time without any usage. I'd rather we dropped at present to allow more clarity in `l3kernel`.


eg9:
I have two use cases of `\tl_count_tokes:n`

https://tex.stackexchange.com/a/408942/4427
https://tex.stackexchange.com/a/399733/4427

josephwright:
@eg9  I'm not doubting that there are potential use cases. I'm just trying to tighten up `l3kernel` to be sure we've reviewed stuff. At present, we've not seen these functions needed in anything in release code. I'm happy to see them stay if that's the decision: really the key is to move things out of `l3candidates`.

car222222:
How do these differ from \tl_count:n  and  \tl_reverse:n ?

eg9:
@car222222 `\tl_count:n { ab~c{def} }` yields 4, whereas `\tl_count_tokens:n { ab~c{def} }` yields 9.

wspr:
The rest of the tl module is fairly high level. I'd be okay with not merging these into the kernel for now. I do wonder what we should do with the code though. If we delete it's rather hard to unearth again, despite git.

Perhaps an unreleased package in l3trial called l3outtakes or something?

josephwright:
We have `l3kernel-extras`, but the concepts here are not that hard to implement so it wasn't a major concern.

blefloch:
I don't mind if \tl_count_tokens:n and \tl_reverse_tokens:n are removed.

If we want to keep them accessible to people, maybe it would make sense
to put l3kernel-extras in l3experimental to distribute it?  I haven't
taken a look in a long time, so I don't know whether it is worth it.
There are a few strange ideas there (like ugly expandable warnings).

josephwright:
We have a plan here, but not using this PR as-is.

PR #544, opened 2019-02-26
Move \prop_count:N to stable

josephwright:


wspr:
The function is good but reading the docs I wonder if we should move away from “key value pairs” in relation to property lists. It’s a bit confusing since we also have “real” keyval. 

Probably not the best place to discuss or address but I might forget to mention again :-) 

In short, happy to merge. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

blefloch:
On 2/27/19 12:52, Will Robertson wrote:
> The function is good but reading the docs I wonder if we should move
> away from “key value pairs” in relation to property lists. It’s a bit
> confusing since we also have “real” keyval.

I don't see any other pair of words that would fit.

Bruno

PR #545, opened 2019-02-26
Remove \skip_split_finite_else_action:nnNN

josephwright:


wspr:
I agree with Joseph this is a bit weird unless there are compelling use cases :)

FrankMittelbach:
There have been in the xor code if I remember correctly. However, the name is not really good/correct and if needed it can be reinvented again. Maybe it should moved into the old xor code just for reference (if my memory is not wrong about where it was used)

blefloch:
On 2/27/19 14:29, Frank Mittelbach wrote:
> There have been in the xor code if I remember correctly. However, the
> name is not really good/correct and if needed it can be reinvented
> again. Maybe it should moved into the old xor code just for reference
> (if my memory is not wrong about where it was used)

grep confirms this, with a delightful use of

    {\global\g_xor_float_text_shrink_dim}

as an N-type argument that is supposed to be a dim variable.  This
presumably works, somewhat by accident :).

josephwright:
I've simply put a copy in `xor`. There really isn't anything complex in the code: if we want some simular idea, working out a new version will be trivial (and probably can be handled directly using `\skip_if_finite:n(TF)`).

PR #549, opened 2019-03-03
Add b-type argument (body of environment) to xparse

blefloch:
This does not include any implementation of verbatim environments.

It also fixes some error messages that used to say `... in command "\environment foo" ...` and now say `... in environment "foo" ...`.

Maybe since there are many commits it's better to squash the commits?


blefloch:
I'm a bit confused how to do that.  Do I simply rebase and squash
commits in my copy of the xparse branch and then force push to the
repository and make a new pull request?

wspr:
@blefloch Github tracks all changes on the branch, so I think you could `rebase -i` your commits then force push and the pull request would still be in sync. I wouldn't go overboard here — if you can squash the commits down a bit then it's probably worthwhile but anything that takes time would have diminishing returns...

josephwright:
@blefloch  I suggest after the meeting today, assuming we go for this, I just do the rebasing to give us a reasonably-clear history.

blefloch:
That would be super helpful.  The structure could be (I've moved commit
4bf753a from near the end to the "b-type argument" section, as I think
the changes do not overlap except in the CHANGELOG).

xparse documentation improvements [ci skip]

- 0e34832 Tighten up xparse doc [ci skip]
- 8ef0d54 Reorder xparse sections a bit [ci skip]

Implement b-type argument for grabbing the body of an environment

- 481b8ed Implement b-type argument for grabbing the body of an environment
- c5ff071 Remove unused integer variable in xparse
- 632ccf6 Correct a test file
- 9e8e3bc In xparse, add reference to environ package and clarify
nesting [ci skip]
- b79c761 Update Changelog
- 4bf753a Remove user-provided setup for xparse b-type arguments

Improve xparse messages, especially for environments

- 7d2a216 Make \IfBooleanTF safer (check its argument is true or false)
- 9ebfb77 Correct an xparse error message and clean up a few others
- 021f524 Tweak internals of xparse environments to track environment name
- 5d2552a Make IfBooleanTF error message be expandable
- bb50b33 Improve xparse messages, especially for environments
- e79c9c7 Avoid second error for environment with too many args

xparse environment names: spaces and stars
- 48eff3b Trim spaces at end of xparse environment names
- ed3a1aa (HEAD, origin/xparse, xparse) Document that xparse doesn't
treat starred environment names specially [ci-skip]

(I notice a mistake: should be "Trim spaces at **ends**")  Anyway, given
how often we have tiny commits in the main history, Will is right that
carefully squashing commits together has diminishing returns.)

PR #559, opened 2019-04-02
Move experimental file ops to stable

josephwright:


wspr:
Looks good, although I'm not sure of the use cases for `\file_input_stop:`. Is that one there for completeness or did someone need it?

josephwright:
@wspr People are used to `\endinput` at the end of their `.sty` files ....

eg9:
@wspr Sooner or later, any primitive (well, almost) will have an `expl3` equivalent. And that `\tex_endinput:D` sitting at the beginning of `siunitx.sty` is really awful looking. ;-)

blefloch:
I think that `\file_path_include:n` should be `\l_file_search_path_seq`, but I don't know `l3file` so well.

Maybe it would be simpler to describe the search as "Searches for \meta{file name} in the path as detailed for \cs{file_if_exist:nTF}", like we do in the doc of `\file_input:n`.

blefloch:
A few `str var` should be `tl var`, and "Where the file is not found, the \meta{str var} will be empty." seems not true anymore now that some people decided it would be better to return `\q_no_value`.

l3kernel/l3file.dtx
```diff
@@ -666,6 +709,40 @@
...
+%   an error if the file is not found, in contrast to \cs{file_input:n}.
+% \end{function}
+%
+% \begin{function}[added = 2017-07-07]{\file_input_stop:}
+%   \begin{syntax}
+%     \cs{file_input_stop:}
+%   \end{syntax}
+%   Ends the reading of a file started by \cs{file_input:n} or similar before
+%   the end of the file is reached. Where the file reading is being terminated
+%   due to an error, \cs{msg_critical:nn(nn)} should be preferred.
```
blefloch: \cs[index=msg_critical:nn]{msg_critical:nn(nn)} or similar to avoid messing up the index.

l3kernel/l3file.dtx
```diff
@@ -655,6 +655,49 @@
...
+%   \end{syntax}
+%   Searches for \meta{file name} using the current \TeX{} search
+%   path and the additional paths controlled by \cs{file_path_include:n}.
+%   If found, sets the \meta{str var} to the MD5 sum generated from the
+%   content of the file. The file is read as bytes, which means that in
+%   contrast to most \TeX{} behaviour there will be a difference in result
+%   depending on the line endings used in text files. The same file will
+%   produce the same result between different engines: the algorithm used
+%   is the same in all cases.
+%   Where the file is not found, the \meta{str var} will be empty.
```
blefloch: Where → If (and similar in other function descriptions)

l3kernel/l3file.dtx
blefloch: It might be useful to comment in this function why we made it protected rather than expandable (for consistency with other `file` functions, or a similar reason).

josephwright:
@eg9 That will _always_ have to be just `\endinput` or `\tex_endnput:D`: it's there to cover the case where someone has a really old `expl3` without `\msg_fatal:nn` ...

l3kernel/l3file.dtx
josephwright: Not happy with the grammar here?

FrankMittelbach:
> @wspr People are used to `\endinput` at the end of their `.sty` files ....

To be honest \endinput is in my opinion a sensible user level command. We can of course also provide an internal form for the sake of using _ names inside code rather than user level ones but whether that means you get people to write \file_input_stop: instead of \endinput at the end of their .sty files I wonder :-). Still it is consistent even though it means 3 names for one and the same thing.

wspr:
I’m a bit of a fool sometimes. Open up fontspec.sty and you will see:

    \sys_if_engine_xetex:T
      {
        \RequirePackageWithOptions{fontspec-xetex}
        \endinput
      }

So yes, I support adding it!

PR #560, opened 2019-04-03
Move box 'view' functions to stable

josephwright:


wspr:
Okay with me.

josephwright:
@davidcarlisle Thoughts?

blefloch:
Perhaps add cross-references from the trim to the viewport functions?  They are the basically same but with different descriptions of the coordinates, right?

PR #566, opened 2019-04-16
Add token counting argument specifier to xparse

zauguin:
Add a new argument type "`c`ount" for `xparse` which behaves like `t` arguments, except that instead of scanning *up to one* token, an *arbitrary number* of matching tokens is scanned and the number used as the argument.

For example

```
\NewDocumentCommand \primes {m c'} {{#1}^{(#2)}}
```
would result in `\primes{f}'''''` being expanded to `{f}^{(5)}`.

### DIFF
--- a/l3packages/xparse/xparse.dtx
+++ b/l3packages/xparse/xparse.dtx
@@ -174,6 +174,9 @@
 %   \item[t] An optional \meta{char}, which will result in a value
 %     \cs{BooleanTrue} if \meta{char} is present and \cs{BooleanFalse}
 %     otherwise. Given as \texttt{t}\meta{char}.
+%   \item[c] Multiple optional \meta{char}s, which will result in a number
+%     representing the number of found \meta{char}s. Given as
+%     \texttt{c}\meta{char}.
 %   \item[e] Given as \texttt{e}\marg{chars}, a set of optional
 %     \emph{embellishments}, each of which requires a \emph{value}.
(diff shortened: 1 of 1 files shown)


eg9:
I'm not really sure how many applications this argument type would have.

josephwright:
Like @eg9, I'm wondering about use cases: other than math mode prime, do you have some?

wspr:
I could imagine a couple of ad hoc uses but not much springs to mind in code I’ve seen in the past. I do like the way it generalises the t-type argument though.

car222222:
For math use-cases, the past is not a good guide.  New uses appear steadily.  For the example, using . instead of ‘ for time derivatives.  I am not supporting such usage (at least not this particular example) but you cannot argue that it should not be added because there are currently only a few use-cases.  Maths notation does not work like that, we are completely free to use any notation we want (after carefully defining its meaning in terms of ‘known notation and concepts. This is unlike Physicists usage, which tends to be more restricted by conventions, but also needs to evolve to cover new concepts.

car222222:
@zauguin Thanks for the non-maths use-cases.

blefloch:
I suggest adding a "regex" arg type to xparse and dealing with all
further requests for argument types by pointing them to this very
general tool.

josephwright:
I think we want to go for a regex type here if we explore more types: as it is, we already have several.

PR #569, opened 2019-05-03
Move \char_<target>_case:n and str versions to stable

josephwright:


blefloch:
Would it be better if \char_lower_case:N (and others without \char_str_ in the name) used their argument's catcode rather than the currently active catcode?  With the current code one can end up with \char_generate:nn throwing an error.

josephwright:
Test failures here are to do with `pgf` changes that impact on `contrib`: they are handled in the main branch and I'm reluctant to rebase at this stage ... I'll do that if we go for the plan.

PR #573, opened 2019-05-10
Missing (meta) token list variable added

dbitouze:


josephwright:
I don't see this: the sigs are `Nnnn` and `Nn`, respectively.

dbitouze:
Sorry for the noise: my bad.

PR #580, opened 2019-05-27
Use \luafunction or \luadef for \str__if_eq:nn

zauguin:
This provides significant speedups for modern LuaTeX versions. There are some TODOs:

- [ ] How should the allocation of `\luafunction` ids be handled in `expl3`? Currently it just uses `\newluafunction` which will only work if `ltluatex` is loaded.
- [ ] Also `\tex_luadef:D` and `\tex_luafunction:D` should get "real" names. Probably they belong into the `\lua_` module but at least without allocation support they might be more appropriate as local interfaces. Are l3kernel modules allowed to use local functions from other modules?
- [ ] Do the same for the other functions in `expl3.lua`.

### DIFF
--- a/l3kernel/l3bootstrap.dtx
+++ b/l3kernel/l3bootstrap.dtx
@@ -230,6 +230,9 @@
       local charcat_table = \number\ucharcat@table\space
       l3kernel.charcat_table = charcat_table
     }%
+    \expandafter\newluafunction\csname
+      c__str_cmp_func_int%
+    \endcsname
 %</package>
     \directlua{require("expl3")}%
 %    \end{macrocode}
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -314,6 +314,11 @@ local tex     = tex
 local unicode = unicode
 %    \end{macrocode}
 %
+%   Local copy of the Lua function table
+%    \begin{macrocode}
+local luafunctions = lua.get_functions_table()
+%    \end{macrocode}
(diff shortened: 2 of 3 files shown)


josephwright:
On point (1): we already require `ltluatex` in generic mode, so the only 'allocator' issue is for a (future) stand-alone format. That can easily be fixed by adding support akin to `ltluatex` to `l3alloc`: probably we should do that, it's just a question of naming.

josephwright:
Actually, the allocator links to the entire 'what is the public name' business: presumably we want `\lua_function_new:N` or similar. Have to decide on a name to define an allocator.

zauguin:
@josephwright Right, but I am not sure what the semantics of `\lua_function_new:N` should be: Is the name it defines a integer representing the function id as with `\newluafunction`? This would be very compatible, but I think the interface would be nicer if it would directly define a cs with `\luadef`. Then there would not be the weird duality of the actual macro and this odd function index. On the other hand, this could not really be ported to older versions, so it would have to be a LuaTeX 1.10+ only thing.

josephwright:
I'm going to open a new issue for discussion on the set up itself.

I suspect once that is sorted, it might make most sense to move the re-definition here to `l3luatex`, as we presumably want the mechanisms in place. But first ...

josephwright:
Just to say this isn't forgetten: I'll prod everyone!

josephwright:
Covered better in the new https://github.com/latex3/latex3/pull/796

PR #592, opened 2019-06-17
Move all debug code to allow loading after the main part of expl3

josephwright:


josephwright:
I might still not quite have the setup right for the `backend` files: I'll likely have to track down where I have the dependency wrong.

josephwright:
I'd thought about the business with logging/checking the kernel itself. We could have all of the debug enabled and turn it off as soon as we are loaded, but that's a bit of a pain. Also, whilst for us it might be 'nice' to have stuff in the `.log`, it's not so great when using `expl3` and debugging a package: there's a _lot_ of redundant stuff first. 

Perhaps we could add some DocStrip guards for a 'really debug the kernel itself' extraction? That could then be tested separately.

josephwright:
I've not really worried about catcodes just yet. I'd rather not go down the Heiko 'defensive' route. Probably I need to finish the `cctab` module, then we can have some simple set up inside `\ExplSyntaxOn` which does everything correctly. That's better than having lots of explicit checks, I think.

blefloch:
About catcodes: we don't need to be nearly as defensive as Heiko,
because it is expl3 code that loads the files l3debug.def: just include
\cctab_use:N \c_ExplSyntaxOn_cctab (or whatever the correct names are)
in the code of \__kernel_sys_configuration_load:n.  This requires a
group, but I think loading l3debug, l3deprecation, and the backend files
in a group can be made to work.

blefloch:
> Perhaps we could add some DocStrip guards for a 'really debug the kernel
> itself' extraction? That could then be tested separately.

I'm wondering if we could set things up such that m3expl001 etc only
need to predefine a few functions; then at the appropriate places in
l3kernel use (the equivalent of) \providecommand.  After all, these test
files are the only places in which we really want to check the kernel.

josephwright:
@blefloch Yes, if we do it by `x`-type expansion such that for normal usage nothing is left to impact on performance. We just have to do a low-level `\tex_ifdefined:D` test I guess on something like `\__kernel_log_debug:n`?

josephwright:
@blefloch  One of the questions with the entire catcode table business is how best to set it up. One way or the other we can make it workable, I'm sure: the files are likely workable in a group, but we should also have 'push'/'pop' for the tables anyway. At present everything is very much a sketch in any case: I wanted to make sure it's workable. For example, I'm still not really sure about how the debug code should load: a package or a `.def` file. Similarly, we might want to make the backend files reloadable so a default is always available in generic mode. So again that would address the grouping I think.

blefloch:
Thanks Joseph.  I won't be available for LaTeX stuff in the foreseeable
future (way too much work), so please decide without my input.

PR #628, opened 2019-09-18
l3doc.cls: fix bug when trying to read l3doc.cfg

frougon:
Don't use `\file_input:nT`, which has been removed in commit ecd4174da983941f30d3231e44b526c079d674ec. Use `\file_input:n` instead.


PhelypeOleinik:
You're right. Thanks!

Although I think that the test `\cs_if_exist:cF { ExplMakeTitle }` to print the `input-cfg` message is not ideal. I think another (private) control sequence should be tested here. I'll add to my todo list.

frougon:
Thank you, Phelype!
Regards

PR #633, opened 2019-09-29
Allow xparse processors to depend on other arguments (fixes #629)

### LINKED ISSUE #629: [xparse] Make value reference to other arguments accessible for argument processors
First of all, many thanks for the excellent work on LaTeX3. I've seen recently that default values for optional arguments can refer to other arguments. I wanted to know if this functionality could be transferred to argument processors as well. The following example is only intended to demonstrate the desired procedure and would not be necessary for this simple case.
```tex
\documentclass{article}
\usepackage{xparse}
\makeatletter
\NewDocumentCommand\TestLabels{o m d==}{%
  \Test@Labels[#1]{#2}=#3=%
}
\NewDocumentCommand{\Processor@Labels}{m m m}{%
  \IfValueTF{#1}%
    {\def\ProcessedArgument{#1:#3}}%
    {\def\ProcessedArgument{#3}}%
  \IfValueT{#2}
    {%
      \expandafter\def\expandafter\ProcessedArgument%
        \expandafter{\ProcessedArgument:#2}%
    }%
}
\NewDocumentCommand\Test@Labels{r[] m r==}{%
  \Processor@Labels{#1}{#3}{#2}%
  \texttt{\ProcessedArgument}
}
%%% requested functionality
%\RenewDocumentCommand\Test@Labels{r[] >{\Processor@Labels{#1}{#3}}m r==}{%
%  \texttt{#2}
%}
\makeatother
\begin{document}

\TestLabels{base}

\TestLabels[prefix]{base}

\TestLabels{base}=suffix=

\TestLabels[prefix]{base}=suffix=

\end{document}
```

blefloch:


josephwright:
Ah, now I see an example 'in the wild' this looks good: I'm happy to go with it.

blefloch:
I think the test failures happen before xparse even shows up, so it cannot be my changes.

josephwright:
Failures are to do with an `l3build` bug ...

josephwright:
I've just double-checked the failures: the one with `ctex` does seem to be 'real', and I don't get it from the parent commit (see now-fixed Travis-CI builds for example). Using a local run, I get:
```
File: ctex-scheme-chinese-book.def 2019/05/29 v2.4.16 Chinese scheme for book (
CTEX)

! LaTeX Error: Missing \begin{document}.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...

l.153   }
```

qinglee:
A MWE is:

    \RequirePackage{xparse}

    \ExplSyntaxOn

    \NewDocumentCommand \TEST
      { m > { \SplitArgument { 1 } { , } } m }
      { \tl_show:n {#2} }

    \TEST { 1 } { test }

    \NewDocumentCommand \TESTb { } { }

    \TEST { 1 } { test }

    \tex_end:D

I get the same error:

    > {test}{-NoValue-}.
    <recently read> }
                     
    l.9     \TEST { 1 } { test }
                                
    ? 
    
    ! LaTeX Error: Missing \begin{document}.
    
    See the LaTeX manual or LaTeX Companion for explanation.
    Type  H <return>  for immediate help.
     ...                                              
                                                      
    l.13     \TEST { 1 } { test }
                                 
    ? 
    Missing character: There is no t in font nullfont!
    Missing character: There is no e in font nullfont!
    Missing character: There is no s in font nullfont!
    Missing character: There is no t in font nullfont!
    Missing character: There is no t in font nullfont!
    Missing character: There is no e in font nullfont!
    Missing character: There is no s in font nullfont!
    Missing character: There is no t in font nullfont!
    > {1}{-NoValue-}.
    <recently read> }
                     
    l.13     \TEST { 1 } { test }
                                 
    ?

And please note that the results of the `\TEST` are not consistent.

blefloch:
Thank you very much, this helped me pinpoint the issue: to allow
processors to depend on other arguments I define an auxiliary that takes
the same number of arguments as the command under consideration.  I was
using the wrong number (the number of arguments of the last defined
command).  Now I think it is fixed.

I realized another issue (not specific to this branch): if someone uses
an xparse command \foo (with processors?) in the definition of some
processor, then \ProcessedArgument and a few other internal variables
get clobbered by \foo, which prevents the processor from working.  Not a
major problem.

PR #678, opened 2020-02-20
implemented an expandable and about 40% faster keyval_parse

Skillmon:


josephwright:
One simple change up-front: we don't need `[internal]` for any of the `macro` environments here as they are auto-detected. Also, given there is a change in the message, I think there should be a `.tlg` file change in the pull request.

Skillmon:
I wasn't sure whether I should update the `.tlg` pre-review. I thought this way the reviewer can more easily see the `diff`, my bad.

Skillmon:
Also, please note that this implementation places a few limitations on the possible contents of keys and values. It is not perfectly safe to place `\q_mark`, `\q_stop`, or `\q_recursion` in the value (depending on where they are put and whether they are used adjacent to each other.

PhelypeOleinik:
> Also, please note that this implementation places a few limitations on the
> possible contents of keys and values. It is not perfectly safe to place
> \q_mark, \q_stop, or \q_recursion in the value (depending on where they
> are put and whether they are used adjacent to each other.
>

There's an issue open about this already (I'm on my phone now, so no link,
sorry). The consensus is that we will eventually move to using private
quarks in kernel code to avoid this type of problem (I was hoping to get
that done in a few weeks).

I think it would be nice (and safer for existing code) if you switched to
the private quarks here already...

josephwright:
@Skillmon We've never had formal review, but I think the idea with pull requests is that they are worked on until they are 'ready to merge', so with things like tests passing.

Skillmon:
Since the meaning of the quarks isn't important in the current code, do you think not defining them is an option? This way they'll throw an error if they are erroneously expanded instead of creating an infinite loop.

josephwright:
I'd go for scan marks here, and to be honest we could move a lot of the kernel quark use that way. Scan marks are a lot newer than quarks in the kernel, so they are not used so much.

Skillmon:
@josephwright thanks for the standardising and logging. I was going to add the `tlg`s today, but you were faster :)

I have but one problem with your edits, there is a stray space in https://github.com/latex3/latex3/pull/678/commits/e5d04524ed9ed5442e936e2eb33adc036b92f259#diff-07972af5a6b29b12e79f928fd78f2084R1121

PR #701, opened 2020-04-09
New splitting implemented

Skillmon:
This pull request ports the changes made to `expkv` in v1.0 (modulo some corrections) to `l3keys`'s `\keyval_parse:NNn`. As a result this is 25%-35% faster than the current implementation.

This merge also fixes some minor (up to now) undetected bugs introduced with https://github.com/latex3/latex3/pull/678 and some inconsistent behaviour which was already observable with the version prior to that PR.

All these behaviour changes are related to the behaviour of empty or blank key names. This version will throw a `blank-key-name` error on all of the following inputs (except for the very first key-value list element which is completely blank, this case is still silently ignored as it should be):

```tex
\keyval_parse:NNn \use_none:n \use_none:nn
  {
    , {}
    , { ~ }
    , = value
    , {} = value
    , { ~ } = value
  }
```

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -939,13 +939,13 @@
 %
 % \subsection{Low-level interface}
 %
-% The low-level key parser is based heavily on \pkg{keyval}, but with a number
-% of additional \enquote{safety} requirements and with the idea that the
-% parsed list of key--value pairs can be processed in a variety of ways.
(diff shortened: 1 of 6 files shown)


Skillmon:
I'd like to add a few more deprecation tests here on which I'm currently working (which should prevent things like `{}=b` from slipping through in the future).

PR #714, opened 2020-04-29
Proposal package: high-level interface for metadata

tobiasdiez:
This PR proposes to add a package to consistently specify metadata of the document (e.g. author, title, emails, etc). It is inspired by https://tex.stackexchange.com/a/137387/34554. I remember that I've seen some early experiments concerning metadata here, but couldn't find the issue/PR.

The usage of the code looks similar to the following:
```latex
  \Title{A fairy tale}

  \Institution{earth}{Under the Earth, Mountain}
  \Institution{tree}{Between Trees, Forest}

  \Author[
	affiliation = earth,
	email = The.Dormant.One@snail-mail.mnt,
	funding = {Partially supported by the World Tree Yggdrasill.}
	]{dvalinn}{Dvalinn, One of the four Stags}
  \Author[
	affiliation = tree,
	email = Wide.Ruler@bird-mail.frs,
	funding = {Supported by leather pieces of shoes of various people.}
	]{vidarr}{Vidarr, Son of Odin}

\Keywords{Norse, Mythology}

\begin{Abstract}
	Once upon a time...well, not quite.
\end{Abstract}
```

It contains also the command `\MakeTitle` which displays the information. This is more for demonstration purposes and in the long term should be moved to a different package (to clearly separate specifying metadata and showing metadata). 

It's a bit of an early version and one might add many more fields to authors or intuitions (also subtitle, short title, etc).
What do you think?

### DIFF
--- a/l3experimental/l3meta/l3meta.sty
+++ b/l3experimental/l3meta/l3meta.sty
@@ -1,0 +1,341 @@
+% \begin{documentation}
+%
+% The \pkg{DocumentMetadata} package provides a high-level interface for specifying metadata.
+% 
+% The benefits of this package are:
+% \begin{itemize}
+%	\item Users need to learn only one interface.
+%	\item Remove the burden from document class writers to handle document metadata by providing a consistent interface.
+%	\item Document classes can be changed without modifying the way metadata is defined, 
+%		i.e. no need to customize it for different journals.
+% \end{itemize}
+%
+% At present, the following functions are provided:
+% \begin{itemize}
+%   \item \cs{Title}
+%   \item \cs{Author}
+%   \item \cs{Institution}
+% 	\item \cs{Abstract}
+%	\item \cs{Keywords}
+%	\item \cs{MSC}: Mathematics Subject Classification 2010
+% \end{itemize}
+%
+% \section{Simple example}
+% \begin{verbatim}
+%   \Title{A fairy tale}
+%
+%   \Institution{earth}{Under the Earth, Mountain}
+%   \Institution{tree}{Between Trees, Forest}
+%
+%   \Author[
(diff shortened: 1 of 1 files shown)


PhelypeOleinik:
Hi Tobias,

Thank you for your interest in contributing to the LaTeX project
however, I think this would fit better as a proposal for a document
class for a scientific journal, since it defines a strict format for
\MakeTitle (which is really a document class-specific thing) and adds
fields such as institution, e-mail address, and funding (for example,
if I were to write a novel, I probably wouldn't use any of those three).

Your proposal doesn't seem flexible enough to be adopted as a general
document metadata handling interface, which would require an interface
generic enough that it can be used regardless of document class, even
already existing ones, without changes in formatting.

We appreciate your effort, but this type of document design proposal is
out of the scope of the project.

tobiasdiez:
Thanks for the quick feedback.

I was perhaps a bit vague about the main purpose of the proposal: it was mainly concerned with the way metadata is specified, i.e. the commands `\Author` and `\Institution` instead of how it is displayed i.e `\MakeTitle`. Since almost every document needs some form of such metadata, I think a common interface would be a very valuable addition. For course, the way this metadata is processed and displayed, strongly depends on the type of document and should go into the document class or dedicated packages.

The above proposal is of course only a very minimal version and needs to be extended to be valuable for different types of documents.

By the way, I found the similar package that was included into Latex3: xmetadata. It was removed in https://github.com/latex3/latex3/commit/8d6a2950538c6a7df923fe8a643ff98658764115, but at least according to the commit message based on code-reasons and not because of the scope. Thus, I was hoping that such a metadata package would be acceptable for latex3.

PR #718, opened 2020-05-11
Split out 'core' xparse code

josephwright:
This splits the 'core' ideas out of `xparse`, i.e. removes g/l/u types so they are *not* preloaded using `xparse.ltx`. The result should be functionally equivalent to the 'cmd' branch, which is the longer-term plan to move this code into the kernel proper.


l3packages/xfp/testfiles/xfp000.tlg
```diff
@@ -3,10 +3,11 @@ Don't change this file in any respect.
 Author: Bruno Le Floch
 (xfp.sty (xparse.sty
 Package: xparse ....-..-.. L3 Experimental document command parser
```
FrankMittelbach, line 5: maybe time to change the package description? and drop "Experimental"? no?

FrankMittelbach: Some further thought on file names: should we go for l3parse and only retain xparse.sty as an historical exception?

u-fischer: @FrankMittelbach september last year we thought about `l3x..` for (user) packages and `l3somethingelse` for kernel stuff. so probably either `l3parse` or `l3xparse` depending on what this is now ...

FrankMittelbach: but we decided back that that we wouldn't change the name xparse because it is so widely used. My point of file names was more for the internal names in that case that get loaded by the kernel.

u-fischer: so l3parse.sty here.  Sounds good to me.

FrankMittelbach:
What we also need to do (but that doesn't need to be now for this -dev release) would be to alter the intro of main documentation to reflect that the majority of the functionality is now part of the kernel and that loading xparse is only needed to support running in older kernels or to make use of the (not recommended) extra specifiers. Maybe this should just be filed as an extra issue but not done now as it will take some effort.

josephwright:
> 
> 
> What we also need to do (but that doesn't need to be now for this -dev release) would be to alter the intro of main documentation to reflect that the majority of the functionality is now part of the kernel and that loading xparse is only needed to support running in older kernels or to make use of the (not recommended) extra specifiers. Maybe this should just be filed as an extra issue but not done now as it will take some effort.

Please see branch [cmd](https://github.com/latex3/latex3/tree/cmd), which was written before the current split plan but which sketches out exactly this idea.

PR #721, opened 2020-05-14
Add expandable pdfname generation

josephwright:
This is needed for some PDF contexts: we only need expandable creation for pdfnames.


l3kernel/l3str-convert.dtx
```diff
@@ -186,6 +186,19 @@
...
+% contexts. These \emph{assume UTF-8} and \emph{no escaping} in the input.
+%
+% \begin{function}[EXP]{\str_convert_pdfname:n}
+%   \begin{syntax}
+%     \cs{str_convert_pdfname:n} \meta{string}
+%   \end{syntax}
+%   As \cs{str_set_convert:Nnnn}, converts the \meta{string} on a byte-by-byte
+%   basis with non-ASCII codepoints  escaped using hashes.
+% \end{function}
+%
```
blefloch, line 201: I'm wondering whether there could be a need for more generally `\str_convert:nnn`, which is probably doable now that we have `e`-expansion.

l3kernel/l3str-convert.dtx
```diff
@@ -2567,6 +2580,102 @@
...
+         \exp_after:wN \@@_convert_pdfname:n \exp_after:wN
+           { \tl_to_str:n {#1} }
+      }
+  }
+\cs_new:Npx \@@_convert_pdfname:n #1
+  {
+    \exp_not:N \@@_convert_pdfname_first:w #1 ~
+      \exp_not:N \q_recursion_tail \c_space_tl
+      \exp_not:N \q_recursion_stop
+  }
```
blefloch, line 2612: There shouldn't need to be a loop over space-delimited words and then over characters within a word.  I expect that applying `\__kernel_str_to_other_fast:n` within `e`-expansion is just as fast.  Then you could simply have a loop like
```
\cs_new:Npn \@@_convert_pdfname:n #1
  {
    \@@_convert_pdfname_loop:n #1 
    { ? \prg_break: }
    \prg_break_point:
  }
\cs_new:Npn \@@_convert_pdfname_loop:n #1
  {
    \use_none:n #1                               % This avoids a slow quark test
    \@@_convert_pdfname_one:N #1
    \@@_convert_pdfname_loop:n
  }
\bool_lazy_or:nnTF
  { \sys_if_engine_luatex_p: }
  { \sys_if_engine_xetex_p: }
  { ... }
  { \cs_new_eq:NN \@@_convert_pdfname_one:N \@@_escape_name_char:N }
```
Also since the `scan-quark` branch was merged, don't use `\q_recursion_tail` etc.

l3kernel/l3str-convert.dtx
```diff
@@ -2567,6 +2580,102 @@
...
+        \quark_if_recursion_tail_stop:n {#1}
+        \int_compare:nNnTF { `#1 } > { "7F }
+          { \@@_convert_pdfname_bytes:n {#1} }
+          { \@@_escape_name_char:N #1 }
+        \@@_convert_pdfname_inner:w
+      }
+    \cs_new:Npn \@@_convert_pdfname_bytes:n #1
+      {
+        \exp_args:Ne \@@_convert_pdfname_bytes_aux:n
+          { \char_to_utfviii_bytes:n {`#1} }
```
blefloch, line 2645: This line reminds me that we probably have utf8 conversion code both as this function `\char_to_utfviii_bytes:n` that you used, and internally in `\str_set_convert:Nnnn`.  Maybe worth merging.

l3kernel/l3str-convert.dtx
u-fischer: @blefloch  we have been discussing this, on the whole I don't think that this is a pressing issue. names pop up in various arguments, so being able to expand them is quite usefull, but other strings can normally be saved and handled later.  More needed is an utf8/xmp conversion/escaping for the xmp-metadata.

l3kernel/l3str-convert.dtx
josephwright: Possibly: `\char_to_utfviii_bytes:n` remains experimental, partly due to the name, partly as I'm still not 100% sure about the form of the output

car222222:
Maybe ‘we’ do not need an expandable version for pdf strings right now, but for the great pdf-producing world out there, this would be a useful addition.  

I guess this is fairly easy to implement by adding an expandable conversion from the syntax for pdf names to that for pdf strings.

l3kernel/l3str-convert.dtx
blefloch: I just thought today the clearest by far would be to use `\str_map_function:nN {#1} \@@_convert_pdfname_one:N`.  It should be of comparable speed to the current version, and would benefit from any speed improvement in `\str_map_function:nN`.

l3kernel/l3str-convert.dtx
```diff
@@ -186,6 +186,19 @@
...
+% A small number of expandable functions are provided for use in PDF string/name
+% contexts. These \emph{assume UTF-8} and \emph{no escaping} in the input.
+%
+% \begin{function}[EXP]{\str_convert_pdfname:n}
+%   \begin{syntax}
+%     \cs{str_convert_pdfname:n} \meta{string}
+%   \end{syntax}
+%   As \cs{str_set_convert:Nnnn}, converts the \meta{string} on a byte-by-byte
+%   basis with non-ASCII codepoints  escaped using hashes.
+% \end{function}
```
blefloch, line 200: Document the precise arguments that need to be given to `\str_set_convert:Nnnn` to obtain the same conversion.

l3kernel/l3str-convert.dtx
```diff
@@ -2567,6 +2580,63 @@
...
+%   after which we use a simple token-by-token approach. In Unicode
+%   engines, we break down everything before one-byte codepoints, but for
+%   $8$-bit engines there is no need to worry. Actual escaping is covered
+%   by the same code as used in the non-expandable route.
+%    \begin{macrocode}
+\cs_new:Npn \str_convert_pdfname:n #1
+  {
+    \exp_args:Ne \tl_to_str:n
+      { \str_map_function:nN {#1} \@@_convert_pdfname:N }
+  }
```
blefloch, line 2601: Presumably `\tl_to_str:n` is not needed and simply `\use:e { \str_map... }` is enough.

l3kernel/l3str-convert.dtx
```diff
@@ -2567,6 +2580,63 @@
...
+          { \@@_convert_pdfname_bytes:n {#1} }
+          { \@@_escape_name_char:N #1 }
+      }
+    \cs_new:Npn \@@_convert_pdfname_bytes:n #1
+      {
+        \exp_args:Ne \@@_convert_pdfname_bytes_aux:n
+          { \char_to_utfviii_bytes:n {`#1} }
+      }
+    \cs_new:Npn \@@_convert_pdfname_bytes_aux:n #1
+      { \@@_convert_pdfname_bytes_aux:nnnn #1 }
```
blefloch, line 2618: `\exp_args:Ne` should be `\exp_last_unbraced:Ne` so that the auxiliary `..._aux:n` is unnecessary.

PR #722, opened 2020-05-14
Replace token expandably

Skillmon:
This pull request adds two functions:

- `\tl_replace_token:nNn { <token list> } <old token> { <new tokens> }`

- `\tl_replace_token_deep:nNn { <token list> } <old token> { <new tokens> }`

Both are fully expandable and replace every occurrence of `<old token>` with `<new tokens>` in the `<token list>` and leave the result in the input stream inside of `\unexpanded`. The difference is that the `_deep` variant also replace every `<old token>` inside braced groups recursively while the former variant will ignore groups.


blefloch:
This cannot work as "advertised": some tokens can have the same \meaning and \string but differ.
```
\tl_set:Nn \l_tmpa_tl { & }
\cs_set_eq:NN \tmp &
{
    \catcode`&=\active
    \cs_set_eq:NN & \tmp

    \tl_replace_once:Nnn \l_tmpa_tl & { new }
    \iow_term:x { \l_tmpa_tl } % gives "&"

    \iow_term:x { \tl_replace_token:oNn \l_tmpa_tl & { new } }
    % certainly gives "new"?
}
```

Skillmon:
@blefloch yes, edge case I didn't consider. But if I add an `\ifcat` test this should work.

blefloch:
No, these tokens are not distinguishable expandably.
Maybe that's an acceptable limitation though.

Skillmon:
Building the docs will fail currently because of the mid-year deprecations.

josephwright:
Looking back, I think we want to cover here with some putative 'traverse' function

PR #734, opened 2020-05-30
correct array writing in dvips mode, index missing

u-fischer:
Example document why this is needed:

~~~~
\RequirePackage{l3pdf}
\ExplSyntaxOn
\pdf_uncompress:
\ExplSyntaxOff
\documentclass{article}
\begin{document}
\ExplSyntaxOn
\pdf_object_new:nn{ar1}{dict}
\pdf_object_new:nn{ar2}{dict}
\pdf_object_write:nn {ar1}{/A/B}
\pdf_object_write:nn {ar2}{/B/C}
\pdf_object_new:nn{blub}{array}
\pdf_object_write:nx {blub}{\pdf_object_ref:n {ar1} \pdf_object_ref:n {ar2}}
\ExplSyntaxOff
abc
\end{document}
~~~~

without the fix this creates (see the various null entries):

~~~~
4 0 obj
<</A/B>>endobj
5 0 obj
<</B/C>>endobj
6 0 obj
[null
null
null
null
 5 0 R ]endobj
~~~~

with the fix the expected:

~~~~
4 0 obj
<</A/B>>endobj
5 0 obj
<</B/C>>endobj
6 0 obj
[ 4 0 R 
 5 0 R ]endobj
~~~~

### DIFF
--- a/l3backend/l3backend-pdf.dtx
+++ b/l3backend/l3backend-pdf.dtx
@@ -167,7 +167,7 @@
 \cs_new_protected:Npn \@@_backend_object_write_array:nn #1#2
   {
     \@@_backend_pdfmark:x
-      { #1 [ ~ \exp_not:n {#2} ~ ] ~ /PUTINTERVAL }
+      { #1 ~0~ [ ~ \exp_not:n {#2} ~ ] ~ /PUTINTERVAL }
   }
 \cs_new_protected:Npn \@@_backend_object_write_dict:nn #1#2
   {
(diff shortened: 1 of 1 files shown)


josephwright:
@u-fischer  Any reason we can't add this directly?

u-fischer:
> @u-fischer Any reason we can't add this directly?

no, only that I was quite in a hurry and didn't know which changes lists/versions etc I have to fill in.

josephwright:
@u-fischer  for the backend, as long as there is something in the log, I don't get too hung up: this is _all_ internal code

PR #736, opened 2020-06-01
Generalise color model support

josephwright:


josephwright:
@u-fischer In the direction you were thinking?

josephwright:
@jbezos  As you are the color space expert, your thoughts welcome here: we are trying to include spot colors but also other related ideas directly into `l3color`.

jbezos:
Sorry, I was not following the development of l3color, so I'm not sure what to review. But you can ask any questions, of course.

josephwright:
@u-fischer Occurs to me we could go with `\color_space_new:nnn` and then spot color is a type of color space: does that make sense? It's all about the initialisation if done that way.

josephwright:
@jbezos I was really wondering how general color space support needs to be/can be; at the moment, I'm working on spot colors, but @u-fischer points out there are other color spaces

u-fischer:
Sounds as if a bit of terminology is needed. pdf knows `color spaces` and `color space families`.

The reference mentions the families DeviceGray, DeviceRGB, and DeviceCMYK, CalGray, CalRGB, Lab, and ICCBased, Pattern, Indexed, Separation, and DeviceN.  Instances/members of this families are then the actual color space.

The DeviceRGB/Gray/CMYK families have only one member and so the notation is here often a bit mixed and e.g. DeviceRGB is used both for the color space family and the color space. 

A "spot color" is a member of the Separation family. The color space consists *not* all possible spot colors, but only of the tints of *one* color. The color space is a bit like the gray color space: the value ranges only from 0 to 1.

josephwright:
We could start with exactly that explanation for the 'creating new models' section, then call my suggested function `\color_space_new:nnnn` with arguments `<name>`, `<family>`, `<init>` and `<pars>`? Spot colors would then have `Separations` as their `family`, and that would be the meta-value we'd be using to set things up. The `init` argument would cover creating spot colors versus other separations.

u-fischer:
> We could start with exactly that explanation for the 'creating new models' section, then call my suggested function `\color_space_new:nnnn` with arguments `<name>`, `<family>`, `<init>` and `<pars>`? Spot colors would then have `Separations` as their `family`, and that would be the meta-value we'd be using to set things up. The `init` argument would cover creating spot colors versus other separations.

The problem is that I'm not sure if one can setup a color space in the various families in a similar way. To setup a Separation space one  needs basically only a name, the "real color name" and a cmyk value. But deviceN (if we ever want to tackle it) needs perhaps more. That's why I suggested to leave this open for now. I could define a "\__color_space_Separation_new:nnn" as first step and we generalize when we really know what is needed. 

For the definition and use  of  colors the families don't really matter. Here you need only the name of the space.

car222222:
The pdf (2.0) color model(s) and specifications is a good place to start for many reasons:
— pdf is currently our major target;
— it comes from Adobe so it will be very widely used and understood in a far wider worlds
— it is extensible (that is pretty much why things like device models can be specified in great generality);
— it fits well with my Master Plan for LaTeX to support ALL of PDF2.0 :-) 

But note that it is not possible to predict what these new device models will look like; there is an enormously diverse variety of complex color-related specifications and manipulations ‘out there’.

jbezos:
> The color space is a bit like the gray color space: the value ranges only from 0 to 1.

This is one of the key points, and `colorspace` ‘misuses’ the macros in `xcolor` for the Gray space to set the tints. Internally, spot colors are not like process colors (CMYK), because each spot color is a colorspace by itself, while process colors belong to a 4-dimensional colorspace. People expects the user interface will treat colors in more or less the same foot, and this makes things a little more complicated.

josephwright:
@u-fischer There are two obvious options. First approach is to have some general 'set up' arguments that have content dependent on the family, so something like `\color_model_new:nnnnn <name> <family> <init> <pars> <cmyk>`. The second is to accept that creating new models needs different stuff depending on the family, and go for one function per family.

josephwright:
@jbezos As we are integrating from scratch, covering internal issues is fine. I'm sure I can make color spaces 'work' together, it's just a question of deciding how that's done (falling back to CMYK is by far the easiest way to combine them).

u-fischer:
@josephwright I will try to setup some example color space definitions from colorspaces in the afternoon.

davidcarlisle:
On Tue, 2 Jun 2020 at 07:02, Javier Bezos <notifications@github.com> wrote:

> The color space is a bit like the gray color space: the value ranges only
> from 0 to 1.
>
> This is one of the key points, and colorspace ‘misuses’ the macros in
> xcolor for the Gray space to set the tints. Internally, spot colors are
> not like process colors (CMYK), because each spot color is a colorspace by
> itself, while process colors belong to a 4-dimensional colorspace. People
> expects the user interface will treat colors in more or less the same foot,
> and this makes things a little more complicated.
>
>
> in the original color package it was a deliberate decision to try to
simplify the interface by treating spot colors (which it calls named
colours) not as separate  colorspace/models but as a single 2-dimensional
model with the first dimension indexed by names rather than numerically;
hence

\color[named]{red,0.5}

rather than

\color[red]{0.5}

which it would be if each spot color were treated like grey as a separate
model.

but at the time few of the backends supported  the optional second numeric
argument, and the only drivers that allowed the names to pass to the back
end were dvips and textures, with everything else using cmyk fallbacks
(dvipsnam.def) ....

Of course the xcolor ! syntax overlaps with this somewhat...

jbezos:
> (falling back to CMYK is by far the easiest way to combine them).

I'm not sure what you mean, but spot colors must not be converted to CMYK in the pre-press stage. This must be done by the printer (except if the monitor was calibrated and you have a proper ICC profile, but these conversions are tricky).

josephwright:
@jbezos I mean that once one starts doing `spot-1!40!spot-2!20!some-other-model` then forcing to CMYK is likely the realistic approach

wspr:
Is it “allowed” to blend spot colours? Oh, I guess on certain printers it is — you use the correct inks for the job and the printer does the blending, right?

u-fischer:
> @jbezos I mean that once one starts doing `spot-1!40!spot-2!20!some-other-model` then forcing to CMYK is likely the realistic approach

I just looked how colorspace mix such colors and it creates a deviceN color space -- which probably mean that we should add that too ...  But the main problem is how to get a rgb or cmyk value that than can be used e.g. in link colors or html export or other places where such color spaces are not allowed or don't make sense.

jbezos:
On the contrary. A printer could reject this, because colorants (I mean, the actual liquid used in the press) won't mix correctly. It took me a good deal of work in `colorspace` mixing correctly spot and process colors,.

josephwright:
Mixing spot colors or similar is something that has to be added, but that's at a slightly later stage (as it's internal and thus we don't have to provide an interface). The first thing is how best to set up spot colors, etc., from a creation-interface point of view.

car222222:
@wspr asked: Is it “allowed” to blend spot colours?

Well, the answer depends on what your words actually mean.  There are color models that describe what you probably mean by ‘blending spot colors’ but they have different names.

car222222:
It is sometimes possible to deduce notional RGB and CMYK ‘values’ for colors that are specified in other colorspaces.  Obviously this is straightforward for those spaces that are ‘special spaces’ whose ‘underlying space’ is one of these two.  But for other colorspaces there may be no canonical relationship between a color value and an RGB or CMYK value.  

But it is important to remember that there is no one method of transforming between RGB values and CMYK values: different formulas are needed for different use-cases.

Also, no value in any finite-dimensional space actually uniquely defines ‘a color‘.  Something like a distribution over the visible spectrum might provide such a unique numerical value, but even that does not explain what an individual person or machine ‘sees’ when looking at a screen, or inked printing plate or piece of paper.  This is why one can have two color spaces between whose color values there is no meaningful conversion function.

car222222:
@josephwright A note for the future: 
The syntax (and even the semantics) of expressions such as these:
      spot-1!40!spot-2!20!some-other-model
will not make sense for some types of color space.  For example, when the color values themselves do not in any sense represent proportions (i.e., they are nothing like RGB or CMYK).

josephwright:
@car222222 Well yes, 'makes sense' is an issue. What I'm trying to plan for is that once you allow some mixing, which is handy for RGB/CMYK, then people expect the same syntax to work for other things. For a LAB colour space simply giving up with an error might well be the best plan, for a spot color mix, I'm not sure (but will extract ideas from `colorspace`). But as I say, first we need to be able to define new colour spaces.

u-fischer:
> But for other colorspaces there may be no canonical relationship between a color value and an RGB or CMYK value.

Yes. But we don't really need "canonical" relationships. We only need some relationship. We don't need to ensure "correct" colors along some color theory, only that users get the colors they want in various contexts. As an example assume that someone uses `\color{myspecialcolor}` in a document and then creates a pdf from this document and a html with tex4ht. Then the color command should work in both cases, but it doesn't need to work identically (which it can't anyway) or even produce similar looking color.

josephwright:
@u-fischer Exactly: what we need is to provide a way for the user to control what happens. In general terms we have plans for that, it's now a question of first setting up the alternative colour models.

For 'fallbacks', I'm currently thinking of something like `xcolor`:
```
{ SpotName/cmyk/rgb } { 0.8/0.1,0.2,0.3,0.4/0.5,0.6,0.7 }
```
where the separation takes a tint and the cmyk/rgb takes whatever.

I'll look at 'multi-model' support after sorting out how we want to set up new models.

jbezos:
>  (but will extract ideas from `colorspace`)

Since the code is somewhat chaotic (it's one of these package I wrote just because I needed, so I didn't care too much about readability or elegance), just ask me if you don't understand something (or anything 🙂). As I said, I ‘misused’ the internal macros of `xcolor` (and even patched a few of them, which is definitely a bad idea 🤞).

josephwright:
@jbezos Oh, don't worry, I usually pick stuff apart and re-work :)

car222222:
@jbezos “ I wrote just because I needed, “

So do you recall what you needed?  
Some use-cases would be useful in knowing what needs support.
Thanks.

car222222:
@u-fischer Yes, I understand the needs you explained.
“We only need some relationship.   . . . so that users get the colors they want in various contexts.”
So let us hope that “some relationship” satisfies these users!

“. . . or even produce similar looking color.”
which is unlikely to satisfy any user!  

But that is all that is possible and, of course, what colors the “user sees” (satisfied or not) depends on an enormous of external factors.  Hence it will vary far more than any differences introduced by any real-world conversion from one color space to another.

jbezos:
> So do you recall what you needed?

@car222222 

1. A package for spot colors which actually worked (`spotcolor` could generate invalid pdfs).
2. The `!` notation.
3. Shades, ie, spot + (process) black.

car222222:
Thanks, Javier.  So no CIE or ICC use-cases . . . yet!

I assume that the spot+black was really CMYK-based.  Did you need to use a Sepraration color space in the pdf, or could you just use DeviceCMYK, by calculating in LaTeX the CMYK color values of each color used, and putting these into the PDF?

u-fischer:
> Did you need to use a Sepraration color space in the pdf, or could you just use DeviceCMYK, by calculating in LaTeX the CMYK color values of each color used, and putting these into the PDF?

As the main point of the colorspace package is to create Separation color spaces I would say that was what Javier needed.

car222222:
Well, that certainly is what Javier did in colorspace.  Nothing wrong with that.  But maybe he could have achieved his purpose without needing a new family of special (auxiliary) colourspaces; hence my question.  

I hope my question made it clear that I am merely seeking further information, with no doubting or criticism of his motives.

u-fischer:
> Well, that certainly is what Javier did in colorspace. Nothing wrong with that. But maybe he could have achieved his purpose without needing a new family of special (auxiliary) colourspaces; hence my question.

I'm not sure what you mean. If you want to use spot colors you need such a colorspace, a spot color is a member of it, it is by definition not a DeviceCMYK color. I mean if someone wants to paint with oil colors one shouldn't start a discussion if it could be done with water colors instead.

car222222:
I just found one reason for using a Separation color space: this is to access subtractive colors that are known (only) by name to the application consuming the pdf. 

The extension of this concept to ‘blending’ a fixed set of multiple named colors needs a DeviceN colorspace

car222222:
‘spot color’ is an informal phrase with many meanings.  

Separation color spaces must have an underlying color space, which is very commonly a CMYK color space.  This is because they are special/auxiliary color spaces used only to deal with ‘named colors that known to a device’.

jbezos:
> I assume that the spot+black was really CMYK-based.

Is spot + black, to darken it.

car222222:
AFAICT from the documentation, colorspace does not even mention or cover the families Separation and DeciceN . . . 

So it seems that they cannot be how it implements spot colors or mixtures thereof.

u-fischer:
> AFAICT from the documentation, colorspace does not even mention or cover the families Separation and DeciceN . . .
> 
> So it seems that they cannot be how it implements spot colors or mixtures thereof.
You only need to look in the code or try it out:

~~~~
\RequirePackage{l3pdf}
\ExplSyntaxOn
\pdf_uncompress:
\ExplSyntaxOff
\documentclass{article}
\usepackage{colorspace}
\definespotcolor{foo}{BarTone 555 GN}{.8,.2,.5,.3}

\definecolorspace{fooshaded}{mixed}{foo,black}
\begin{document}
\color{foo!50} abc
\color[fooshaded]{.6,.3} abc
\end{document}
~~~~

The pdf then contains 

~~~~
1 0 obj
[ /Separation /BarTone#20555#20GN /DeviceCMYK << /FunctionType 2 /Domain [0 1] /C0 [0 0 0 0] /C1 [0.8 0.2 0.5 0.3] /N 1 >> ]
endobj
~~~~

and 

~~~~
3 0 obj
[ /DeviceN [ /BarTone#20555#20GN /Black ] /DeviceCMYK 2 0 R << /Colorants << /BarTone#20555#20GN 1 0 R >> >> ]
endobj
~~~~

So yes, colorspace implements this with separation and deviceN spaces.

car222222:
Good news, thanks!  All under the hood, but that’s fine.

car222222:
Thanks for the new stuff, Joseph.

I have one unrelated question about the medium term connection between l3color’s  low-level basics and the complex interfaces/syntax of xcolor.  Maybe something for a separate issue?

Also some suggestions most of which concern (I think) only the documentation of the various types of PDF color spaces, including the need, in cases like Separation, to provide meaningful (to the pdf post-processor/printer)  ‘pdf names‘.  
Should I add these to the discussion here?  
Or add them, suitably signposted, directly to the .dtx?

car222222:
One more question about strategy.  Are you supporting output other than PDF?  

This might affect more than just the backend and it certainly can affect what information is needed about a color or colorspace (and the syntax for this). 

 The main example being ‘just postcript’ for documents that are primarily intended to be further processed by a full postscript interpreter rather than simply converted to PDF.

josephwright:
@car222222 This PR is about generalising model support, so comments about how to describe that (naming "Separations" or whatever) seem to belong here. More general comments would be a separate issue: note that the backend is set up for SVG and PostScript output as well as PDF. (At present, I don't really know how to handle separations in those: probably have to fall-back to CMYK/RGB.)

car222222:
@josephwright   I think that there will be some (possibly several, RIP dependent)  postscript set-ups to target: have you looked already looked into this?  Do we know much about contemporary usage of, for example, postscript RIPs?

josephwright:
@car222222  Thus-far, my spot colour support efforts have been based on `colorspace`, so pdfTeX/LuaTeX only. I didn't get far reading the PostScript ref. to find 'matching' details. And of course there is SVG to think about: there, really we have to fall back to a parameter-based model.

davidcarlisle:
In the original color package drivers named/spot colors were only supported by the dvips and textures def files; the model being that the generated Postscript just had the name and optionally a numeric tint.  The actual separation (or not) was assumed to be downstream of that, so the supplied color.pro that comes with dvips species all its names as cmyk but you could supply a custom header file that just printed one color. For separating cmyk colors from postscript there was Sebastian's dvips/colorsep/colorsep.pro

josephwright:
@davidcarlisle Thanks for the pointer. Probably we'll want to add the PostScript to our own header file. That will need to be selective for the strictly-PostScript-based route. I'll look at that 'later': first we need to finalise the TeX interfaces. (PostScript separations are a backend thing.)

car222222:
Te SVG:  I just looked briefly at the latest (2018) standard.  It is AFAICT entirely concerned with (very sophisticated) presentation of stuff (including color effects) on screens; also, its simple color-related stuff is not much more than CSS, so that Separations, Process colors, etc do not feature.

Thus, at least until someone comes up with further needs, it should be OK to restrict to just RGB-based/related specifications (although I see that both CSS and SVG now have some ideas about CMYK, and maybe even ICC profiles, but it is not clear how central,these are).

jbezos:
> maybe even ICC profiles

I would include it. They may be essential in some workflows and it's easy, at least with the 3 basic color spaces. `colorspace` takes one more step because it reads the `.icc` file to detect its color space.

car222222:
@jbezos   Yes, sure, ICC profiles for specifying colors/spaces is straightforward.  

But some of the SVG ideas seem to go beyond this, using them as part of the processing workflow to do color matching at different stages.  I need to look into thus further.

I agree that reading them (at least as the default) from .icc files is the way to go as it means that no more than a file name needs to be specified.

josephwright:
Certainly professional-level ideas such as ICC profiles are on the 'to do' list, but first we want to get the core stuff sorted. @u-fischer and I have a flexible set-up planned for colour spaces, and I'm hoping to build on that to allow these ideas to fit in.

car222222:
@josephwright  More levels:-)!   

Where can we read something about your planned “flexible set-up”?

car222222:
Re CSS and ‘color mixing‘ :  when I wrote ‘not much more than CSS‘ above, I had not looked here:

https://www.w3.org/TR/compositing/

[Ignoring for now the  verbationizing  of ‘composite’ .  .  .  ] I found that, when it comes to putting colors on a screen, CSS supports a very wide gamut of features.
Not quite ‘ray tracing’ yet, but those ducks really get the lights shown on them:-).

But in CSS (and hence also SVG) mixing is treated as part of graphics and treated as the combination of graphical objects (‘compositing’), rather than ‘making new colors’, although the end-product of both is the production of a ‘coloured pixel’.

An example of one of the more complex types of color mixing effects:

10.1.10. soft-light blend mode

Darkens or lightens the colors, depending on the source color value. 
The effect is similar to shining a diffused spotlight on the backdrop.

Do you want to see the formula for this?

wspr:
@car222222 this is getting rather off topic for this pull request discussion, it would be better to keep these thoughts in a place that is easier to track / get back to in the future.

car222222:
@wspr I agree, so I asked @josephwright  about this.

Do you suggest a new issue, or what?

Maybe somewhere that includes the ‘flexible set-up plan‘ for wider discussion?

josephwright:
Indeed: I think we need a set of issues and a project. What I'll do is close this PR (as I want to set up a new interface approach anyway),set up a series of focussed issues and a small 'project', and hopefully things will stay organised. We can then add new issues to the project to address specifics as new ideas come up.

josephwright:
https://github.com/latex3/latex3/projects/1

josephwright:
@car222222 I've included an issue about PostScript support: would you like a separate one to cover SVG?

car222222:
@josephwright Well, it was you who introduced SVG. 
 
I do not have anything more to say about it at this stage but if we want to come back to it then an issue would be a good idea.

josephwright:
@car222222 OK, for the present we'll leave that: I have a decent list as it is

Issue #740, opened 2020-06-09
Add support for all PDF color space families


josephwright:
At present, `l3color` deals with separations as a `spot` model. This is rather limited. As discussed in #736, we should support all models offered by PDF (at least). To do this, we need something of the form
```latex
\color_model_new:nnn { name } { family } { settings }
```
plus of course the underlying mechanisms. The latter may need separate issues, for example to discuss how to mix spot colors.

car222222:
@josephwright wrote:  l3color deals with separations as a spot model

Is it not more the other at round?  It uses ‘spotcolor’ in the command to set up a Separation color space.  

It is not clear that it either should or does do anything extra in order to ‘deal with separations’.   But maybe you are including here the cases of more than one separation and of mixing them (plus mixing with process colors).

josephwright:
@car222222 I'm re-working support for Separations: expect updates next week

car222222:
"All PDF color spaces" done, then? 😃

PR #757, opened 2020-06-25
Enable creation of new color models

josephwright:
For dealing with #740.


josephwright:
One thing to think about here: do we want the family names to be case-sensitive. At present I've folded, but I'm not sure about `devicen` in place of `DeviceN` (`separation` vs `Separation` looks OK).

josephwright:
@jbezos Have you ever needed `\pagecolorspace`, by the way: I'm not clear if we need that functionality.

u-fischer:
@josephwright  I made a few tests last year and had the impression that one doesn't need it, imho there is somewhere even a comment and a test files about this in the pdfresource project, but I cant look now.

car222222:
What commands \...colorspace (and \...color) are you providing so far?

car222222:
My \\s got lost!

jbezos:
> @jbezos Have you ever needed `\pagecolorspace`, by the way: I'm not clear if we need that functionality.

The main uses are covers and plates (see sec. 4 in the manual). Without `\pagecolorspace` you may need to prepare different PDF files for them, which can be cumbersome. These special pages are expected to be in the PDF at the correct places and with the correct color spaces.

u-fischer:
@jbezos but what is a "correct colorspace " in this context? That it should contain the resources of the page is clear, but do additional resources from other pages harm?

jbezos:
It tells the printer how each page must processed (very likely with different inks in a different press). I've personally never used it, because I have created no book covers, but I've come across several PDFs with a different color space.

josephwright:
@jbezos  Perhaps I'm naive, but that sounds like a job normally done by making separate PDFs first, then combining them using a dedicated tool.

u-fischer:
@jbezos the problem is that while it is possible (at least with pdflatex/lualatex) to create an individual ColorSpace resource for every page, it requires a certain amount of code  and enlarges the pdf. tikz/pgf has now for many years creates one ColorSpace object for all pages and I haven't seen a complaint about this, so I would for now not do it, unless I see a concrete example of a problem with it.

jbezos:
Actually, two or three resources are enough (eg, body, book cover and plates), and once created they can be reused. Anyway, most prepress houses are used to deal with this, so I think it's not so essential and it can be left to an external package. This is one of the features I didn't manage to implement in `xetex`, and if “universality“ is a concern, this can be another reason to leave it out.

car222222:
Below are some facts about the (very small amount of) stuff that is needed for color spaces in the pdf page resources, plus the single object that is needed to set up a color space (just once) for use in any way or any place in the pdf.  

In particular, within the pdf there is no need for anything such as a ‘page color space’.

These facts may not be relevant to this discussion.  If not, please can someone explain to me why concepts such as ‘page color space’ are needed, and how they are in any way related to the anything in the ‘pdf page resources’.   

I understand that, on the LaTeX side, whenever any ‘new color’ is set up for use in a document, there is a need to check whether the color space used to define the ‘new color’ has already been ‘defined in the pdf’ or is a ‘new color space’; and thus, if necessary, add an (indirect) array object and an internal name for the ‘new color space’.
Note that this is related only to the color spaces used by a color and has no relation to how that color is used.

Thus there is no such concept as the ‘the color space needed for all page colors’.

Finally, The Facts (AFAICT) About pdf named and color spaces:

1.  Each ColorSpace resource only sets up ‘local internal names’ for color spaces, these names are needed only for use in a page content stream or streams.

2.  These internal names are arbitrary and are not in any way related to the colors that use each color space or how/where that color is used.  They are not needed for the three standard Device... spaces.  They are simply pdf name objects that are needed because direct references are not allowed within content streams.

3.  Each such ‘named color space’ also needs a one-time (per document) definition as an indirect array object.  (Thus the only (possibly) significant amount of 'space in the pdf' that is needed is: one indirect array object per distinct color space that may (or may not) be used in the document.  There is no need to duplicate this array, even if it gets more than one internal name, since an object reference will be used within the page resource.)

4.  Thus the necessary number of ‘color space arrays’ is just one per color space and is in no way related to types/places of use of each color space.

u-fischer:
@car222222  the question under discussion here is if we need page specific resource objects. If e.g. page one uses the color /color1 and /pattern1, and page two uses /color2, /shading1, /pattern2, there are two ways to sort the resources:

You can create a large object with the entries `<</color1 REF1 /color2 REF2 /pattern1 REF3 /shading1 REF4 /pattern2 REF4>>`  and reference this object in the ColorSpace resource of page 1 and page 2. 

Or you can create two object `<</color1 REF1 /pattern1 REF3>> ` and `<< /color2 REF2 shading1 REF4 /pattern2 REF5>>`  and reference the first object in the page resource of page one, and the second in the page resource of page two.

Currently both `pgf/tikz` and the resource management use method one. It is clearly easier to create: you simply collect all resources and dump then in one object. And gives less objects. The question is if there is a need to support method two.

car222222:
@u-fischer Thanks, I understand now, and agree that your final question needs an answer.

FWIW, I cannot envisage a use-case that needs more than one such Colorspace name-dictionary.

PR #758, opened 2020-06-25
Move l3cctab to stable

josephwright:


josephwright:
@blefloch 

> I've committed some easy refactorings to simplify and speed up the
> non-LuaTeX side. 

Thanks

>     * Really, cctabs should include the `\endlinechar` setting.  In
>       non-LuaTeX engines just store it as entry 257 of the intarray.  In
>       LuaTeX just make a cctab into a two-item token list with first the
>       `\endlinechar` setting and then the `\catcodetable` number.  I
>       remember you mentioned some issue with manipulating that from the Lua
>       side?  Presumably we can also populate a table `cctab.endlinechar`
>       with the endlinechar value corresponding to each catcodetable.  In the
>       same vein, in XeTeX, should we include a few code-points beyond 256
>       which are likely to be made active or spaces or whatever?

On LuaTeX: the point was that onthe Lua side, there are places one can pass a `<catcode table>` for setting the category code of tokens returned to TeX. Those will *only* respect the 'real' catcode part, not any additional `\endlinechar` setting we include at the macro level. This was the issue really: is that confusing/misleading?

On XeTeX, I picked 0-255 as this was consistent with upTeX and easy to document. I'd rather not have a 'plus various arbitrary places' aspect to document.

>     * I don't like that `\cctab_const:Nn #1 { }` and `\cctab_new:N #1` don't
>       result in the same catcode assignments: prevailing one versus IniTeX
>       one.  Perhaps make both give the prevailing one?  (That requires
>       changing slightly how `\c_initex_cctab` itself is defined.)

The idea was that `\cctab_new:N` is like any other `new` function and starts from _a fixed point_, whereas `\cctab_const:Nn` is more 'convenient'. If we want a change, I'd rather we altered `\cctab_const:Nn`, since all other `new` functions produce a known situation.

>     * I've stated in the documentation that `\cctab_begin:N` and
>       `\cctab_end:` must be used in the same group (and same group level)
>       because any other use is confusing.  Maybe I'm the only one confused
>       and someone can clarify what should happen for `\group_begin: \cctab_begin:N \group_end: \cctab_end:` (or the opposite order).

That's true: the idea is they don't _necessitate_ a group, not that they can be intermixed.

>     * Can we make it so that `\cctab_select:N` in LuaTeX does not need an
>       `\aftergroup` part (which breaks in some kinds of groups and prevents
>       us from implementing `\cctab_begin:N` as "store current catcodes for
>       later use and apply `\cctab_select:N`").

The idea of `\cctab_select:N` is it only applies in a few places. Perhaps we could restrict it strictly to those places (by having an internal name and have the use cases activate it)? Or we could go back to my original idea of requiring dedicated 'use a catcode table' of `\tl_set_rescan:Nnn`, etc.

>     * In non-LuaTeX, rather than having a bunch of `+1` or `-1` we should
>       just treat the code-point 0 separately.  We should also optimize by
>       using the low-level `\tex_catcode:D`.

That seemed the easiest way to me, but if you'd rather special-case point 0, shift it to 256?

>     * Document that the second argument of `\cctab_gset:Nn` and similar can
>       contain `\cctab_select:N`.  Same comment for `\tl_rescan:nn` and
>       similar: document in `l3tl` and in `l3cctab`.

I'd not sure I'd want to allow nesting, whereas the whole point of `\cctab_select:N` is for `\tl_rescan:nn` and similar, so yes that can be added.

>     * More tests such as what happens when we mix in some catcode settings
>       with catcode tables (to detect if we mess up when tweaking the LuaTeX
>       implementation).  Test `\cctab_select:N` and groups.

Fair enough: I'll perhaps add a few.

>     * Use catcode tables for `\ExplSyntaxOn/\ExplSyntaxOff`?  This requires
>       some care.

This was on my list but not in the first step: I thought we could think about the risk/reward balance separately.

blefloch:
Thanks, I agree with, or am convinced by, most of your comments.

- Should \c_code_cctab really include @=letter?

- I'm confused how our use of the stack interacts with the LaTeX2e's
  kernel definition for \newcatcodetable, which seems to allocate
  catcodetables consecutively.

>     Really, cctabs should include the `\endlinechar` setting. [...] populate
>     a table `cctab.endlinechar` with the endlinechar value corresponding
>     to each catcodetable.
> 
> On LuaTeX: the point was that onthe Lua side, there are places one can pass a
> |<catcode table>| for setting the category code of tokens returned to TeX. Those
> will /only/ respect the 'real' catcode part, not any additional |\endlinechar|
> setting we include at the macro level. This was the issue really: is that
> confusing/misleading?

I think we could provide wrappers that do respect the endlinechar setting,
provided it is accessible from the Lua side.  Which Lua functions are you
thinking about?  I didn't see any l3cctab.lua file for instance.

>     |* I've stated in the documentation that `\cctab_begin:N` and `\cctab_end:`
>     must be used in the same group (and same group level) because any other use
>     is confusing. Maybe I'm the only one confused and someone can clarify what
>     should happen for `\group_begin: \cctab_begin:N \group_end: \cctab_end:` (or
>     the opposite order). |
> 
> That's true: the idea is they don't /necessitate/ a group, not that they can be
> intermixed.

Do you see a good way to check that they are not intermixed?  \currentgrouplevel
is not enough.

>     |* Can we make it so that `\cctab_select:N` in LuaTeX does not need an
>     `\aftergroup` part (which breaks in some kinds of groups and prevents us
>     from implementing `\cctab_begin:N` as "store current catcodes for later use
>     and apply `\cctab_select:N`"). |
> 
> The idea of |\cctab_select:N| is it only applies in a few places. Perhaps we
> could restrict it strictly to those places (by having an internal name and have
> the use cases activate it)? Or we could go back to my original idea of requiring
> dedicated 'use a catcode table' of |\tl_set_rescan:Nnn|, etc.

I think \cctab_select:N could be implemented in LuaTeX by having a single
"scratch" \catcodetable (say, 2) and switching to that: something like

\catcodetable = #1
\savecatcodetable 2
\catcodetable = 2

No need to touch the stack etc I think.

>     |* In non-LuaTeX, rather than having a bunch of `+1` or `-1` we should just
>     treat the code-point 0 separately. We should also optimize by using the
>     low-level `\tex_catcode:D`. |
> 
> That seemed the easiest way to me, but if you'd rather special-case point 0,
> shift it to 256?

That would be faster for sure.

>     |* Document that the second argument of `\cctab_gset:Nn` and similar can
>     contain `\cctab_select:N`. Same comment for `\tl_rescan:nn` and similar:
>     document in `l3tl` and in `l3cctab`. |
> 
> I'd not sure I'd want to allow nesting, whereas the whole point of
> |\cctab_select:N| is for |\tl_rescan:nn| and similar, so yes that can be added.

Actually, it turns out \cctab_select:N is safe also in \cctab_const:Nn and
\cctab_gset:Nn so I've documented that.

>     |* More tests such as what happens when we mix in some catcode settings with
>     catcode tables (to detect if we mess up when tweaking the LuaTeX
>     implementation). Test `\cctab_select:N` and groups. |
> 
> Fair enough: I'll perhaps add a few.

I'll try to, but not today.

blefloch:
Remaining questions:

- What added/updated dates should we put for the various macros?  Now catcode
tables include endlinechar.

- Should \c_code_cctab really include @=letter?

- In my new implementation, \cctab_begin:N \foo \cctab_end: leaves
\catcodetable=5 (the scratch table I allocated) rather than \catcodetable=value
before the begin.  This is more or less on purpose, because the other
catcodetable could be altered by other non-LaTeX3 code.

- We should detect \group_begin: \cctab_begin:N \group_end: \cctab_end:

blefloch:
On 7/2/20 5:13 PM, Bruno Le Floch wrote:
> Remaining questions:
> 
> - What added/updated dates should we put for the various macros?  Now catcode
> tables include endlinechar.
> 
> - Should \c_code_cctab really include @=letter?

More generally, I'm not sure what catcodes \c_document_cctab should hold,
especially when e.g., inputenc is loaded.  Should chars 128-255 be active in
pdfTeX, upTeX, pTeX?

> - In my new implementation, \cctab_begin:N \foo \cctab_end: leaves
> \catcodetable=5 (the scratch table I allocated) rather than \catcodetable=value
> before the begin.  This is more or less on purpose, because the other
> catcodetable could be altered by other non-LaTeX3 code.
> 
> - We should detect \group_begin: \cctab_begin:N \group_end: \cctab_end:
>

PhelypeOleinik:
> - Should \c_code_cctab really include @=letter?

Probably not, if we want to have `\ExplSyntaxOn` be a cctab setting, because it doesn't change `@`.  However making `\ExplSyntaxOn` a cctab will break code that does `\makeatletter \ExplSyntaxOn`, unless we take extra care not to change `@`.  I'm not sure which way is best.

>  - We should detect \group_begin: \cctab_begin:N \group_end: \cctab_end:

We do now :-)

> More generally, I'm not sure what catcodes \c_document_cctab should hold,
especially when e.g., inputenc is loaded.  Should chars 128-255 be active in
pdfTeX, upTeX, pTeX?

There's also `babel`, making things active... Should the `\c_document_cctab` be defined (or updated, but then it wouldn't be `\c_...`) at `\begin{document}`? (probably not)

josephwright:
I'm going to revise the 'standard' catcode settings, document the plan, see how people feel about it.

josephwright:
I'm planning to alter the semantics of `\cctab_gset:Nn` and thus `\cctab_const:Nn`, so that they always work from IniTeX codes and things have to be set explicitly. That will be a little slower (have to reset everything) but I think clearer. Then we can work out exactly what we want to do about other catcodes: I have ideas.

blefloch:
On 7/6/20 3:41 PM, Joseph Wright wrote:
> I'm planning to alter the semantics of |\cctab_gset:Nn| and thus
> |\cctab_const:Nn|, so that they always work from IniTeX codes and things have to
> be set explicitly. That will be a little slower (have to reset everything) but I
> think clearer. Then we can work out exactly what we want to do about other
> catcodes: I have ideas.

Ok.  We should point out in the doc that \cctab_gset:Nn behaves rather
differently from \tl_rescan:nn in that respect since the latter uses the current
catcodes by default.

josephwright:
@blefloch It's the business that `\cctab_new:N` otherwise acts differently from `\cctab_const:Nn`; in the `tl` case, there is no confusion.

josephwright:
I think we should have addressed @blefloch's concerns now, other than the idea of using tables for `\ExplSyntax...` (which I think is a separate change that needs more thought).

blefloch:
Most of my concerns are addressed.

The quote character is "other" in \c_code_cctab and \c_document_cctab.  What
about other characters like "=" or "," that are occasionally made active?  In
\c_code_cctab I would expect catcodes of chars 0-127 (at least) to be completely
normalized.  In \c_document_cctab I don't know what to expect, but it's not
great to have a dependence in loading order of various packages.

I find it hard to follow the code that checks grouping of \cctab_begin:N /
\cctab_end: so I'm not sure if it does the right thing or not.  I have some
other approach ("git stash"-ed away) and I'll try to dig it up.

Where do we put the information for debug options to catch local/global
assignments etc?  There should be some stuff for cctab there, to catch when one
assigns to a constant cctab, and so on.

PhelypeOleinik:
> I find it hard to follow the code that checks grouping of \cctab_begin:N / \cctab_end: so I'm not sure if it does the right thing or not. I have some other approach ("git stash"-ed away) and I'll try to dig it up.

Sorry about that, I tend to over-complicate code.  For each `\cctab_begin:N` the code stores the current group level on a stack and locally defines a macro `\@@_group_<register-no>_chk:`.  Then, when the cctab `<register-no>` ends, it checks that the end occurred in the same group level and that `\@@_group_<register-no>_chk:` still exists (the group level can be the same but a group might have ended and another started in between).

The stack is mostly needed to know when the `\cctab_end:` occurred in a lower group level to print it in the error message.  If that's not necessary, then the stack is useless and just the sentinel `\@@_group_<register-no>_chk:` is needed, and the code gets a bit simpler:
```latex
\cs_new_protected:Npn \@@_chk_group_begin:n #1
  {
    \cs_set:cpx { @@_group_ #1 _chk: }
      { \int_use:N \tex_currentgrouplevel:D }
  }
\cs_new_protected:Npn \@@_chk_group_end:n #1
  {
    \cs_if_exist:cTF { @@_group_ #1 _chk: }
      {
        \int_compare:nNnF
            { \use:c { @@_group_ #1 _chk: } } = { \tex_currentgrouplevel:D }
          { \__kernel_msg_error:nn { kernel } { cctab-group-mismatch } }
        \cs_set_eq:cN { @@_group_ #1 _chk: } \tex_undefined:D
      }
      { \__kernel_msg_error:nn { kernel } { cctab-group-mismatch } }
  }
```

> Where do we put the information for debug options to catch local/global assignments etc? There should be some stuff for cctab there, to catch when one assigns to a constant cctab, and so on.

In `l3debug.dtx`. I'll add entries for cctabs there.

josephwright:
@blefloch For pre-loaded `expl3`, there is no issue with load-order: you'll get what the kernel sets. For the code table, I think we don't want stray active chars: we already suppress `"` in `\ExplSyntaxOn`. For the document one, things are more tricky as there is no one set of active chars. We probably could document that the catcode table is the _standard_ set up. Arguably `"` could be active as that's pretty common and should be relatively safe. Other chars are not, and we get into the whole '`$` would be better active' business.

PhelypeOleinik:
@blefloch It seems that reverting commit https://github.com/latex3/latex3/pull/758/commits/a5ef4aec3e66e85fd64b73231682c31b0836e0bb gives the correct errors with your new tests (trying to break my code, eh? ;-)

Though still, I'm not 100% sure it does the right thing, so feel free to change.  A much easier way to catch that would be with `\aftergroup`, but I don't think we want to use that.

blefloch:
I agree with Joseph's comment that \c_document_cctab and \c_code_cctab should be a fixed set of catcodes corresponding to what comes out of the kernel, but that means we must set catcodes of all chars <127 explicitly when defining \c_code_cctab.  I've done that in a commit.  Perhaps we should also freeze the catcodes of chars 128-255?

I've fixed the group checking code.  We also need to check at the very end of the run that all `\cctab_begin:N` have been closed by `\cctab_end:`.  How do we put such a hook?  Maybe it should wait until after hook management is in the kernel?

josephwright:
@blefloch The reason I didn't 'force' catcodes here was that LaTeX2e makes the 'decision': I set only those I could see we needed to. Things are of course somewhat blurred once you start using catcode tables as the LaTeX3 settings will apply wherever someone uses a cctab.

blefloch:
On 7/15/20 12:16 PM, Joseph Wright wrote:
> @blefloch <https://github.com/blefloch> The reason I didn't 'force' catcodes
> here was that LaTeX2e makes the 'decision': I set only those I could see we
> needed to. Things are of course somewhat blurred once you start using catcode
> tables as the LaTeX3 settings will apply wherever someone uses a cctab.

Feel free to revert my commit on this or not, and merge the branch.  On the
other hand, checking that cctabs are closed at the end of the run will have to
wait I think.

PhelypeOleinik:
I think the unclosed cctabs can be added now, with a proper check if `\hook_gput_code:nnn` exists, and do `\hook_gput_code:nnn {enddocument/end} {kernel} { \@@_check_end: }`, or something like that. And if `\hook_gput_code:nnn` doesn't exist, we can just say this check isn't supported in older kernel, if we don't want to patch `\enddocument`.

PR #760, opened 2020-06-27
Register prefix (ptxcd)

TeXhackse:
As discussed via email I'd like to register this prefix to collect multiple corporate design templates which won't interact with each other. 

Because it's not directly a package i called the module ptxcd but was not sure how use the url fields. I can offer to add a webpage where I collect all templates, which use this prefix. 

Thanks for your work.


FrankMittelbach:
Am 27.06.20 um 13:30 schrieb Marei Peischl:
> Because it's not directly a package i called the module ptxcd but was 
> not sure how use the url fields. I can offer to add a webpage where I 
> collect all templates, which use this prefix.

the idea behind the url is to give a pointer to where the source is, 
e.g. on github or ... or where on CTAN   --- whatever might be helpful 
for somebody who wants to track down information on the prefix

TeXhackse:
I got that point, but since it's more than one URL I will leave this empty and think about a good way to collect this information. As I realized I can add this later.

PR #764, opened 2020-07-03
Register prefix (pgf)

hmenke:
I would like to register the `pgf` prefix for future usage and in anticipation of someone else attempting to register it.  It is not actively used yet, but eventually we hope to be able to migrate PDF resource management to `pdfresources` (at least for LaTeX) which will make this necessary.


PhelypeOleinik:
Thanks for registering!

PR #765, opened 2020-07-03
Fix two typos in doc [ci skip]

muzimuzhi:
- In `xtemplate.dtx`, replace mis-spelled `\cs{EvaluteNow}` to `\cs{EvaluateNow}`.
 - In `xparse.dtx`, avoid short verb `|...|` in the argument of `\texttt`.

The second typo, introduced in https://github.com/latex3/latex3/commit/dd0062fa65d32f4974691f51d773c6949484b197, causes following error when running `lualatex xparse.dtx` to get the doc of `xparse` with documented implementation:
```
! LaTeX Error: \verb illegal in command argument.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...

l.2243 %   of~\dots~has an extra~|}
                                 |}.
```

### DIFF
--- a/l3packages/xparse/xparse.dtx
+++ b/l3packages/xparse/xparse.dtx
@@ -2240,7 +2240,7 @@
 %   that is not possible without the risk of mistakenly grabbing the
 %   entire brace group (potentially leading to a~\texttt{!~Runaway argument}
 %   error) or trying to grab a |}|$_2$, leading to a~\texttt{!~Argument
-%   of~\dots~has an extra~|}|}.
+%   of~\dots~has an extra~\}}.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_allowed_token_check:N #1
   {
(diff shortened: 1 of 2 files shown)


PhelypeOleinik:
Whoever put verbatim in an argument seems to have no idea how TeX works ;-)

muzimuzhi:
Perhaps latex3 can add the building of full documentation of `xparse` (and similar ones) to its CI, in order to find similar (severe) typos as early as possible?

PhelypeOleinik:
The implementation of `l3kernel` is run by the `doc` target to build `source3.pdf`, and when `xparse` is moved to the kernel, it will be checked as well.  The other packages don't have the implementation included in the doc PDF, but perhaps they could be checked for errors, at least...

josephwright:
@PhelypeOleinik  Newer packages have explicit code doc/user doc splits: I think at present that's likely enough.

PR #768, opened 2020-07-14
Peek analysis

blefloch:
This is a tool to parse input, more flexible than individual `\peek_meaning` and similar.  I am not sure if I got the best interface.  It is maybe not extensively tested enough.  I don't know if I should put the documentation of `\peek_analysis_map_inline:n` in `l3candidates` or `l3token` instead of `l3tl-analysis`.

Eventually I plan to use `\peek_analysis_map_inline:n` in a `\peek_regex` function that would look for a regular expression in the input stream.  I still have to reorganize the regex code somewhat.

Note to Phelype: I use the `\afterassignment ... \futurelet ...` construction I had mentioned in an email some time back.


josephwright:
Looks pretty reasonable to me, but like you say, the documentation should be in `l3token`. (I'd rather avoid any additions to `l3candidates`: we should review in the PR mechanism then commit to new functions.)

blefloch:
There is no rush to merge.  Before we do I'd like someone to sanity check the following:
- The mechanism I chose for stopping `\peek_analysis_map_inline:n`, namely a `break` function that jumps out of the loop, and such that tokens in the input stream that have already been looped over are completely removed.
- The name of `\peek_analysis_map_break:`, which is longish.  As a comparison, `\tl_analysis_map_inline:nn` is broken by `\tl_map_break:`, not by some hypothetical `\tl_analysis_map_break:`, so perhaps here we should use `\peek_map_break:`?
- Should there be a dedicated way to put input tokens back (like the difference between `\peek_remove` and other `\peek` functions.  Perhaps provide `\peek_map_break:` vs `\peek_remove_map_break:`, where one would put the input tokens cleanly back into the input, and the other would not.
- Whether to document edge cases where my code breaks:
  - if there are two control sequences with the same meaning equal to an implicit begin-group/end-group/space token, say `\bgroup` and `\bgroup@another` and the input is `\bgroup@another @` with `@` of catcode other, then what `\peek_analysis_map_inline:n` sees is `\bgroup @another@` (namely wrong control sequence, and a bunch of catcode-other character tokens);
  - upon reaching the very end of the main file if `\everyeof` ends with a begin-group/end-group/space token;
  - possibly `\outer` tokens, although I tried to be careful to pass them all the way to the user code without breakage;
  - possibly in tabulars, although I naively expect not.

blefloch:
I pushed a separate branch, peek-regex, showing an application of
\peek_analysis_map_inline:n (but that separate branch is not yet complete).

PR #773, opened 2020-07-16
Correct implementation of \str_set_convert:Nnnn for 8-bit encodings

blefloch:
The code was undocumented and abused plenty of `\toks`, `\count` etc.  Now documented and hopefully no more bugs, based on intarrays.

Doing this as a pull request because I'm deprecating a user command that was most likely not used at all: `\str_declare_eight_bit_encoding:nnn`.  Realistically no one is going around defining new 8-bit encodings for use in `\str_set_convert:Nnnn`.


josephwright:
@mvsfrasson  Probably best to discuss in the issue rather than this closed PR: I'll comment there!

PR #779, opened 2020-07-23
Peek regex

blefloch:
This branch (which branches out of the `peek-analysis` branch) implements ways to peek ahead into the input stream.  The code should be ok, but I am not sure if the interface I chose is ok.

I implemented `\peek_regex:nTF`, `\peek_regex_remove_once:nTF`, and `\peek_regex_replace_once:nnTF`, and the same with first argument being a regex variable instead of an n-type regex.  I provided `\peek_regex_replace_once:nn`, but not the other "no TF" versions of the other functions because they don't seem really sensible.  In fact, `\peek_regex:nTF` and `\peek_regex_remove_once:nTF` could be implemented as

    \cs_new_protected:Npn \peek_regex:nTF #1
      { \peek_regex_replace_once:nnTF {#1} { \0 } }
    \cs_new_protected:Npn \peek_regex_remove_once:nTF #1 
      { \peek_regex_replace_once:nnTF {#1} { } }

so maybe the `remove_once` functions could be omitted.  We could also reasonably provide `\peek_regex_replace_all:nnTF` which would repeatedly search for the pattern.

Perhaps tests I put in `m3regex011` are not thorough enough.


josephwright (approved):
This all looks good to me. I think the names `\peek_regex:n(TF)`, etc., are good. I guess `\peek_analysis_map_inline:n` is slightly awkward, but I don't see a better one - iot's quite specialised anyway.

Overall I think we go for this then can move to extending `xparse`.

PR #785, opened 2020-08-02
Faster keys

Skillmon:
This pull request changes the definition of `\__keys_trim_spaces:n` to not use `\tl_trim_spaces:n`, but instead directly remove the spaces around slashes in its loop. As a result `\keys_set:nn{test}{key=val}` with `\keys_define:nn{test}{key .tl_set:N = \l_test_key_tl}` gets more than 10% faster according to my benchmarks.

Also a faster `\__keys_parent:o` version is added (and the unused `:n` type dropped).

The function `\__keys_find_key_module:NNw` got replaced by `\__keys_find_key_module:wNN`.

The function `\__keys_property_find:n` also got faster and two (to my knowledge) unreported bugs got fixed. If a key path contained at least one period (additionally to the one starting the property), an empty `\l__keys_module_str` wasn't handled correctly, and a trailing space wasn't stripped from the `\l_keys_path_str`. So `\__keys_property_find:n { abc.def~.code:n }` with an empty `\l__keys_module_str` would have resulted in `\str_set:Nn \l_keys_path_str { /abc.def~ }` instead of `{ abc.def }`.

In total this PR should speed up `\keys_set:nn` by at least 20 % and `\keys_define:nn` by at least 30 % (I only benchmarked for the simple key already mentioned above).

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -1584,7 +1584,10 @@
 %   }
 %   Searching for a property means finding the last |.| in the input,
 %   and storing the text before and after it. Everything is turned into
-%   strings, so there is no problem using an \texttt{x}-type expansion.
+%   strings, so there is no problem using an \texttt{x}-type expansion. Since
+%   |\__keys_trim_spaces:n| will turn its argument into a string anyway, this
+%   function uses \cs{cs_set_nopar:Npx} instead of \cs{tl_set:Nx} to gain some
+%   speed.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_property_find:n #1
   {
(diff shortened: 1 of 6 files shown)


Skillmon:
> I don't think there is test coverage here: could you add some as part of the PR, so we can be sure there are no impacts?

Well, all the original code does is turn something into a string and call `\tl_trim_spaces:n` on every slash delimited part. It is only used inside of either `c` or `x` type arguments, so one doesn't have to pay attention on returning inside a single `\exp_not:n` or anything like that.

But yes, I can add some tests :)

Skillmon:
@josephwright  Are more changes ok inside of this PR or should I open a new one and stick to one PR per function (just in case I find more things which might get much faster)?

josephwright:
Pull requests can be extended, certainly here: I'll not merge at the moment

josephwright:
Tests should be at the documented API level unless there is a real pressing reason

PR #786, opened 2020-08-05
Faster keys (again)

Skillmon:
This PR will speed up `l3keys` by many small optimizations. Those optimizations include:

- curry every argument that could be curried
- change `\tl_set:N(n|o|x)` to `\cs_set_eq:NN` where possible
- change `\(tl|str)_set:(n|x)` to `\cs_set_nopar:Np(n|x)` where possible (places where we know the argument will expand to a string)
- remove `o`-variants and use `\exp_args:No` directly on the places they would be used (where applicable)
- remove `\exp_not:(n|o)` in places where we know the argument is (or expands to) a string
- change `x`-type expansion to `e`-type expansion where possible

[Small benchmark.](https://github.com/latex3/latex3/pull/786#issuecomment-669474611)


FrankMittelbach:
Am 05.08.20 um 14:22 schrieb Skillmon:
>   * curry every argument that could be curried
>   * change |\tl_set:N(n|o|x)| to |\cs_set_eq:NN| where possible
>   * change |\(tl|str)_set:(n|x)| to |\cs_set_nopar:Np(n|x)| where
>     possible (places where we know the argument will expand to a string)
>   * remove |o|-variants and use |\exp_args:No| directly on the places
>     they would be used (where applicable)
>   * remove |\exp_not:(n|o)| in places where we know the argument is (or
>     expands to) a string
>   * change |x|-type expansion to |e|-type expansion where possible

sorry, I'm not agreeing to the above changes unless the time 
improvements are extreme (and then only in those place) because some 
operations are used extremely often (or as Don Knuth would but are in 
the inner loop). Otherwise I'm against sacrificing  structure and 
readability.

Skillmon:
@FrankMittelbach which of the changes would be ok? E.g., I'd argue changing `\use:x` to `\use:e` doesn't affect structure, but `\use:e` is faster than `\use:x`. Using `\tl_set:No #1 { \l__keys_unused_clist }` doesn't seem more structured than `\cs_set_eq:NN #1 \l__keys_unused_clist`. Also I don't think that `\cs_new:Npn \foo:nn #1 #2 { \bar:nnn { g } {#1} {#2} }` is much easier to understand compared to `\cs_new:Npn \foo:nn { \bar:nnn { g } }` (which is indeed used elsewhere in the module already).

The only changes I'd perhaps agree do indeed sacrifice readability are the uses of `\cs_set_nopar:Np(n|x)` (but I made sure to always state that the argument will already be a string in those cases).

And I paid attention to not fully destroy readability by using stuff like `\exp_after:wN \foo:n \exp_after:wN {#1}` but always stick to `\exp_args:No` and friends.

josephwright:
On  when to absorb arguments, the older code tends to leave them, but I've favoured (re)grabbing at each step in newer stuff: it's a lot clearer what's going on when `\tracingall`.

FrankMittelbach:
> @FrankMittelbach which of the changes would be ok? 

none other than your first point.

> E.g., I'd argue changing `\use:x` to `\use:e` doesn't affect structure, but `\use:e` is faster than `\use:x`. 

agreed it does not change the structure  and it is arguably better (as long as you are using a recent engine (do we stil have emulations of "e" for older engines?)

> Using `\tl_set:No #1 { \l__keys_unused_clist }` doesn't seem more structured than `\cs_set_eq:NN #1 \l__keys_unused_clist`. 

no, you put unnecessary implementation details in your code and it sets a bad example. clists or strings or tls should be manipulated by their own mutator functions.

> Also I don't think that `\cs_new:Npn \foo:nn #1 #2 { \bar:nnn { g } {#1} {#2} }` is much easier to understand compared to `\cs_new:Npn \foo:nn { \bar:nnn { g } }` (which is indeed used elsewhere in the module already).

in the early days we have done this to conserve space, but again it is bad style. For something as trival as your foo/bar example, it is ok but in real code where things are typically more complicated, no.
 
> And I paid attention to not fully destroy readability by using stuff like `\exp_after:wN \foo:n \exp_after:wN {#1}` but always stick to `\exp_args:No` and friends.

again cs variant are better than explicit \exp_args:No.

Skillmon:
@FrankMittelbach I really don't get how `\cs_set_eq:NN \l_tmpa_tl \l_tmpa_clist` is assuming more implementation details than `\tl_set:No \l_tmpa_tl { \l_tmpa_clist }` or `\tl_set:Nx \l_tmpa_tl { \exp_not:o \l_tmpa_clist }`. Both assume that in one expansion step a `clist` variable expands to its contents, not more not less, but `\tl_set:No` is 2.3 times slower for an empty `clist` and 2.7 times slower for a `clist` containing 3 single-token elements (and `\tl_set:Nx ... { \exp_not:o ... }` is even worse). Would you be happier if I used `\tl_set_eq:NN \l_tmpa_tl \l_tmpa_clist` instead?

`\exp_args:No` is used 12 times in `l3keys` already, after this PR it'll be used 15 times instead. That's 3 more usages.

I understand that you're against these changes, and after all it is the team to decide whether to merge this PR or not, not mine. But I really don't understand your arguments for these points.

josephwright:
On `\exp_args:No` vs generating a variant, I've tended to go with 'if it's a one-off, use `\exp_args:...`, if it's used more than once, make a variant'. I hope that's OK!

Skillmon:
**Finally benchmarked!**

Benchmarking done with `\keys_define:nn { test } { key .tl_set:N = \l_test_key_tl }` and `\keys_set:nn { test } { key = val }`, benchmarked 12 times each, target time 1s.

Compared to the latest master (that means compared to aad4dc348) this PR speeds up `\keys_define:nn` by more than 7 % and `\keys_set:nn` by more than 12 %.

car222222:
Frank wrote concerning \cs_set_eq:NN  
 "you put unnecessary implementation details in your code and it sets a bad example. clists or strings or tls should be manipulated by their own mutator functions."
which suggests to me that he tl mutator function \tl_set_eq:NN #1  would be the best command to use here (assuming it exists:-).

FrankMittelbach:
> Frank wrote concerning \cs_set_eq:NN
> "you put unnecessary implementation details in your code and it sets a bad example. clists or strings or tls should be manipulated by their own mutator functions."
> which suggests to me that he tl mutator function \tl_set_eq:NN would be the best command to use here.

that is why they exist for all data types tl clist str etc

Skillmon:
> which suggests to me that he tl mutator function \tl_set_eq:NN would be the best command to use here.

Which would mean to use `\tl_set_eq:NN` with one `tl` variable and one `clist` variable as its arguments. I don't get the reasoning why this should be the correct way at all.

car222222:
@Skillmon Aha!  You spotted the flaw in my suggestion!  Well, I was 50% correct:-).

Actually this points out a possible missing command, that produces the underlying plain tl from a 'any structure built from that tl'.  (Sometimes called the 'forgetful functor'.) I shall open an issue about that.

l3kernel/l3keys.dtx
```diff
@@ -2371,17 +2370,17 @@
...
 %   }
 % \begin{macro}{\@@_set:nnn}
-%   A simple wrapper allowing for nesting.
+%   A simple wrapper allowing for nesting. The first argument to |\@@_set:nnn|
+%   should always be a string.
 %    \begin{macrocode}
 \cs_new_protected:Npn \keys_set:nn #1#2
   {
-    \use:x
+    \use:e
```
blefloch, line 2378: Why?  `\use:e` is faster in recent engines but extremely slow otherwise, and this is not worth a separate code branch depending on whether `\expanded` is available, because it is not in the "inner loop" of setting each individual key.

l3kernel/l3keys.dtx
```diff
@@ -2371,17 +2370,17 @@
...
       {
         \bool_set_false:N \exp_not:N \l_@@_only_known_bool
         \bool_set_false:N \exp_not:N \l_@@_filtered_bool
         \bool_set_false:N \exp_not:N \l_@@_selective_bool
-        \tl_set:Nn \exp_not:N \l_@@_relative_tl
-          { \exp_not:N \q_@@_no_value }
+        \cs_set_eq:NN \exp_not:N \l_@@_relative_tl \exp_not:N \q_@@_no_value
```
blefloch, line 2383: Ok, but use `\tl_set_eq:NN`, see comments below on clist.

l3kernel/l3keys.dtx
```diff
@@ -2394,13 +2393,13 @@
...
-  { \exp_args:No \@@_set:nnn \l_@@_module_str {#1} {#2} }
+\cs_new_protected:Npn \@@_set:nn
+  { \exp_args:No \@@_set:nnn \l_@@_module_str }
 \cs_new_protected:Npn \@@_set:nnn #1#2#3
   {
-    \str_set:Nx \l_@@_module_str { \@@_trim_spaces:n {#2} }
+    \cs_set_nopar:Npx \l_@@_module_str { \@@_trim_spaces:n {#2} }
     \keyval_parse:NNn \@@_set_keyval:n \@@_set_keyval:nn {#3}
-    \str_set:Nn \l_@@_module_str {#1}
+    \cs_set_nopar:Npn \l_@@_module_str {#1}
```
blefloch, line 2402: Same as for `\@@_define:nnn`.

l3kernel/l3keys.dtx
```diff
@@ -2429,31 +2428,31 @@
...
-      \l_@@_unused_clist {#3} {#1} {#2} #4
+      \l_@@_unused_clist {#3} {#1} {#2}
   }
 \cs_generate_variant:Nn \keys_set_known:nnnN { nV , nv , no }
 \cs_new_protected:Npn \@@_set_known:nnnnN #1#2#3#4#5
   {
     \clist_clear:N \l_@@_unused_clist
     \@@_set_known:nnn {#2} {#3} {#4}
-    \tl_set:Nx #5 { \exp_not:o \l_@@_unused_clist }
+    \cs_set_eq:NN #5 \l_@@_unused_clist
```
blefloch, line 2447: Please use `\tl_set_eq:NN` now.  Frank clarified that he was complaining about the use of `\cs_...` to set a tl equal to a clist.  We will be documenting explicitly that clists are a subtype of tl, so that `\tl_set_eq:NN \l_my_tl \l_your_clist` is valid code.  The same holds for strings (not used in this pull request I think) and quarks.

l3kernel/l3keys.dtx
```diff
@@ -1519,21 +1519,21 @@
...
+  { \exp_args:No \@@_define:nnn \l_@@_module_str }
 \cs_new_protected:Npn \@@_define:nnn #1#2#3
   {
-    \str_set:Nx \l_@@_module_str { \@@_trim_spaces:n {#2} }
+    \cs_set_nopar:Npx \l_@@_module_str { \@@_trim_spaces:n {#2} }
     \keyval_parse:NNn \@@_define:n \@@_define:nn {#3}
-    \str_set:Nn \l_@@_module_str {#1}
+    \cs_set_nopar:Npn \l_@@_module_str {#1}
   }
-\cs_generate_variant:Nn \@@_define:nnn { o }
```
blefloch, line 1536: I'm ok with expanding `\@@_define:onn` because it is only used there and nowhere else.  This is not so much a speed optimization as an optimization of not needlessly introducing csnames.

The `\cs_set_nopar:Npx` speed optimizations are not worth the uglier code: this is typically done once per `\keys_define:nn`, hence a few times per package.  Optimizations concerning every key (like your earlier optimizations of `\keyval_parse:NNn` are more important.

On the other hand, maybe we should have a more general way to "set a tl variable to some value then restore the old value" as it is done here and in many other places in l3keys.
```
\cs_new_protected:Npn \keys_define:nn #1#2
  {
    \__kernel_with_tl_set:Nxn \l_@@_module_str
      { \@@_trim_spaces:n {#2} }
      { \keyval_parse:NNn \@@_define:n \@@_define:nn {#3} }
  }
\cs_new_protected:Npn \__kernel_with_tl_set:Nxn #1
  { \exp_args:No \__kernel_with_tl_set:nNxn {#1} #1 }
\cs_new_protected:Npn \__kernel_with_tl_set:nNxn #1#2#3#4
  {
    \tl_set:Nx #2 {#3}
    #4
    \tl_set:Nn #2 {#1}
  }
```

blefloch (commented):
Thanks.  Overall it's a bunch of minor speed ups, but mostly ok.  See my inline comments.

l3kernel/l3keys.dtx
Skillmon: I like the idea of `\__kernel_with_tl_set:Nxn` to unify this code, but that should go to another issue/PR.

l3kernel/l3keys.dtx
```diff
@@ -1530,9 +1530,9 @@
   { \exp_args:No \@@_define:nnn \l_@@_module_str }
 \cs_new_protected:Npn \@@_define:nnn #1#2#3
   {
-    \cs_set_nopar:Npx \l_@@_module_str { \@@_trim_spaces:n {#2} }
+    \str_set:Nx \l_@@_module_str { \@@_trim_spaces:n {#2} }
     \keyval_parse:NNn \@@_define:n \@@_define:nn {#3}
-    \cs_set_nopar:Npn \l_@@_module_str {#1}
+    \str_set:Nn \l_@@_module_str {#1}
```
blefloch, line 1535: I think you missed some related changes for `@@_set:nnn`.  You also missed some `\use:e`→`\use:x` and perhaps some more.  With `git diff HEAD~2` you can see the total changes from 2 commits prior (obviously works for more than 2) and thus more easily see what may need to be reconsidered after my review.  Let me know if you want me to be more thorough pointing out places.

Skillmon: I did this commit on a bus ride, so it is very likely I've missed things :) Would using `\__kernel_tl_set:Nx` be fine to set a `str` if one knows that the argument is already a string?

blefloch: I think it's ok (and I'm ready to defend this to Frank et al), since we document that strings are particular kinds of tl.

Skillmon:
I merged main, because `\__kernel_tl_set:Nx` was added in the meantime. I hope I spotted every evil line here.

josephwright:
I suspect we've taken the bits from here we are likely to: anyone object if I close?

josephwright:
I think we have decided against this one - the balance with code clarity doesn't quite work out.

Issue #781, opened 2020-07-23
expl3 cannot be explicitely loaded with latex-dev


dbitouze:
The following MCE:

```latex
\documentclass{article}
\usepackage{expl3}
\begin{document}
\end{document}
```

run with `latex-dev`:

>This is pdfTeX, Version 3.14159265-2.6-1.40.21 (TeX Live 2020) (preloaded format=latex-dev)

leads to the following error:

```
! Undefined control sequence.
l.67   \bool
            _new:N \g__expl_reload_bool
```

eg9:
I get this:

```tex
latex-dev  dberr
This is pdfTeX, Version 3.14159265-2.6-1.40.21 (TeX Live 2020) (preloaded format=latex-dev)
 restricted \write18 enabled.
entering extended mode
(./dberr.tex
LaTeX2e <2020-10-01> pre-release-7 (develop 2020-7-23 branch)
L3 programming layer <2020-07-17> xparse <2020-03-03>
(/usr/local/texlive/2020/texmf-dist/tex/latex-dev/base/article.cls
Document Class: article 2020/04/10 v1.4m Standard LaTeX document class
(/usr/local/texlive/2020/texmf-dist/tex/latex-dev/base/size10.clo))
(/usr/local/texlive/2020/texmf-dist/tex/latex/l3kernel/expl3.sty
(/usr/local/texlive/2020/texmf-dist/tex/latex/l3backend/l3backend-dvips.def))
(./dberr.aux) (./dberr.aux) )
No pages of output.
Transcript written on dberr.log.
```

josephwright:
@dbitouze That error happens when the format files get out of sync with the TeX tree: can you do `fmtutil-sys -byfmt pdflatex-dev`?

u-fischer:
if the problem persists check with `kpsewhich --all --engine=pdftex latex-dev.fmt` if you have some local format lurking around.

dbitouze:
@josephwright Same result. Even more weird: `pdflatex test` is okay but, despite a `fmtutil-sys -byfmt latex`, `latex test` gives the same error.

dbitouze:
@u-fischer
```
$ kpsewhich --all --engine=pdftex latex-dev.fmt
/home/bitouze/.texlive2020/texmf-var/web2c/pdftex/latex-dev.fmt
/usr/local/texlive/2020/texmf-var/web2c/pdftex/latex-dev.fmt
```
I will be offline from now on until tomorrow.

[...]

PR #787, opened 2020-08-06
Version check

PhelypeOleinik:
The proposal here adds a command `\__kernel_dependency_version_check:Nn` that checks if the loaded `expl3` kernel supports the dependency being loaded (dependency being `expl3.sty` itself, `l3backend`, or other files not part of the 'core' `expl3`).

`\__kernel_dependency_version_check:Nn \MinKernelDate { current-file }` checks if the version date of the `expl3` kernel currently loaded is at least `\MinKernelDate`, and if it's not an error is raised and `\endinput` is used.

This might be a redundant test in some cases, but in others it will save us from trying to figure out the cause of some apparently unrelated error, like in https://github.com/latex3/latex3/issues/781, or previous problems we had with a version mismatch between `l3backend` and `l3kernel`.

### DIFF
--- a/l3backend/l3backend-basics.dtx
+++ b/l3backend/l3backend-basics.dtx
@@ -69,26 +69,25 @@
 % place in the files.
 %    \begin{macrocode}
 %<*package>
-\def\ExplBackendFileDate{2020-06-29}
 \ProvidesExplFile
 %<*dvipdfmx>
-  {l3backend-dvipdfmx.def}{\ExplBackendFileDate}{}
+  {l3backend-dvipdfmx.def}{2020-06-29}{}
   {L3 backend support: dvipdfmx}
 %</dvipdfmx>
 %<*dvips>
-  {l3backend-dvips.def}{\ExplBackendFileDate}{}
+  {l3backend-dvips.def}{2020-06-29}{}
   {L3 backend support: dvips}
 %</dvips>
 %<*dvisvgm>
-  {l3backend-dvisvgm.def}{\ExplBackendFileDate}{}
+  {l3backend-dvisvgm.def}{2020-06-29}{}
   {L3 backend support: dvisvgm}
 %</dvisvgm>
 %<*pdfmode>
-  {l3backend-pdfmode.def}{\ExplBackendFileDate}{}
+  {l3backend-pdfmode.def}{2020-06-29}{}
   {L3 backend support: PDF mode}
 %</pdfmode>
 %<*xdvipdfmx>
-  {l3backend-xdvipdfmx.def}{\ExplBackendFileDate}{}
+  {l3backend-xdvipdfmx.def}{2020-06-29}{}
   {L3 backend support: xdvipdfmx}
 %</xdvipdfmx>
 %    \end{macrocode}
@@ -100,9 +99,9 @@
 %   With time, this test should vanish and only the dependency check
 %   should remain.
 %    \begin{macrocode}
-\cs_if_exist:NTF \__kernel_dependency_version_check:Nn
+\cs_if_exist:NTF \__kernel_dependency_version_check:nn
   {
-    \__kernel_dependency_version_check:Nn \ExplBackendFileDate
+    \__kernel_dependency_version_check:nn {2020-06-29}
 %<dvipdfmx>      {l3backend-dvipdfmx.def}
 %<dvips>      {l3backend-dvips.def}
 %<dvisvgm>      {l3backend-dvisvgm.def}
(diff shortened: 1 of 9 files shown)


l3backend/l3backend-basics.dtx
```diff
@@ -90,6 +90,33 @@
...
+%   Check if the loaded kernel is at least enough to load this file.
+%   The kernel date has to be at least equal to \cs{ExplBackendFileDate}
+%   or later.  If \cs{__kernel_dependency_version_check:Nn} doesn't
+%   exist we're loading in an older kernel, so it's an error anyway.
+%   With time, this test should vanish and only the dependency check
+%   should remain.
+%    \begin{macrocode}
+\cs_if_exist:NTF \__kernel_dependency_version_check:nn
+  {
+    \__kernel_dependency_version_check:nn {2020-06-29}
```
FrankMittelbach, line 104: By what process/method is the date  in \__kernel_dependency_version_check:nn set?

PhelypeOleinik: Manually.  As Joseph said in the [comment above](https://github.com/latex3/latex3/pull/787#commitcomment-41146960), the release cycle of `l3backend` is usually separate from `l3kernel`, so just setting the same date doesn't work, so the date has to be set manually to the version required.

FrankMittelbach: so I feared :-)  I wonder if that needs documentating (not sure where) and perhaps all moved to a separated filewhere the setting is all confined to one place.  I'm just wondering about how updates happen if they are done by somebody not doing them normally ... and how much fails because he/she doesn't know where to do adjustments.

PhelypeOleinik: Perhaps a `l3version.dtx` somewhere that contains docstrip guards for all dependencies.  The problem is that these files are under different locations in the source (and distribution) trees, so I don't know how well (or bad) that would work out. Or maybe this can be updated semi-automatically with `l3build`, as the release date is. @josephwright probably knows the most details about this process... Would something like that be possible?

That said, the worst that would happen if some file had the wrong minimum date would be an error deeper down in the code as we already have, so back at square one in that case.

FrankMittelbach:
Am 07.08.20 um 00:20 schrieb Phelype Oleinik:
> Perhaps a |l3version.dtx| somewhere that contains docstrip guards for 
> all dependencies. The problem is that these files are under different 
> locations in the source (and distribution) trees, so I don't know how 
> well (or bad) that would work out. Or maybe this can be updated 
> semi-automatically with |l3build|, as the release date is. @josephwright 
> <https://github.com/josephwright> probably knows the most details about 
> this process... Would something like that be possible?

I guess I would appreciate a single place  l3version.dtx or 
l3dependencies.dtx  where that lives nd is document (what to do) or 
where perhaps it can be (semi)automated (as long as you have a 
repositories locally next to each other

PhelypeOleinik:
> At the moment, there are several things in here that are not to do with the PR. Although we are normally quite relaxed about minor clean-up, there's a bit more than that going on, e.g. the `ior` stream change. Could you split those out so the only changes are directly to do with this topic?

The `ior` change is required because the now the stream is used by the error message code to run `kpsewhich`, so I removed the conditional loading (which seems that is always false in current engines).  Other than that there are some typo fixes... I'll see how to remove those commits from the PR.

FrankMittelbach:
> The `ior` change is required because the now the stream is used by the error message code to run `kpsewhich`, so I removed the conditional loading (which seems that is always false in current engines). Other than that there are some typo fixes... I'll see how to remove those commits from the PR.

I think it would be overkill to remove typos to add them separately. So I guess the only question here is is the ior change ok but if you say it is needed for the code then the answer to that decides if it should go in (@josephwright's call)

l3kernel/l3sys.dtx
```diff
@@ -361,6 +373,50 @@
...
+        lua \lua_now:e
+          {
+            if (pcall(require, 'luaharfbuzz')) then ~
+              tex.print("hb") ~
+            end
+          }
+      }
+    tex
+  }
+\str_const:Nx \c_sys_engine_format_str
```
josephwright, line 407: Doesn't this need some conditional code to deal with 'might not be LaTeX' cases, e.g. plain, ConTeXt, LuaMetaTeX, ...

josephwright (approved):
I'm happy to see this merged, but I think you might want to look again at the 'format' data and cover non-LaTeX cases (at least as far as leave it blank it we are not on LaTeX).

josephwright:
Actually, we can likely tidy up after merging: I'm just wondering whether to go to the effort of rebasing, squash 
or create a merge commit!

josephwright:
I decided to rebase ... give me a few minutes

PR #789, opened 2020-08-09
Faster l3tl

Skillmon:
This PR does slight adjustments to make `l3tl` faster. These adjustments include:

- use `\if:w` in favour of `\if_catcode:w` where applicable (`\if:w` is slightly faster)
- use `\if:w` in favour of `\if_meaning:w` where applicable (`\if:w` is faster than `\exp_after:wN \if_meaning:w`)
- preexpand `\__tl_if_empty_if:o` once where applicable
- use `\prg_do_nothing:` (or something equivalent with a private name) instead of a marker which is then gobbled using `\use_none:n` (expanding `\prg_do_nothing:` once is faster than expanding `\use_none:n <token>` once)
- reduce number of needed `\exp_after:wN` where applicable


Skillmon:
I've done some quick benchmarks (6 benchmark-runs per macro, target time 1s each) and these are the results for the macros (I didn't benchmark them all; each macro is benchmarked *without* the changes to the other macro, such that, e.g., `\__tl_act:NNNnn` doesn't profit from the faster `\tl_if_head...` tests in my version):

- `\tl_put_right:Nn`: 24 % (empty list, adding one), 6 % (empty list, adding
  100); similar results should be obtained for the other `\tl_put_...` and
  `\tl_concat:NN` functions in which `\exp_not:o` was switched out
- `\tl_if_empty:nTF`: 16 % (empty list or one token); roughly the same should be
  obtained for `\tl_if_empty:oTF` and `\tl_if_blank:nTF`
- `\tl_if_single:nTF`: 4 % (empty, one token, two tokens)
- `\__tl_act:NNNnn`: 2 % (one token, using `\use_none:nn` and `\use_none:n`)
- `\tl_head:n`: negligible without the changes to `\__tl_if_empty_if:o`
- `\tl_if_head_is_N_type:nTF`: 13 % (empty arg), 13 % (space), 1 % (single
  group), negligible (single N-type token)
- `\tl_if_head_is_group:nTF`: negligible (`\if:w` *is* faster than
  `\if_catcode:w`, but that's the only change and it does more than just that)
- `\tl_if_head_is_space:nTF`: 10 % (empty), 9 % (single token, no space), 10 %
  (space)

Skillmon:
If b7b5888 isn't accepted, 15a5705 should be reverted. Since `\exp_not:o` does grab its argument once before applying `\__kernel_exp_not:w \exp_after:wN`, it is actually slower for long token lists!

josephwright:
I'm happy here: let's see what @FrankMittelbach thinks

FrankMittelbach:
> I'm happy here: let's see what @FrankMittelbach thinks

Frank thinks that optimizing is more @blefloch 's domain :-) 

What worries me a bit is that I'm not sure how good our test coverage on that level is.  It think when we start doing serious refactoring of those low-level areas we should make sure that we have a better test coverage to ensure that border cases don't change or fail.

There is also the question about the benchmarks, especially with "tl"s there is a rather large likelihood of large tokenlists and and I'm not sure that all the optimizatons are actually remain optimal in such cases. 

But as I said not really my area. On first glance it looks ok, so if you are happy and Bruno has no concerns (assuming he has the time to look it over) I'm ok with it.

Skillmon:
@FrankMittelbach I deliberately didn't change the logic in any way, except for the two functions `\tl_if_head_is_N_type:n`, which now shortcuts if the first part of the test is true (meaning the tl is either empty or starts with a space) such that the test for a leading group is skipped if that one was true. The other macro in which I changed the logic is `\tl_if_head_is_space:n`, which used to use `\exp:w` and a bunch of `\exp_after:wN`. Now it uses the same structure as already used by `\tl_if_head_is_N_type:n` (so `\if:w` to start the expansion context, meaning it can use the shortcuts of `\prg_new_conditional:Npnn` for the `\prg_return_true: \else: \prg_return_false: \fi:` pattern). Also ending `\if:w` with either `f \scan_stop:` for the false or `\scan_stop: \scan_stop:` for the true branch is faster than ending `\exp:w` with `\exp_end:`, just to start another `\exp:w` for `\prg_return_true:` or `..._false:`.

All the other changes are switching out slower solutions for equivalent but faster ones (e.g., `\if:w` instead of `\if_catcode:w`, `\__kernel_exp_not:w \exp_after:wN` instead of `\exp_not:o`, etc.).

`\if:w \scan_stop: \tl_to_str:n {#1} \scan_stop:` is just as robust as `\exp_after:wN \if_meaning:w \exp_after:wN \q__tl_nil \tl_to_str:n {#1} \q__tl_nil`, but considerably faster.

In `\tl_if_single:n` I changed out a `\use_none:nn #1 ??` for a `\use_none:n #1 ?`, that is not behaving exactly the same if `#1` has a trailing space after a single `n` type argument, but that doesn't break the test as the next step (`\__tl_if_single:nnw`) uses an `n` type argument as its second argument (which is reinserted), so the space is ignored at that step.

There was indeed a change which would've made some macros slower for long token lists (those were the changes in 15a5705), but that one I corrected with b7b5888.

But I guess it's a good idea to let @blefloch have a look.

FrankMittelbach:
Am 11.08.20 um 22:03 schrieb Skillmon:
> I deliberately didn't change the logic in any way, except for ...

I believe you :-) but that wasn't my point.

I'm speaking from 30+ years experience here that there is nearly never 
such as thing as a harmless improvement or change with TeX (not even 
with expl3, if you go down to the bowls as you do).

This is why the regression test suite for 2e has saved the day many 
times over and over again for us. My point here is that many areas 
around the expl3 development have less test coverage than they should 
have (some have).

So my suggestion is that in the future the first step is going to be

  - examine and improve the test coverage to test all standard behavior 
and the boundary cases (often forgotten)

  - and only then attempt any change/improvement

The improvement is now already there, but we should still ensure not 
mentally but physically that it has 100% the same behavior (except being 
faster) than the current version.

With expl3 now in the kernel and *more importantly* with expl3 now being 
used in the kernel an update however harmless might suddenly break a 
huge number of user documents out there and that without any test phase 
through LaTeX-dev because an expl3 updates (at least currently) goes in 
on the sideline.

That is scary and I don't want any bad press coming from it, especially 
now that we are at the start of doing bigger changes to LaTeX and I'm 
telling people that we know what we are doing and it will be safe.

Skillmon:
@FrankMittelbach Very good explanation for the cautious approach to my PRs (I'm sorry, if I was borderline unfriendly in the thread of #786).

I'll look into setting up more tests hopefully next week (don't have too much time the rest of the week).

FrankMittelbach:
Am 11.08.20 um 22:33 schrieb Skillmon:
> @FrankMittelbach <https://github.com/FrankMittelbach> Very good 
> explanation for the cautious approach to my PRs (I'm sorry, if I was 
> borderline unfriendly in the thread of #786 

you were not, don't worry (another thing I learned in the last 3 decades 
is that email is not a good communication method and that one never 
should get angry directly about somebody writing something while 
"thinking aloud") it is missing the visual feedback to interpret.

anyway, I still owe you an explanation why I don't wan't \cs_... --- it 
is on my list but already burried by other stuff :-(

blefloch:
> Frank thinks that optimizing is more @blefloch <https://github.com/blefloch> 's
> domain :-)

True that.  Currently I'm on a one-week vacation so I'll have a lot of work when
I come back and I cannot review that until roughly 2 weeks.  I'm wondering if
there could be a more general "speed up the following def by pre-expanding some
things".

\cs_compile:NNpn
\cs_new_protected:Npn \tl_set:No #1#2
  { \cs_set_nopar:Npx #1 { \exp_not:o {#2} } }

with \cs_compile:NNpn defined to "inline" all variants (replacing \something:oox
by \exp_args:Noox \something:nnn) etc.

> But as I said not really my area. On first glance it looks ok, so if you are
> happy and Bruno has no concerns (assuming he has the time to look it over) I'm
> ok with it.

Most likely ok but I'd like to have a look.

FrankMittelbach:
>  But as I said not really my area. On first glance it looks ok, so if you are happy and Bruno has no concerns (assuming he has the time to look it over) I'm ok with it.
> Most likely ok but I'd like to have a look.

Then my suggestion would be that we wait for that as there is no hurry applying them (and eventually we will). That would also mean that Jonathan (it is Jonathan, isn't it?) would have some time to take a look how comprehensie the unit tests are and possibly extend them

car222222:
@FrankMittelbach wrote: 'there is a rather large likelihood of large tokenlists'

It would be useful to have some order of magnitude for how long a 'large token list' is here 10, 100, 1000, or 1,000,000 tokens ?  

This level of discussion should perhaps be moved to an issue.

Also, the general need for more and better test coverage in 'expl3(!!)' needs its own issue.

josephwright:
@car222222 While we don't have tests for _everything_, in general the test suite is not bad: between the explicit lower-level stuff and tests for usage at a higher level, we catch most issues.

Skillmon:
> it is Jonathan, isn't it?

Yes, it is.

Fine, then I'll take a look at the tests next week.

car222222:
@josephwright I, of course, have total faith in the test coverage, as much as I do in the use of language in the documentation!

It was Frank who questioned it.

Skillmon:
> It would be useful to have some order of magnitude for how long a 'large token list' is here: 10, 100, 1000, or 1,000,000 tokens ?

From what I've learned from all the benchmarking I've done is, that you can see whether a macro is really faster if you benchmark it once with only a handful of tokens and then again with a few thousand tokens. If it holds up in both cases chances are good that it will still keep up with a million tokens.

car222222:
@Skillmon That makes a lot of sense.

Skillmon:
Actually the tests were rather good, imho, so this should be ready to merge when you are (@blefloch?).

car222222:
We do not want to chance losing a fascinating idea introduced  above.

Should I transfer it to an issue, or are they too old-hat? 

@blefloch wrote here (some time ago) about some new(?) concepts, dubbed ‘pre-compilation’ that may have some applications around optimisation or elsewhere.
I reproduce his comment here.

—
I'm wondering if there could be a more general 
"speed up the following def by pre-expanding some things".

\cs_compile:NNpn
\cs_new_protected:Npn \tl_set:No #1#2
  { \cs_set_nopar:Npx #1 { \exp_not:o {#2} } }

with \cs_compile:NNpn defined to "inline" all variants (replacing
 \something:oox by \exp_args:Noox \something:nnn) etc.
—

l3kernel/l3tl.dtx
```diff
@@ -2271,26 +2327,26 @@
...
 %   \cs{__kernel_tl_to_str:w} are primitives that take care of
 %   expansion.
 %    \begin{macrocode}
 \prg_new_conditional:Npnn \tl_if_single:n #1 { p , T , F , TF }
   {
-    \if_catcode:w ^ \exp_after:wN \@@_if_single:nnw
+    \if:w \scan_stop: \exp_after:wN \@@_if_single:nnw
         \__kernel_tl_to_str:w
-          \exp_after:wN { \use_none:nn #1 ?? } ^ ? \s_@@_stop
+          \exp_after:wN { \use_none:n #1 ? } \scan_stop: ? \s_@@_stop
```
blefloch, line 2349: This optimization is wrong because a token list that is not blank may still have a string representation made of spaces: this happens for the control space <code>\ </code> when the escapechar is not printable; it also happens for space characters with weird category codes.  With your code, if what follows the first item in `#1` has a blank string representation, then it will be ignored by `\@@_if_single:nnw` just as if it were really blank in the token list.  The code will then deem the token list blank.  An example, which gives `\ERROR` with your new code:

    \char_set_lccode:nn { `a } { 32 }
    \lowercase { \tl_if_single:nT { b a } } { \ERROR }

On the other hand, I think `\@@_if_single:nnw` can be replaced by `\use_none:n` and `\scan_stop: ? \s_@@_stop` by `\prg_do_nothing: \scan_stop:`, but I am not sure whether this speeds things up (it replaces a three-parameter macro by a one-parameter one `\use_none:n` and a zero-parameter one `\prg_do_nothing:` in the true case).

l3kernel/l3tl.dtx
```diff
@@ -2792,13 +2851,12 @@
...
   {
-    \if_meaning:w \q_@@_act_mark #1
-      \exp_after:wN \@@_act_end:wn
-    \fi:
+    \@@_use_none_delimit_by_act_mark:w #1 \@@_act_end:wn \q_@@_act_mark
     #4 {#3} #1
     \@@_act_loop:w #2 \q_@@_act_stop
     {#3} #4
   }
+\cs_new:Npn \@@_use_none_delimit_by_act_mark:w #1 \q_@@_act_mark { }
```
blefloch, line 2859: I think it should be "`by_q_act_mark`".

l3kernel/l3tl.dtx
```diff
@@ -3087,36 +3162,50 @@
...
-%   the first two cases, the line involving \cs{@@_if_head_is_N_type:w}
-%   produces~|^| (and otherwise nothing).  In the third case
-%   (begin-group token), the lines involving \cs{exp_after:wN} produce a
+%   the first two cases, the line involving \cs{@@_if_head_is_N_type_auxi:w}
+%   produces~|f| (and otherwise nothing).  In the third case
+%   (begin-group token), the lines involving \cs{token_to_str:N} produce a
 %   single closing brace.  The category code test is thus true exactly
 %   in the fourth case, which is what we want.  One cannot optimize by
-%   moving one of the |*| to the beginning: if |#1| contains primitive
+%   moving one of the |\scan_stop:| to the beginning: if |#1| contains primitive
```
blefloch, line 3179: Replace `|\scan_stop:|` by `\cs{scan_stop:}`.

l3kernel/l3tl.dtx
```diff
@@ -3087,36 +3162,50 @@
...
+    \exp_after:wN { \if_false: } \fi: 
+  }
+\cs_new:Npn \@@_if_head_is_N_type_auxii:nn #1 #2
+  {
+    f
+  }
+\cs_new:Npn \@@_if_head_is_N_type_auxiii:n #1
+  {
+    \exp_after:wN \use_none:n \exp_after:wN 
   }
```
blefloch, line 3209: Perhaps replace `\exp_after:wN \@@_if_head_is_N_type_auxii:nn` by `f \exp_after:wN \use_none:nn`.
Perhaps put the code of `\@@_if_head_is_N_type_auxiii:n` in one line (and rename to `auxii`).

l3kernel/l3tl.dtx
```diff
@@ -3131,10 +3220,10 @@
...
   {
-    \if_catcode:w
+    \if:w
         \exp_after:wN \use_none:n
           \exp_after:wN { \exp_after:wN { \token_to_str:N #1 ? } }
-        * *
+        \scan_stop: \scan_stop:
       \prg_return_false:
     \else:
       \prg_return_true:
```
blefloch, line 3229: Conditionals are optimized when they end with `\prg_return_true: \else: \prg_return_false: \fi:` but not in the opposite order.  Here I think doing `\reverse_if:N \if:w ... \prg_return_true: \else: \prg_return_false: \fi:` would be faster despite the small overhead of `\reverse_if:N`.  This change may also be useful in a few other places, but I know I already tried to make many conditionals end in the right way to get the optimization.  Seems like I missed this one.

l3kernel/l3tl.dtx
```diff
@@ -3146,26 +3235,28 @@
...
+    \else:
+      \prg_return_false:
+    \fi:
   }
 \cs_new:Npn \@@_if_head_is_space:w #1 ~
   {
-    \tl_if_empty:oTF { \use_none:n #1 }
-      { \exp_after:wN \exp_end: \exp_after:wN \prg_return_true: }
-      { \exp_after:wN \exp_end: \exp_after:wN \prg_return_false: }
+    \tl_if_empty:oF {#1} { f }
```
blefloch, line 3259: Perhaps do `\@@_if_empty_if:o {#1} f \fi:`, and perhaps even the `\exp_args:Nno \use:n` construction you used a few times?  I wonder if we should simply define `\cs_new:Npo` and sprinkle that throughout most of `expl3`.

blefloch (approved):
Sorry for the long delay.  Besides a mistaken optimization in `\tl_if_single:nTF`, I only have minor nitpicks, and some ideas for further optimizations, especially defining `\cs_new:Npo` and using it throughout low-level `expl3` kernel code to save one expansion here and there at run time.

l3kernel/l3tl.dtx
Skillmon: This one is true, but I'm not sure whether it then should be "`by_q__tl_act_mark`" (which on the other hand looks weird for the double underscore), instead.

FrankMittelbach: whichever way you turn it, is is a bit odd. Perhaps be imprecise and call it `\@@_use_none_delimit_by_quark:w`.

Skillmon: @FrankMittelbach the version suggested by @blefloch is equivalent to the public ones in `l3basics` (except for being private to `l3tl` hinting at the argument being delimited by a private quark). I had these thoughts when I finally settled for leaving out the `q` instead to make it less ambiguous and circumventing the issue with the double underscore in the macro name. I have no strong opinion on this though, so more input by the team (or a definitive decision by someone) would be welcome.

blefloch: The question already arose when Phelype made the switch to mostly internal quarks.  See in `l3basics`

    \cs_new:Npn \@@_use_i_delimit_by_q_recursion_stop:nw
      #1 #2 \q_@@_recursion_stop {#1}

Skillmon: Oh, I wasn't aware of it, thank you for pointing it out!

l3kernel/l3tl.dtx
Skillmon: Replacing `\@@_if_single:nnw` with `\use_none:n` and putting `\prg_do_nothing: \scan_stop:` there would result in the test being true for a blank argument, so doesn't work as well.

blefloch: Oops, correct, so I guess the original code may be optimal.

l3kernel/l3tl.dtx
```diff
@@ -1269,6 +1269,13 @@
 %
 % \subsection{Functions}
 %
+% \begin{macro}{\__kernel_tl_set:Nx,\__kernel_tl_gset:Nx}
+%    \begin{macrocode}
+\cs_new_eq:NN \__kernel_tl_set:Nx \cs_set_nopar:Npx
+\cs_new_eq:NN \__kernel_tl_gset:Nx \cs_gset_nopar:Npx
```
PhelypeOleinik, line 1275: Could you please document these (as trivial as they are) in `l3kernel-functions.dtx`?

l3kernel/l3tl.dtx
FrankMittelbach: I have no strong feelings either way (my suggestion was just a side remark), What I don't like though would be double underscores in the middle of names, if we w have already established a convention elsewhere let's stick to that convention as suggested by Bruno and Phelype.

l3kernel/l3tl.dtx
blefloch: …and replace all(?) uses of `\tl_set:Nx` by `\__kernel_tl_set:Nx` (and same for `\tl_gset:Nx`).  This would speed up slightly `\tl_concat:NNN`, `\tl_replace_...`, `\tl_trim_spaces:N`, `\tl_reverse:N`.  In fact, maybe we should make that change much more broadly in `l3kernel`, which would for instance speed up `\clist_set:Nn` or `\seq_put_left:Nn`.

Skillmon: @blefloch I'd say yes, but this might break some assumptions of `l3debug` of macros using `\tl_set:Nx`/`\tl_gset:Nx` internally for stuff like the scope checks.

PhelypeOleinik: `l3debug` completely destroys performance by adding the checks, so since the “user interface” of `\__kernel_tl_set:Nx` and `\tl_set:Nx` are the same, I think it should be fine to restore the old definition of `\tl_set:Nx` and define `\__kernel_tl_set:Nx` in terms of that.  Or even easier, just add `\__kernel_tl_set:Nx` to be patched by `l3debug` the same way `\tl_set:Nx` is.

l3kernel/l3tl.dtx
Skillmon: Alternative idea: We change `\prg_new_conditional` to also optimize in this case.

blefloch: I don't like that very much, as it would require quite a bit more code than just putting a `\reverse_if:N`.  It's not hard, but to optimize this case in `\prg_new_conditional` we would need to implement the test for presence of `\prg_return_false: \else: \prg_return_true: \fi:` and implement specific code for the `p`, `T`, `F`, `TF` cases.

l3kernel/l3tl.dtx
Skillmon: It would have to be the reverse behaviour, so I would've to put `\exp_after:wN \reverse_if:N \@@_if_empty_if:o {#1} f \fi:` there, but that doesn't seem like a problem.

Yes the `\exp_args:Nno \use:n` construct feels strange when there could be a `\cs_new:Npo` variant.

blefloch: Indeed, I messed up the logic.  No need for `\reverse_if:N`, just use `\else:`, so do `\@@_if_empty_if:o {#1} \else: f \fi:`.

Skillmon: `\reverse_if:N` is faster than `\else:`, hence I'd actually go down that route.

Skillmon: Ok, there is some strangeness happening in `m3str002.lvt`. With `\else:` it works whereas with `\reverse_if:N` it fails, and I have no idea why.

blefloch:
On 9/19/20 10:27 AM, Skillmon wrote:
> *@Skillmon* commented on this pull request.
> 
> --------------------------------------------------------------------------------
> 
> In l3kernel/l3tl.dtx
> <https://github.com/latex3/latex3/pull/789#discussion_r491329935>:
> 
>>    }
>  \cs_new:Npn \@@_if_head_is_space:w #1 ~
>    {
> -    \tl_if_empty:oTF { \use_none:n #1 }
> -      { \exp_after:wN \exp_end: \exp_after:wN \prg_return_true: }
> -      { \exp_after:wN \exp_end: \exp_after:wN \prg_return_false: }
> +    \tl_if_empty:oF {#1} { f }
> 
> Ok, there is some strangeness happening in |m3str002.lvt|. With |\else:| it
> works whereas with |\reverse_if:N| it fails, and I have no idea why.

Ah, the problem is that

\reverse_if:N \if:w \scan_stop: \tl_to_str:n {#1} \scan_stop:

does not work correctly in case #1 is non-empty, because it will leave the rest
of #1 and \scan_stop: in the input stream, while

\if:w \scan_stop: \tl_to_str:n {#1} \scan_stop: \else:

correctly skips the rest of #1.

Skillmon:
@blefloch yes, I just realised the same.

Skillmon:
> I think actually the right strategy could be to make `l3debug` hook into `\__kernel_tl_set:Nx` and not anymore into `\tl_set:Nx`, but I haven't looked at `l3debug` in a while, so feel free to ignore this comment.

I didn't get this to work, failing many of the tests. The current approach is the only one that doesn't fail them. But I have another problem with using `\__kernel_tl_set:Nx` all over the place I just realised: The test suite might very well not really test this change, as many of the tests use `debug` changing `\__kernel_tl_set:Nx` to not have the same behaviour of `\cs_set_nopar:Npx` but instead the behaviour of the old `\tl_set:Nx`. I know that this is breaking `\tl_build_...` (hence I didn't apply it there) but I don't know whether this might break more exotic usages of `\tl_set:Nx` I might've overlooked.

l3kernel/l3tl.dtx
Skillmon: Using `\reverse_if:N` wouldn't work correctly, that'd keep a stray `\scan_stop:` in the input stream (same issue as with `\@@_if_head_is_space:w` below), so the only ways to further optimise this is either by changing `\prg_new_conditional` or by defining the `p`, `TF`, `T`, and `F` variants by hand here.

Skillmon: I'd really prefer changing `\prg_new_conditional` (and while at it, change it to not need `\exp_after:wN` but instead let it input something like `\use_i_iii:nnnn \fi: \use_ii:nn` (or, being even faster, some macro defined as `\cs_new:Npn \__prg_fast_fi_i:w \fi: \use_ii:nn #1 #2 { \fi: #1 }`).

blefloch: I would really like not to lengthen `\prg_new_conditional` just to optimize one or two conditionals.  One way out is to replace `\prg_return_false: \else: \prg_return_true: \fi:` by `\exp_after:wN \reverse_if:N \fi: \if_true: \prg_return_true: \else: \prg_return_false: \fi:` in this particular code.

On the other hand, you may be right that the precise code used in `\prg_new_conditional` could be optimized a bit.

PhelypeOleinik:
> I didn't get this to work, failing many of the tests. The current approach is the only one that doesn't fail them. But I have another problem with using `\__kernel_tl_set:Nx` all over the place I just realised: The test suite might very well not really test this change, as many of the tests use `debug` changing `\__kernel_tl_set:Nx` to not have the same behaviour of `\cs_set_nopar:Npx` but instead the behaviour of the old `\tl_set:Nx`. I know that this is breaking `\tl_build_...` (hence I didn't apply it there) but I don't know whether this might break more exotic usages of `\tl_set:Nx` I might've overlooked.

Not trying to be stubborn, but adding `\__kernel_tl_set:Nx` to the list of patches in `l3debug` seems really to be the way to go.  You just have to be careful, with debug on, to have `\__kernel_tl_set:Nx` be `\protected\long`, otherwise it will break if the token list contains a `\par`.  Changing that allows you to use `\__kernel_tl_set:Nx` in `l3tl-build` as well:

```diff
diff --git a/l3kernel/l3candidates.dtx b/l3kernel/l3candidates.dtx
index 76edde1..abf4677 100644
--- a/l3kernel/l3candidates.dtx
+++ b/l3kernel/l3candidates.dtx
@@ -1439,5 +1439,5 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \tl_build_get:NN
-  { \@@_build_get:NNN \tl_set:Nx }
+  { \@@_build_get:NNN \__kernel_tl_set:Nx }
 \cs_new_protected:Npn \@@_build_get:NNN #1#2#3
   { #1 #3 { \if_false: { \fi: \exp_after:wN \@@_build_get:w #2 } } }
@@ -1464,10 +1464,10 @@
 \cs_new_protected:Npn \tl_build_end:N #1
   {
-    \@@_build_get:NNN \tl_set:Nx #1 #1
+    \@@_build_get:NNN \__kernel_tl_set:Nx #1 #1
     \exp_args:Nc \@@_build_end_loop:NN { \cs_to_str:N #1 ' } \tl_clear:N
   }
 \cs_new_protected:Npn \tl_build_gend:N #1
   {
-    \@@_build_get:NNN \tl_gset:Nx #1 #1
+    \@@_build_get:NNN \__kernel_tl_gset:Nx #1 #1
     \exp_args:Nc \@@_build_end_loop:NN { \cs_to_str:N #1 ' } \tl_gclear:N
   }
diff --git a/l3kernel/l3debug.dtx b/l3kernel/l3debug.dtx
index 5511de5..d76f989 100644
--- a/l3kernel/l3debug.dtx
+++ b/l3kernel/l3debug.dtx
@@ -741,11 +741,4 @@
 % \end{macro}
 %
-% Set \cs{__kernel_tl_set:Nx} and \cs{__kernel_tl_gset:Nx} to the same as
-% \cs{tl_set:Nx}.
-%    \begin{macrocode}
-  \cs_gset_protected:Npn \__kernel_tl_set:Nx  { \tl_set:Nx }
-  \cs_gset_protected:Npn \__kernel_tl_gset:Nx { \tl_gset:Nx }
-%    \end{macrocode}
-%
 % Patching the second argument to ensure it exists. This happens before
 % we alter |#1| so the ordering is correct.  For many variable types
@@ -796,4 +789,15 @@
 %    \end{macrocode}
 %
+%   The debug code will take the macros in this list, copy, and redefine
+%   them to add the debugging code using the same prefix spec as the
+%   original macro.  \cs[no-index]{__kernel_tl_(g)set:Nx} is a copy of a
+%   \TeX{} primitive, so it has no prefix, which means that the debug
+%   patch will make it non-\cs{protected} and non-\cs{long}. To avoid
+%   that we define them as \cs{protected}\cs{long} here:
+%    \begin{macrocode}
+\cs_gset_protected:Npn \__kernel_tl_set:Nx { \cs_set_nopar:Npx }
+\cs_gset_protected:Npn \__kernel_tl_gset:Nx { \cs_gset_nopar:Npx }
+%    \end{macrocode}
+%
 % Patching where the first argument to a function needs scope-checking:
 % either local or global (so two lists).
@@ -842,4 +846,5 @@
       \tl_set:No
       \tl_set:Nx
+      \__kernel_tl_set:Nx
       \tl_set_eq:NN
       \tl_put_left:Nn
@@ -906,4 +911,5 @@
       \tl_gset:No
       \tl_gset:Nx
+      \__kernel_tl_gset:Nx
       \tl_gset_eq:NN
       \tl_gput_left:Nn
```

blefloch:
@Skillmon Is the pull request now ready?  If so I'll review the latest commits you made and we can merge.

Skillmon:
@blefloch  Yes, it is, thank you for your time!

Issue #582, opened 2019-05-27
Interfaces for \luadef (and \luafunction(call)?)


josephwright:
(Raised in https://github.com/latex3/latex3/pull/580.)

For some time, LuaTeX has provided `\luafunction` to allow pre-tokenization of heavily-used Lua functions. That was not that not so useful until the addition of `token.scan_string()` and more recently `token.scan_argument()`. With the addition of `\luadef` in LuaTeX v1.10, these combine to mean that anything significant using Lua should probably be defined that way.

There are two parts to the issue here. The first is whether or not to have an interface for `\luafunction` or just for `\luadef`: the latter more-or-less allows us to wrap up `\luafunction(call)` inside a csname without needing a macro. A toy example
```latex
\directlua{
  local t = lua.get_functions_table()
  t[1] = function()
      local a = token.scan_argument()
      tex.print("\string\\showtokens{" .. a .. "}")
    end
}
\luadef\luashow 1
\protected\def\baz{}
\luashow{foo bar\baz}
\show\luashow
```

Perhaps the cleanest way to provide an interface is only to make `\luadef` public and deal with the `\luafunction` stuff internally. That still leaves where to put it all. For example, is it `\lua_cs_new:Nn`/`\lua_cs_(g)set:Nn` (plus `\protected` variants). If so, how do we deal with the fact that stored Lua functions do have to be allocated: can be done without exposing that, but is that what we want?

(In format mode we have to work out how to cover some of what `ltluatex` does: that's a separate issue, probably can be covered reality easily.)

zauguin:
I see one problem with `lua_cs_new:Nn`, which is basically what to pass as second argument.
We could take the code of the to be invoked Lua function, but (especially optimized) Lua functions will normally always need additional context before the function. Another option would be to pass a function name defined in another file, but then the function would need to *have* a public name which will often be odd. After all these are functions which take their arguments by scanning things on the TeX level and not by real Lua arguments, so they are not useful interfaces from a Lua perspective.

Therefore I think I would prefer to only provide a nice wrapper for `set_lua` in Lua and not provide any TeX level interface.

hmenke:
No idea, whether anyone cares about my opinion, but I don't think an interface is necessary at all.  There is also not really a need for an allocator because the Lua function table knows its own length and hence if users abide by the following idiom it's impossible to clash:
```lua
local t = lua.get_functions_table()
t[#t + 1] = function()
    ...
end
token.set_lua("luashow", #t)
```
[...]

PR #790, opened 2020-08-10
Lua implementation for l3intarray

### LINKED ISSUE #788: LaTeX3 packages break in ConTeXt (LuaMetaTeX)
Since the last update, the following happens when expl3-generic is loaded with ConTeXt using the LuaMetaTeX engine:

```
tex error       > tex error on line 18105 in file expl3-code.tex: ! Font \c__fp_exp_intarray has only 7 fontdimen parameters

<recently read> \c__fp_exp_intarray 
                    
l.18105   }
   

<empty file>

Font parameter arrays are not automatically increased in this engine.
```

Marcel Krüger wrote about a related issue to the ConTeXt mailing list: https://www.mail-archive.com/ntg-context@ntg.nl/msg95273.html. What is curious is the fact the previous release (patching the code for catcode tables) worked fine both for LuaTeX and LuaMetaTeX. 

This remarks does not apply to ConTeXt with LuaTeX, as expl3-generic and friends work fine there.

zauguin:
An attempt to fix #788 and also an experiment about switching in Lua implementations for existing modules.

## TODO:
  - [x] The current system for defining the commands works with `luatexbase.new_luafunction`, which isn't available in ConTeXt. Add an abstraction for this allocator. (Related to #582)
  - [x] Fix the test. (The tests currently pass because the LaTeX2e version does not load ltluatex early enough, so the code is disabled. (See https://github.com/latex3/latex2e/pull/368) When testing with adapted 2e version, the test mostly work except for differences in behavior because of different limits.)
  - [x] Add a system for preserving intarrays created during format building for use in regular runs. An initial draft is part of this PR, but it's not used yet.

### DIFF
--- a/l3kernel/l3intarray.dtx
+++ b/l3kernel/l3intarray.dtx
@@ -198,7 +198,7 @@
 %
 %    \begin{macrocode}
 \cs_if_exist:NTF \@@_gset_count:Nw
-{
+  {
 %    \end{macrocode}
 %
 % \subsubsection{Allocating arrays}
@@ -206,8 +206,8 @@
 % \begin{variable}{\g_@@_table_int, \l_@@_bad_index_int}
 %   Used to differentiate intarrays in Lua and to record an invalid index.
 %    \begin{macrocode}
-\int_new:N \g_@@_table_int
-\int_new:N \l_@@_bad_index_int
+    \int_new:N \g_@@_table_int
+    \int_new:N \l_@@_bad_index_int
 %</tex>
 %    \end{macrocode}
 % \end{variable}
@@ -234,23 +234,23 @@ end, 'protected', 'global')
 %   check in \cs{intarray_new:Nn}.
 %    \begin{macrocode}
 %<*tex>
-\cs_new_protected:Npn \@@_new:N #1
-  {
-    \__kernel_chk_if_free_cs:N #1
-    \int_gincr:N \g_@@_table_int
-    \cs_gset_nopar:Npx #1 { \@@:w \int_use:N \g_@@_table_int \c_space_tl }
-  }
-\cs_new_protected:Npn \intarray_new:Nn #1#2
-  {
-    \@@_new:N #1
-    \@@_gset_count:Nw #1 \int_eval:n {#2} \scan_stop:
-    \int_compare:nNnT { \intarray_count:N #1 } < 0
+    \cs_new_protected:Npn \@@_new:N #1
       {
-        \msg_error:nnx { kernel } { negative-array-size }
-          { \intarray_count:N #1 }
+        \__kernel_chk_if_free_cs:N #1
+        \int_gincr:N \g_@@_table_int
+        \cs_gset_nopar:Npx #1 { \@@:w \int_use:N \g_@@_table_int \c_space_tl }
       }
-  }
-\cs_generate_variant:Nn \intarray_new:Nn { c }
+    \cs_new_protected:Npn \intarray_new:Nn #1#2
+      {
+        \@@_new:N #1
+        \@@_gset_count:Nw #1 \int_eval:n {#2} \scan_stop:
+        \int_compare:nNnT { \intarray_count:N #1 } < 0
+          {
+            \msg_error:nnx { kernel } { negative-array-size }
+              { \intarray_count:N #1 }
+          }
+      }
+    \cs_generate_variant:Nn \intarray_new:Nn { c }
 %</tex>
 %    \end{macrocode}
 % \end{macro}
@@ -338,7 +338,7 @@ end, 'global')
 %
 %    \begin{macrocode}
 %<*tex>
-\cs_generate_variant:Nn \intarray_count:N { c }
+    \cs_generate_variant:Nn \intarray_count:N { c }
 %</tex>
 %    \end{macrocode}
 % \end{macro}
@@ -381,17 +381,17 @@ end, 'protected', 'global')
 %   are within bounds.
 %    \begin{macrocode}
 %<*tex>
-\cs_new_protected:Npn \__kernel_intarray_gset:Nnn #1#2#3
-{ \@@_gset:w #2 #1 #3 \scan_stop: }
-\cs_new_protected:Npn \intarray_gset:Nnn #1#2#3
-  {
-    \@@_gset:wF \int_eval:n {#2} #1 \int_eval:n{#3}
+    \cs_new_protected:Npn \__kernel_intarray_gset:Nnn #1#2#3
+      { \@@_gset:w #2 #1 #3 \scan_stop: }
+    \cs_new_protected:Npn \intarray_gset:Nnn #1#2#3
       {
-        \msg_error:nnxxx { kernel } { out-of-bounds }
-          { \token_to_str:N #1 } { \int_use:N \l_@@_bad_index_int } { \intarray_count:N #1 }
+        \@@_gset:wF \int_eval:n {#2} #1 \int_eval:n{#3}
+          {
+            \msg_error:nnxxx { kernel } { out-of-bounds }
+              { \token_to_str:N #1 } { \int_use:N \l_@@_bad_index_int } { \intarray_count:N #1 }
+          }
       }
-  }
-\cs_generate_variant:Nn \intarray_gset:Nnn { c }
+    \cs_generate_variant:Nn \intarray_gset:Nnn { c }
 %</tex>
 %    \end{macrocode}
 % \end{macro}
@@ -409,7 +409,7 @@ luacmd('intarray_gzero:N', function()
 end, 'global', 'protected')
 %</lua>
 %<*tex>
-\cs_generate_variant:Nn \intarray_gzero:N { c }
+    \cs_generate_variant:Nn \intarray_gzero:N { c }
 %</tex>
 %    \end{macrocode}
 % \end{macro}
@@ -449,18 +449,18 @@ end, 'global')
 %
 %    \begin{macrocode}
 %<*tex>
-\cs_new:Npn \__kernel_intarray_item:Nn #1#2
-  { \@@_item:w #2 #1 }
-\cs_new:Npn \intarray_item:Nn #1#2
-  {
-    \@@_item:wF \int_eval:n {#2} #1
+    \cs_new:Npn \__kernel_intarray_item:Nn #1#2
+      { \@@_item:w #2 #1 }
+    \cs_new:Npn \intarray_item:Nn #1#2
       {
-        \msg_expandable_error:nnfff { kernel } { out-of-bounds }
-          { \token_to_str:N #1 } { \int_use:N \l_@@_bad_index_int } { \intarray_count:N #1 }
-        0
+        \@@_item:wF \int_eval:n {#2} #1
+          {
+            \msg_expandable_error:nnfff { kernel } { out-of-bounds }
+              { \token_to_str:N #1 } { \int_use:N \l_@@_bad_index_int } { \intarray_count:N #1 }
+            0
+          }
       }
-  }
-\cs_generate_variant:Nn \intarray_item:Nn { c }
+    \cs_generate_variant:Nn \intarray_item:Nn { c }
 %    \end{macrocode}
 % \end{macro}
 % \end{macro}
@@ -468,9 +468,9 @@ end, 'global')
 % \begin{macro}{\intarray_rand_item:N, \intarray_rand_item:c}
 %   Importantly, \cs{intarray_item:Nn} only evaluates its argument once.
 %    \begin{macrocode}
-\cs_new:Npn \intarray_rand_item:N #1
-  { \intarray_item:Nn #1 { \int_rand:n { \intarray_count:N #1 } } }
-\cs_generate_variant:Nn \intarray_rand_item:N { c }
+    \cs_new:Npn \intarray_rand_item:N #1
+      { \intarray_item:Nn #1 { \int_rand:n { \intarray_count:N #1 } } }
+    \cs_generate_variant:Nn \intarray_rand_item:N { c }
 %    \end{macrocode}
 % \end{macro}
 %
@@ -482,16 +482,16 @@ end, 'global')
 %   This is not implemented in Lua to ensure that the clist parsing is consistent
 %   with the clist module.
 %    \begin{macrocode}
-\cs_new_protected:Npn \intarray_const_from_clist:Nn #1#2
-  {
-    \@@_new:N #1
-    \int_zero:N \l_@@_loop_int
-    \clist_map_inline:nn {#2}
+    \cs_new_protected:Npn \intarray_const_from_clist:Nn #1#2
       {
-        \int_incr:N \l_@@_loop_int
-        \__kernel_intarray_gset:Nnn #1 \l_@@_loop_int { \int_eval:n {##1} } }
-  }
-\cs_generate_variant:Nn \intarray_const_from_clist:Nn { c }
+        \@@_new:N #1
+        \int_zero:N \l_@@_loop_int
+        \clist_map_inline:nn {#2}
+          {
+            \int_incr:N \l_@@_loop_int
+            \__kernel_intarray_gset:Nnn #1 \l_@@_loop_int { \int_eval:n {##1} } }
+      }
+    \cs_generate_variant:Nn \intarray_const_from_clist:Nn { c }
 %    \end{macrocode}
 % \end{macro}
 %
@@ -504,8 +504,8 @@ end, 'global')
 %   an issue since the delimiter is always just a comma or a comma and a space.
 %   In both cases \texttt{sprint(2, ...)} provides the right catcodes.
 %    \begin{macrocode}
-\cs_new:Npn \intarray_to_clist:N #1 { \@@_to_clist:Nn #1 { , } }
-\cs_generate_variant:Nn \intarray_to_clist:N { c }
+    \cs_new:Npn \intarray_to_clist:N #1 { \@@_to_clist:Nn #1 { , } }
+    \cs_generate_variant:Nn \intarray_to_clist:N { c }
 %</tex>
 %<*lua>
 local concat = table.concat
@@ -523,11 +523,11 @@ end, 'global')
 %   Loop through part of the array.
 %    \begin{macrocode}
 %<*tex>
-\cs_new:Npn \__kernel_intarray_range_to_clist:Nnn #1#2#3
-  {
-    \@@_range_to_clist:w #1
-    \int_eval:n {#2} ~ \int_eval:n {#3} ~
-  }
+    \cs_new:Npn \__kernel_intarray_range_to_clist:Nnn #1#2#3
+      {
+        \@@_range_to_clist:w #1
+        \int_eval:n {#2} ~ \int_eval:n {#3} ~
+      }
 %</tex>
 %<*lua>
 luacmd('@@_range_to_clist:w', function()
@@ -544,10 +544,10 @@ end, 'global')
 %   Loop through part of the array. We allow additional commas at the end.
 %    \begin{macrocode}
 %<*tex>
-\cs_new_protected:Npn \__kernel_intarray_gset_range_from_clist:Nnn #1#2#3
-  {
-    \@@_gset_range:w \int_eval:w #2 #1 #3 , , \scan_stop:
-  }
+    \cs_new_protected:Npn \__kernel_intarray_gset_range_from_clist:Nnn #1#2#3
+      {
+        \@@_gset_range:w \int_eval:w #2 #1 #3 , , \scan_stop:
+      }
 %</tex>
 %<*lua>
 luacmd('@@_gset_range:w', function()
@@ -579,9 +579,9 @@ luacmd('@@_gset_range:w', function()
 %   \cs{tex_ifabsnum:D}.
 %    \begin{macrocode}
 %<*tex>
-\cs_new_protected:Npn \@@_gset_overflow_test:nw #1
-  {
-  }
+    \cs_new_protected:Npn \@@_gset_overflow_test:nw #1
+    {
+    }
 %    \end{macrocode}
 % \end{macro}
 %
@@ -589,8 +589,8 @@ luacmd('@@_gset_range:w', function()
 %
 % Go to the false branch of the conditional above.
 %    \begin{macrocode}
-}
-{
+  }
+  {
 %    \end{macrocode}
 %
 % \subsubsection{Allocating arrays}
@@ -598,22 +598,22 @@ luacmd('@@_gset_range:w', function()
 % \begin{macro}{\@@_entry:w, \@@_count:w}
 %   We use these primitives quite a lot in this module.
 %    \begin{macrocode}
-\cs_new_eq:NN \@@_entry:w \tex_fontdimen:D
-\cs_new_eq:NN \@@_count:w \tex_hyphenchar:D
+    \cs_new_eq:NN \@@_entry:w \tex_fontdimen:D
+    \cs_new_eq:NN \@@_count:w \tex_hyphenchar:D
 %    \end{macrocode}
 % \end{macro}
 %
 % \begin{variable}{\c_@@_sp_dim}
 %   Used to convert integers to dimensions fast.
 %    \begin{macrocode}
-\dim_const:Nn \c_@@_sp_dim { 1 sp }
+    \dim_const:Nn \c_@@_sp_dim { 1 sp }
 %    \end{macrocode}
 % \end{variable}
 %
 % \begin{variable}{\g_@@_font_int}
 %   Used to assign one font per array.
 %    \begin{macrocode}
-\int_new:N \g_@@_font_int
+    \int_new:N \g_@@_font_int
 %    \end{macrocode}
 % \end{variable}
 %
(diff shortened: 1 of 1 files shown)


zauguin:
The older version was about three times slower (!) than the fontdimen version, the new version is much faster: For setting entries, we are about 30% faster than the fontdimen variant, about 10% faster for getting values and quite a bit slower (80%) for the kernel internal versions with less checking.

The non fontdimen verison has the additional advantage that it does not limit how big th eintegers are allowed to become.
(I guess that there is no big chance to get competitive performance for the kernel version because the fontdimen version does basically the same thing, except that it uses optimized C code instead of Lua...)

l3kernel/l3luatex.dtx
```diff
@@ -677,6 +677,84 @@ end
...
+%<@@=lua>
+%    \end{macrocode}
+%
+% The Lua state is not dumped when a forat is written, therefore any Lua
+% variables filled doing format building need to be restored in order to
+% be accessible during normal runs.
+%
+% We provide some kernel-internal helpers for this. They will only be available if
+% \texttt{luatexbase} is available. This is not a big restriction though, because
+% ConTeXt (which does not use \texttt{luatexbase}) does not load expl3 in the format.
```
josephwright, line 692: 'Con\TeX{}t', `\pkg{expl3}`

l3kernel/l3luatex.dtx
```diff
@@ -677,6 +677,84 @@ end
...
+% It accept a string which uniquely identifies the data object and has to be
+% provided to retrieve it later. Additionally it accepts a function which is
+% called in the \texttt{pre_dump} callback and which has to return a string that
+% evaluates to a valid Lua object to be preserved.
+%    \begin{macrocode}
+    local luadata, luadata_order = {}, {}
+
+    function register_luadata(name, func)
+      if luadata[name] then
+        error(string.format("LaTeX3 error: data name %q already in use", name))
```
josephwright, line 713: I think we are just using 'LaTeX error' nowadays

l3kernel/l3luatex.dtx
```diff
@@ -677,6 +677,84 @@ end
...
+% 
+% The actual work is done in \texttt{pre_dump}. The \texttt{luadata_order} is used
+% to ensure that the order is consistent over multiple runs.
+%    \begin{macrocode}
+luatexbase.add_to_callback("pre_dump", function()
+      if next(luadata) then
+        local str = "return {"
+        for i=1, #luadata_order do
+          local name = luadata_order[i]
+          str = string.format('%s[%q]=%s,', str, name, luadata[name]())
```
josephwright, line 729: Local copy of `string.format`?

l3kernel/l3luatex.dtx
```diff
@@ -677,6 +677,84 @@ end
...
+        end
+        lua.bytecode[register] = assert(load(str .. "}"))
+      end
+    end, "latex3.luadata")
```
josephwright, line 733: Are we using `latex3` for this? I thought just `ltx`?

josephwright (commented):
Some queries - nothing major so merging then looking at them is fine.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+% \end{macro}
+%
+% \subsubsection{Array items}
+%
+% \begin{macro}{\@@_gset:wNwF, \@@_gset:wNw}
+%   The setter provided by Lua. The argument order somewhat emulates the |\fontdimen|
+%   This has been chosen over a more conventional order to provide a delimiter for the numbers.
+%    \begin{macrocode}
+%<*lua>
+luacmd('@@_gset:wNwF', function()
```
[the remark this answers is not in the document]
blefloch: I'm with @zauguin on this one, but my opinion was a minority when we decided a few years back, so let's do what Joseph suggests.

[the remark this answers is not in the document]
zauguin: I think it would make sense to keep the `F` though to indicate that it behaves like a conditional, but I can drop that too if that's preferred.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+% \subsection{Lua implementation}
+% First, let's look at the Lua variant:
+%
+% We select the Lua version if the Lua helpers were defined. This can be detected by
+% the presence of \cs{@@_gset_count:Nw}.
+%
+%    \begin{macrocode}
+%<*tex>
+\cs_if_exist:NTF \@@_gset_count:Nw
+{
```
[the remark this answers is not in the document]
zauguin: I'll change this last since indenting the whole file makes working with diffs a pain.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+      }
+  }
+\cs_generate_variant:Nn \intarray_gset:Nnn { c }
+%</tex>
+%    \end{macrocode}
+% \end{macro}
+%
+% \begin{macro}{\intarray_gzero:N, \intarray_gzero:c}
+%   Set the appropriate \tn{fontdimen} to zero.  No bound checking
+%   needed.
```
blefloch, line 404: Stale comment.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+  local t = @@_table()
+  local item = t[i]
+  if item then
+    put_next(use_none)
+  else
+    tex.l_@@_bad_index_int = i
+    put_next(use_i)
+  end
+  sprint(-2, item or 0)
+end, 'global')
```
blefloch, line 439: This seems to leave something like `1234\use_none:n{false code}` in the input stream, instead of actually expanding the `\use_none:n` and getting just `1234`.  This matters in things like `\exp_after:Nf \iow_term:n { \intarray_item:Nn \g_tmpa_intarray { 12 } }`.

l3kernel/l3luatex.dtx
blefloch: Also `forat` should be `format` a few lines above.

zauguin:
@josephwright Thanks. I addressed most of the remarks. There is one question left about `\@@_gset:wF` vs `\@@_gset_aux:w` and then I'll indent everything and merge :)

l3kernel/testfiles/m3intarray001.luatex.tlg
```diff
@@ -22,26 +22,19 @@ This is a coding error.
...
-l. ...  }
-An attempt was made to store -2000000000 at position 1 in the array
-'\g_testa_intarray'. The largest allowed value -1073741823 will be used
-instead.
 ! LaTeX3 Error: Access to an entry beyond an array's bounds.
 For immediate help type H <return>.
  ...                                              
 l. ...  }
-An attempt was made to access or store data at position 13 of the array
+An attempt was made to access or store data at position 0 of the array
```
blefloch, line 37: This change is surprising and suggests that the LuaTeX version is losing track of the index number.

l3kernel/testfiles/m3intarray001.luatex.tlg
```diff
@@ -114,11 +75,11 @@ TEST 5: Zeroing
 ============================================================
 0,0,24,0,0,0,0,28,0,0
 The integer array \g_testd_intarray contains 10 items:
->  0, 0, 24, 0, 0, 0, 0, 28, 0, 0.
+> 0, 0, 24, 0, 0, 0, 0, 28, 0, 0.
 24
 0,0,0,0,0,0,0,20,0,0
 The integer array \g_testd_intarray contains 10 items:
->  0, 0, 0, 0, 0, 0, 0, 20, 0, 0.
+> 0, 0, 0, 0, 0, 0, 0, 20, 0, 0.
```
blefloch, line 82: Ideally the extra space would be avoided in the non-Lua case?

blefloch (commented):
I'm not really done but I have to stop.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+%   Declare |#1| as a tokenlist with the scanmark and a unique number.
+%   Pass the array's size to the Lua helper.
+%   Every \texttt{intarray} must be global; it's enough to run this
+%   check in \cs{intarray_new:Nn}.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_new:N #1
+  {
+    \__kernel_chk_if_free_cs:N #1
+    \int_gincr:N \g_@@_table_int
+    \cs_gset:Npx #1 { \s_@@ \int_use:N \g_@@_table_int \c_space_tl }
```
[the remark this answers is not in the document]
Skillmon: For kernel internal code there is the (marginally) faster `\__kernel_tl_set:Nx`.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+
+local use_i = token.create'use:n'
+if use_i.cmdname == "undefined_cs" then
+  use_i = nil
+else
+  use_none = token.create'use_none:n'
+end
+
+local scan_argument = token.scan_argument
+local @@_table do
```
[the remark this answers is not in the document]
zauguin: Lua doesn't care and I like doing this as a convention to indicate that the block is intended to initialize the local variable.
But we can split them, it's not really important.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+local s_@@ = token.create's_@@'
+local use_none
+local comma = token.create(string.byte',')
+
+local use_i = token.create'use:n'
+if use_i.cmdname == "undefined_cs" then
+  use_i = nil
+else
+  use_none = token.create'use_none:n'
+end
```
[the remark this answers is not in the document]
zauguin: That happens by default. Since the Lua modules get combined into one `expl3.lua` and then get loaded as one of the first steps, `\use:n` will almost certainly be undefined at this point. By itself that wouldn't be an issue since the token references the command sequence ad not it's definition, but if the document never contained a `\use:n` token before then LuaTeX will not yet have added it to it's hash table and `token.create` returns a useless marker.

We could avoid the test here by  always assuming that they are undefined, but trying to assign values to them makes it IMO more obvious that these will end up being their values.

l3kernel/testfiles/m3intarray001.luatex.tlg
zauguin: I agree and have a patch for that, but it's a bit hacky and I wanted to keep it separate from this PR since it's rather unrelated.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: Could you help me with an issue here? I tried changing this to `\__kernel_tl_gset:Nx` which complained about the variable not being declared. Is there some way to indicate to the debugging system to not check this instance?

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+%    \begin{macrocode}
+\int_new:N \g_@@_table_int
+\int_new:N \l_@@_bad_index_int
+%    \end{macrocode}
+% \end{variable}
+%
+%    \begin{macrocode}
+\msg_new:nnn { kernel } { negative-array-size }
+  { Size~of~array~may~not~be~negative:~#1 }
+%    \end{macrocode}
```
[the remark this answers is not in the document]
zauguin: Thanks, I moved it into the common part.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
Skillmon: Haven't thought of the debug-redefinition... In that case maybe use `\cs_gset_nopar:Npx` instead of `\cs_gset:Npx`. Not sure whether this would be the correct style.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
blefloch: Yes, `wF` should be fine and much clearer than just `w`.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
blefloch: Ah, I forgot about the debug code, sorry.  Using `\cs_gset_nopar:Npx` seems the best then, that's what's used in `\tl_const:Nx`.  We cannot use `\tl_const:Nx` directly because the variable name starts with `g_` not `c_`.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
blefloch: I knew Lua doesn't care; I didn't know about this convention, thanks.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
blefloch: Looking at the order, it seems indeed that `ltluatex` and all the lua code is loaded before `ltexpl` and all the `expl3` code.  This means that the lines we're talking about never do anything.  I suggest removing these lines and simply adding a comment saying that `use_i` and `use_none` will eventually mean `\use:n` and `\use_none:n`.  Alternatively, to find a way to have `\use:n` and `\use_none:n` in the token stream before that point, possibly with `token.create'csname'` and so on.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+  end
+  if register_luadata then
+    register_luadata('@@', function()
+      local t = "{[0]={},"
+      for i=1, #tables do
+        t = string.format("%s{%s},", t, table.concat(tables[i], ','))
+      end
+      return t .. "}"
+    end)
+  end
```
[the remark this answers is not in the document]
blefloch: Ok, sorry, I was misled by the fact that here `table` refers to the built-in, not to your local variable.

l3kernel/testfiles/m3intarray001.luatex.tlg
blefloch: Sounds good.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: > This means that the lines we're talking about never do anything

That's not quite true either because the Lua code is reloaded at `\everyjob` and at that point they are defined. But I'll find some other approach.

l3kernel/l3luatex.dtx
```diff
@@ -370,9 +370,10 @@ local scan_int     = token.scan_int or token.scan_integer
 local scan_string  = token.scan_string
 local scan_keyword = token.scan_keyword
 local put_next     = token.put_next
+local token_create = token.create
 
-local true_tok     = token.create'prg_return_true:'
-local false_tok    = token.create'prg_return_false:'
+local true_tok     = token_create'prg_return_true:'
+local false_tok    = token_create'prg_return_false:'
 %    \end{macrocode}
```
blefloch, line 377: `true_tok` and `false_tok` seem unused.  Besides, don't they end up being undefined cs for the same reason as you explained for `use_none` and `use_i`?

l3kernel/l3luatex.dtx
```diff
@@ -677,6 +677,85 @@ end
...
+      luadata[name] = func
+      luadata_order[#luadata_order + 1] = func and name
+    end
+%    \end{macrocode}
+% \end{macro}
+% 
+% The actual work is done in \texttt{pre_dump}. The \texttt{luadata_order} is used
+% to ensure that the order is consistent over multiple runs.
+%    \begin{macrocode}
+luatexbase.add_to_callback("pre_dump", function()
```
blefloch, line 725: Wrong indent.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,454 @@
...
+%
+% \begin{variable}{\g_@@_table_int, \l_@@_bad_index_int}
+%   Used to differentiate intarrays in Lua and to record an invalid index.
+%    \begin{macrocode}
+\int_new:N \g_@@_table_int
+\int_new:N \l_@@_bad_index_int
+%    \end{macrocode}
+% \end{variable}
+%
+% \begin{macro}{\s_@@}
```
blefloch, line 216: Given that `\s_@@` is not actually a scan mark (which are equal to `\relax`), it should probably be called something else like the `\@@_chk:w` used in `l3fp`.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,454 @@
...
+%
+% \begin{macro}{@@_table}
+%   Internal helper to scan an intarray token, extract the associated
+%   Lua table and return an error if the input is invalid.
+%    \begin{macrocode}
+%<*lua>
+
+local scan_token = token.scan_token
+local scan_toks = token.scan_toks
+local get_csname = token.get_csname
```
blefloch, line 277: Seems `scan_toks` and `get_csname` are unused.  Same for `scan_argument` later on.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,454 @@
...
+  if item then
+    put_next(use_none)
+  else
+    tex.l_@@_bad_index_int = i
+    put_next(use_i)
+  end
+  put_next(expand_after_scan_stop)
+  scan_token()
+  sprint(-2, item or 0)
+end, 'global')
```
blefloch, line 447: This is not suitable yet in the false branch: because f-expansion gives `0\msg_expandable_error...` instead of producing the error first.  One idea is to avoid `sprint(0)` in the error case and simply put a `0` on the TeX side after the error.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+  local i = scan_int()
+  local t = @@_table()
+  if t[i] then
+    t[i] = scan_int()
+    put_next(use_none)
+  else
+    scan_int()
+    put_next(use_i)
+  end
+end, 'protected', 'global')
```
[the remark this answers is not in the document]
blefloch: In fact `@@_gset:w` should simply do no check at all I think, directly `local i=scan_int() @@_table()[i]=scan_int()`.  Then it can be used in some functions like setting from a clist, below.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: Thanks, with `_nopar` it works fine.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,454 @@
...
+%   to count the elements first. Otherwise it is basically the same as
+%   \cs{intarray_new:Nn} (which we don't use because it would complain about the
+%   wrong qualifier in debugging mode)
+%   This is not implemented in Lua to ensure that the clist parsing is consistent
+%   with the clist module.
+%    \begin{macrocode}
+\cs_new_protected:Npn \intarray_const_from_clist:Nn #1#2
+  {
+    \@@_new:N #1
+    \@@_gset_count:Nw #1 \clist_count:n {#2} \scan_stop:
```
blefloch, line 496: This step is wasteful and can be simply omitted I think provided `\__kernel_intarray_gset:Nnn` just does `t[i]=...` with no tests, extending the table as needed.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: The new version introduces the tokens explicitly.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,454 @@
...
+%   Loop through part of the array.
+%    \begin{macrocode}
+%<*tex>
+\cs_new:Npn \__kernel_intarray_range_to_clist:Nnn #1#2#3
+  {
+    \exp_after:wN \@@_range_to_clist:ww
+    \int_value:w \int_eval:w #2 \exp_after:wN ;
+    \int_value:w \int_eval:w #3
+    #1
+  }
```
blefloch, line 538: Perhaps just `\@@_range_to_clist:w #1 \int_eval:n {#2} ~ \int_eval:n {#3} ~` and then below `sprint(-2, concat(@@_table(), ',', scan_int(), scan_int()))`.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,454 @@
...
+    t[from] = scan_int()
+    scan_token()
+    from = from + 1
+  end
+  end, 'global', 'protected')
+%</lua>
+%    \end{macrocode}
+% \end{macro}
+%
+% \begin{macro}{\intarray_show:N, \intarray_show:c, \intarray_log:N, \intarray_log:c}
```
blefloch, line 584: The definition of `\intarray_show:N` etc seems identical in all engines, should be shared.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: I renamed the variable to hopefully make this less confusing and added some remarks about register_luadata/luadata.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: Fixed

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+
+luacmd('intarray_count:N', function()
+  sprint(-2, #@@_table())
+end, 'global')
+%</lua>
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+%<*tex>
+\cs_generate_variant:Nn \intarray_count:N { c }
```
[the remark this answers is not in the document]
zauguin: I tried moving it, but it feels rather artificial. I feel like the value of keeping the definition of `\intarray_count:N` and `\intarray_count:c` together seems higher than the value of avoiding duplication here. (Of course the same question appears in other cases of `\cs_generate_variant:Nn`.

l3kernel/l3luatex.dtx
zauguin: They are used in `l3token` and defined here since they were considered to be generally useful. They would be undefined, but `l3bootstrap` explicitly makes sure that they appear to avoid that.

l3kernel/l3intarray.dtx
zauguin: Maybe just `\__intarray:w`?

l3kernel/l3intarray.dtx
zauguin: The definition of `\@@_show:NN` will probably be slightly different once the second space at the beginning is fixed, therefore it has been kept separate. The `\intarray_show:N` etc. could still be merged but then `\@@_show:NN` would need separate documentation which would effectively just be a textual version of the definition of `\intarray_show:N` and `\intarray_log:N`, so it seemed useful to keep them here too.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
blefloch: You're right actually, sorry.

l3kernel/l3luatex.dtx
blefloch: Should we have a wrapper like (untested code, and doesn't deal correctly with crazy category codes; it should `tex.print` some of it with the `-2` catcodetable)
```
token_create_safe = function(s)
  tex.runtoks(function()
    tex.sprint("\\edef\\__luatex_dummy:{\z
        \\expandafter\\noexpand\\csname" .. s .. "\\endcsname\z
    }\\endgroup")
  end)
  return token_create(s)
end
```

l3kernel/l3intarray.dtx
blefloch: Ok, if you prefer.

l3kernel/l3intarray.dtx
blefloch: I think the fix should be in `\@@_to_clist:Nn` for non-LuaTeX, such as adding an argument `#4` to `\@@_to_clist:w`, which is initially `\use_none:n` (when called from the main `_to_clist:Nn` and later is `\exp_not:n` (when called from `_to_clist:w`).  Many other ways obviously.

l3kernel/l3luatex.dtx
zauguin: I like the idea, but would probably implement it based on `token.set_char` or so to avoid having to deal with catcode issues.

blefloch: I don't understand how to avoid leaving the weird meaning in place on the TeX side when using `token.set_char`.  Is there a `token.set_undefined` or similar (I found nothing in the manual)?

zauguin: There isn't. The idea would be to first run `\begingroup`, then `token.set_char` and finally `\endgroup`. That's a bit of a pain, but since other approaches probably require groups too it seems reasonable.

zauguin: I added a `token_create_safe`  based on`set_char` and `let`.

l3kernel/l3intarray.dtx
zauguin: Sounds reasonable, so I moved the `show` functions to the shared part.

l3kernel/l3intarray.dtx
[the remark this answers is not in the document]
zauguin: Should be fixed.

l3kernel/l3intarray.dtx
```diff
@@ -173,7 +173,529 @@
...
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+% \end{macro}
+%
+% \subsection{Font dimension based implementation}
+%
+% Go to the false branch of the conditional above.
+%    \begin{macrocode}
+}{
```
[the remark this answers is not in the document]
zauguin: Done

josephwright:
We good-to-go here now?

l3kernel/l3luatex.dtx
blefloch: Just looked at the code for `token_create_safe`.  It's really neat!

blefloch:
As far as I can tell, @zauguin has applies all suggested changes, and all conversations are resolved, so we should be good to merge if he is ok with it.

PR #798, opened 2020-08-25
Split pdfmode

josephwright:
This PR splits the `l3backend-pdfmode` file into two. After a bit of thought, I feel we want two entirely separate files as we are likely to see more Lua code for PDF support, colour, etc., to the point where a joint file is unworkable. We also may need to allow for LuaMetaTeX, where we really need to use Lua interfaces.

The new files are called `l3backend-pdftex` and `l3backend-luatex`: I couldn't find a better way to have short names but include the idea these are 'engine built-ins for PDF creation'.


car222222:
This split sounds sensible but what pdf-producing backends do we now have?  

Just a list of names would be useful, or a pointer.
Thanks.

josephwright:
On 26/08/2020 05:39, Chris Rowley wrote:
> This split sounds sensible but what pdf-producing backends do we now have?
> 
> Just a list of names would be useful, or a pointer.
> Thanks.
> 

Directly:

- pdfTeX in PDF mode
- LuaTeX in PDF mode

Indirectly

- dvips + ps2pdf
- dvips + Distiller
- dvipdfmx
- xdvipdfmx

car222222:
OK: so are all the 'indirect methods' combined in one file?
If so, does the name of the file make it clear that it is 'every backend that uses specials'?

car222222:
And how are other (non-pdf) backend files named?

josephwright:
On 26/08/2020 07:54, Chris Rowley wrote:
> And how are other (non-pdf) backend files named?
> 

For expl3, all l3backend-<name>.def, <name> one of

- dvips
- dvipdmx
- luatex
- pdftex
- xdvipdfmx

plus dvisvgm (doesn't make PDF at all).

Joseph

car222222:
I am getting confused!  

What do 'dvips' and 'dvipsmx/dvipdfmx' contain?  Is it stuff for both ps and pdf (via distiller or something else)?

I am just trying to get a system of names that makes it somewhat clear what these files do.  

Another small question: is it totally well-known that -luatex must imply pdf output?

josephwright:
On 26/08/2020 09:58, Chris Rowley wrote:
> What do 'dvips' and 'dvipsmx/dvipdfmx' contain?  Is it stuff for both ps and pdf (via distiller or something else)?

No. dvipdfmx takes DVI and makes PDF directly itself, so supports it's 
own specials plus a *subset* of PostScript and dvips specials. No 'full 
fat' PostScript is involved. In contrast, dvips knows it's own specials, 
and raw PostScript is passed 'to the next step'. So dvips targets 
*PostScript* and dvipdfmx targets *PDF*.

(Caveat: some PDF structures are hard or impossible to set up in 
PostScript, and it is possible to include in the dvips file code that 
tells PostScript to pass info straight to PDF.)

> I am just trying to get a system of names that makes it somewhat clear what these files do.

The names are very long-standing: other than the fact I choice 
"xdvipdfmx" instead of "xetex", they are based on the graphics-def ones. 
For direct PDF output, pdftex.def was the 'classical' name but I choice 
"pdfmode" as I *knew* it would also cover LuaTeX. However, that was 
before Hans made major changes to LuaTeX: over time the two have drifted 
apart.

> Another small question: is it totally well-known that -luatex must imply pdf output?

I was trying to capture the 'real' name of the backend producing the 
output (see "xetex" vs "xdvipdfmx"). That's why I originally choice 
"pdfmode", but it's no longer really viable. We could have 
"l3backend-pdfmode-pdftex", "l3backend-pdfmode-luatex", but I felt this 
got a bit wordy.

Joseph

car222222:
'I am just trying to get a system of names that makes it somewhat clear what these files do.'

A splendid aim.

So it is now (and always has been) somewhat difficult in general to deduce from the name of a backend file exactly what it output, or how it does it.  In particular, although dvips correctly indicates  that its initial output is ps, the final,destination may we'll be pdf.

The only one of your current names that does include its primary output in the name is -luatex.  That is why I asked if omitting 'pdf' from this name is OK since 'everyone knows that luatex is conjoint with pdf ouput'.  Otherwise, maybe a good idea to add to the name, giving pdf-luatex or luatexpdf (hyphenated or not?).

josephwright:
The names I chose were meant to reflect the program targetting by the `\special`s, so `dvips` doesn't mention 'PostScript', for example. As we are talking low-level files for the kernel to load, the naming isn't also that visible. As I said, I originally went for `pdfmode` as a shortcut for 'direct PDF output from pdfTeX or LuaTeX', but that's not really quite as simple today: the primitives have different names, etc. So as I said, we could keep `pdfmode` but I'm not sure it's helpful.

car222222:
'dvips' mentions 'ps' which I guess most people associate with PostScript.  That is what I meant.  Probably documenting what each file covers is more important than the names.

u-fischer:
> The names are very long-standing: other than the fact I choice "xdvipdfmx" instead of "xetex",

personally I think xetex would make more sense. dvipdfmx and xdvipdfmx are the same application, so the difference in the two backend files don't address differences here, instead the xdvipdfmx file address specific xetex primitives, like e.g. Xetexpicfile or \pdfpagewidth. 

Beside this imho using the "standard" names is sensible. It is easy to understand that hdvips.def, dvips.def and l3backend-dvips.def have the same target and that you can select them with the "dvips" option.

car222222:
"xetex would make more sense".  I had wondered about that too, but again it has no obvious implication of 'pdf output'.

josephwright:
@u-fischer  My reasoning was that the programs involved are called as `dvipdmx` and `xdvipdfmx`:  XeTeX does _not_ interpret the specials.

josephwright:
I do wonder how important the names are: other than in `\listfiles`, they are basically an internal detail. I'd like people to be 'happy' with the names, but don't want to get too bogged down: we don't in the end control the program toolchains.

josephwright:
I do wonder how important the names are: other than in `\listfiles`, they are basically an internal detail. I'd like people to be 'happy' with the names, but don't want to get too bogged down: we don't in the end control the program toolchains.

josephwright:
@u-fischer  Of course, it the overall feeling is to change, we can: it's no big issue.

davidcarlisle:
On Sat, 29 Aug 2020 at 17:51, Joseph Wright <notifications@github.com>
wrote:

> @u-fischer <https://github.com/u-fischer> Of course, it the overall
> feeling is to change, we can: it's no big issue.
>


I don't think the name matters but as we are discussing it anyway I think
xetex is probably better than xdvipdfmx as a name as
\tex_XeTeXpicfile:D means it can only work with xetex.

In general, with regard to Chris's comments about making the eventual file
format clearer  I don't think the name is naming the final format just the
API used between the
engine and the dvi driver (or its own backend) eg  -dvips doesn't
necessarily mean you are making PostScript, just that you are using dvips
compatible \special
(which covers a lot of dvi drivers eg xdvi so long as you are not too
adventurous in literal postscript specials)

so I'd split up pdfmode into pdftex and luatex (which we had to do in
graphics-def and hyperref previously as well) and possibly rename xdvipdfmx
to xetex.

car222222:
Thanks for this clarification @davidcarlisle:
"the name is (not) naming the final format 
just the API used between the engine and the dvi driver,
or its (the engine's) own backend".

If this is clearly documented (for those who need to know), together with a reference to the documentation of that API, then the names do not matter.

Not important, but are you here thinking of "xetex" as 'having its own backend' because the dvipdf driver is closely integrated?

josephwright:
@car222222 `dvipdfmx` and `xdvipdfmx` are today the same program, simply operating on `dvi` and `xdv`, respectively. Thus one could use the `l3backend-dvipdmfx.def` file with XeTeX (I do when testing). The difference comes in that XeTeX has two primitives for measuring graphics, whereas pdfTeX, etc., do not, so measurement is done using `extractbb`. That needs (restricted) shell escape: today people likely don't even notice that.

So as @davidcarlisle says, one could argue for `xetex` here, although the 'backend work' is being done by `xdvipdfmx` (it's interpreting the specials, etc.).

josephwright:
Seems we have agreement on the pdfTeX/LuaTeX stuff, and I guess @davidcarlisle has a point about XeTeX, so I'll rebase/fix conflicts, do the XeTeX part and then merge. This needs an update to `l3backend` and `l3kernel` (As filenames change), so I'll get a release out (plus update the new date-in-backend business).

PR #812, opened 2020-09-30
Register prefix (platex)

aminophen:
Starting from platex 2020-09-30, we decided to use "platex".


PhelypeOleinik:
Thanks for registering!

PR #819, opened 2020-10-31
[WIP] Implement \tl_map_inline:npn & \tl_map_function_unbraced:Nn

bb010g:
# [WIP] Implement `\tl_map_function_unbraced:Nn`

With variants `\tl_map_function_unbraced:NN` and `\tl_map_function_unbraced:cN`.

Also implement `\tl_map_function_unbraced:Nn`.

These are basically the same as the `\tl_map_function:Nn` family, but without `#1` receiving `#2` braced. Careful use of this macro allows for cleanly mapping multi-argument functions.

**TODO:** The docs here are sucky, but I wrote this months ago and haven't been LaTeXing enough to feel out better wording that isn't incorrect. This is a useful macro, even if a bit awkward & subtle to describe.

---

# [WIP] Implement `\tl_map_inline:npn`

With internal `\@@_map_inline:nnn` & variants `\tl_map_inline:Npn` and `\tl_map_inline:cpn`.

It's like `\tl_map_inline:nn`, but you can supply a consistent number of arguments (of your choosing) in the token list, match your parameters, and map using multiple arguments instead of just one! It's really handy when factoring out slightly more complex templates.

Theoretically, `\tl_map_inline:nn` could be a simple wrapper of this, but do we want to rely on `\tl_map_function_unbraced:Nn` when we don't have to?

---

Finally got around to extracting this fiddle from a personal package. This PR probably has problems, especially on the documentation side, but at least it's out and other better at TeX than me can word stuff better? (The concepts themselves aren't that complex, just kinda foreign or uncommon IMO.)

Some example uses from that package:

```tex
% \begin{macro}{\keyval_parse_inline:Nnnn}
% \begin{macro}{\keyval_gparse_inline:Nnnn}
%    \begin{macrocode}
\cs_if_free:NT \keyval_parse_inline:Nnnn
  {
    \tl_map_inline:npn
      {
        {
          \keyval_parse_inline:Nnnn
          \tl_build_begin:N \tl_build_put_right:Nx \tl_build_end:N
        }
        {
          \keyval_gparse_inline:Nnnn
          \tl_build_gbegin:N \tl_build_gput_right:Nx \tl_build_gend:N
        }
      }
      #1#2#3#4
      {
        \cs_new_protected:Npn #1 ##1##2##3##4
          {
            #2 ##1
            \keyval_parse_inline:nnn
              { #3 ##1 {##2} }
              { #3 ##1 {##3} }
              {##4}
            #4 ##1
          }
      }
  }
%    \end{macrocode}
% \end{macro}
% \end{macro}
```

```tex
% \begin{macro}[pTF]{\ecv_string_if_exist:n}
% \begin{macro}[pTF]{\ecv_string_if_empty:n}
%    \begin{macrocode}
\tl_map_inline:nn { {exist} {empty} }
  {
    \tl_map_inline:npn
      { { {_p} {} } { {} {T} } { {} {F} } { {} {TF} } } ##1##2
      {
        \cs_new:cpx { ecv_string_if_ #1 ##1 : n ##2 } ####1
          {
            \exp_not:c { tl_if_ #1 ##1 : c ##2 }
              { l_@@_string_ ~ ####1 _tl }
          }
      }
  }
%    \end{macrocode}
% \end{macro}
% \end{macro}
```

This honestly just feels like a natural extension of the existing `\tl_map_inline:nn` API.

---
This change is [Reviewable](https://reviewable.io/reviews/latex3/latex3/819)

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -550,6 +550,29 @@
 %   \texttt{n}-type arguments. See also \cs{tl_map_function:NN}.
 % \end{function}
 %
+% \begin{function}[updated = 2020-10-30, rEXP]
+%   {\tl_map_function_unbraced:NN, \tl_map_function_unbraced:cN}
+%   \begin{syntax}
+%     \cs{tl_map_function_unbraced:NN} \meta{tl~var} \meta{function}
+%   \end{syntax}
+%   Applies \meta{function} to every \meta{item} in the \meta{tl~var}.
+%   The \meta{function} receives one unbraced argument for each iteration.
+%   This may be a number of tokens if the \meta{item} was stored within
+%   braces. Hence the \meta{function} should anticipate receiving
+%   \texttt{n}-type arguments. See also \cs{tl_map_function_unbraced:nN}.
+% \end{function}
+%
+% \begin{function}[updated = 2020-10-30, rEXP]{\tl_map_function_unbraced:nN}
+%   \begin{syntax}
+%     \cs{tl_map_function_unbraced:nN} \Arg{token list} \meta{function}
+%   \end{syntax}
+%   Applies \meta{function} to every \meta{item} in the \meta{token list},
+%   The \meta{function} receives one unbraced argument for each iteration.
+%   This may be a number of tokens if the \meta{item} was stored within
+%   braces. Hence the \meta{function} should anticipate receiving
+%   \texttt{n}-type arguments. See also \cs{tl_map_function_unbraced:NN}.
+% \end{function}
+%
 % \begin{function}[updated = 2012-06-29]
 %   {\tl_map_inline:Nn, \tl_map_inline:cn}
 %   \begin{syntax}
@@ -2479,6 +2502,32 @@
     #1 {#2} \@@_map_function:Nn #1
   }
 \cs_generate_variant:Nn \tl_map_function:NN { c }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\tl_map_function_unbraced:nN}
+% \begin{macro}{\tl_map_function_unbraced:NN, \tl_map_function_unbraced:cN}
+% \begin{macro}{\@@_map_function_unbraced:Nn}
+%   Expandable loop macro for token lists. These have the advantage of not
+%   needing to test if the argument is empty, because if it is, the stop
+%   marker is read immediately and the loop terminated.
+%    \begin{macrocode}
+\cs_new:Npn \tl_map_function_unbraced:nN #1#2
+  {
+    \@@_map_function_unbraced:Nn #2 #1
+      \q_@@_recursion_tail
+    \prg_break_point:Nn \tl_map_break: { }
+  }
+\cs_new:Npn \tl_map_function_unbraced:NN
+  { \exp_args:No \tl_map_function_unbraced:nN }
+\cs_new:Npn \@@_map_function_unbraced:Nn #1#2
+  {
+    \@@_if_recursion_tail_break:nN {#2} \tl_map_break:
+    #1 #2 \@@_map_function_unbraced:Nn #1
+  }
+\cs_generate_variant:Nn \tl_map_function_unbraced:NN { c }
 %    \end{macrocode}
 % \end{macro}
 % \end{macro}
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -573,6 +573,27 @@
 %   \texttt{n}-type arguments. See also \cs{tl_map_function_unbraced:NN}.
 % \end{function}
 %
+% \begin{function}[updated = 2020-10-30]
+%   {\tl_map_inline:Npn, \tl_map_inline:cpn}
+%   \begin{syntax}
+%     \cs{tl_map_inline:Npn} \meta{tl~var} \meta{parameters} \Arg{inline function}
+%   \end{syntax}
+%   Applies the \meta{inline function} to every \meta{item} stored within the
+%   \meta{tl~var}. The \meta{inline function} should consist of code which
+%   receives the \meta{item} unbraced as \meta{parameters} (|#1|, |#2|, \emph{etc.}).
+%   See also \cs{tl_map_function_unbraced:NN}.
+% \end{function}
+%
+% \begin{function}[updated = 2020-10-30]{\tl_map_inline:npn}
+%   \begin{syntax}
+%     \cs{tl_map_inline:npn} \Arg{token list} \meta{parameters} \Arg{inline function}
+%   \end{syntax}
+%   Applies the \meta{inline function} to every \meta{item} stored within the
+%   \meta{token list}. The \meta{inline function} should consist of code which
+%   receives the \meta{item} unbraced as \meta{parameters} (|#1|, |#2|, \emph{etc.}).
+%   See also \cs{tl_map_function_unbraced:nN}.
+% \end{function}
+%
 % \begin{function}[updated = 2012-06-29]
 %   {\tl_map_inline:Nn, \tl_map_inline:cn}
 %   \begin{syntax}
@@ -2530,6 +2551,33 @@
 \cs_generate_variant:Nn \tl_map_function_unbraced:NN { c }
 %    \end{macrocode}
 % \end{macro}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\tl_map_inline:npn, \@@_map_inline:nnn}
+% \begin{macro}{\tl_map_inline:Npn, \tl_map_inline:cpn}
+%   The inline functions are straight forward by now. We use a little
+%   trick with the counter \cs{g__kernel_prg_map_int} to make
+%   them nestable. We can also make use of \cs{@@_map_function:Nn}
+%   from before.
+%    \begin{macrocode}
+\cs_new_protected:Npn \tl_map_inline:npn #1#2#
+  { \@@_map_inline:nnn {#1} {#2} }
+\cs_new_protected:Npn \@@_map_inline:nnn #1#2#3 
+  {
+    \int_gincr:N \g__kernel_prg_map_int
+    \cs_gset_protected:cpn
+      { @@_map_ \int_use:N \g__kernel_prg_map_int :w } #2 {#3}
+    \exp_args:Nc \@@_map_function_unbraced:Nn
+      { @@_map_ \int_use:N \g__kernel_prg_map_int :w }
+      #1 \q_@@_recursion_tail
+    \prg_break_point:Nn \tl_map_break:
+      { \int_gdecr:N \g__kernel_prg_map_int }
+  }
+\cs_new_protected:Npn \tl_map_inline:Npn
+  { \exp_args:No \tl_map_inline:npn }
+\cs_generate_variant:Nn \tl_map_inline:Npn { c }
+%    \end{macrocode}
 % \end{macro}
 % \end{macro}
 %


bb010g:
(Also, would you be willing to label this as [`hacktoberfest-accepted`](https://hacktoberfest.digitalocean.com/hacktoberfest-update)? I know this is a bit rough in current state, but at least Hacktoberfest pushed me to PR this at all instead of probably continuing to let it language for more months, and I'm up for fixing this more when my brain isn't in a LaTeX3-foggy state.)

PhelypeOleinik:
First of all, thank you for your contribution! :D

What follows is my _personal_ opinion on these functions (thus others may have different views):  
They feel a bit unpredictable.  I mean that in the sense that depending on the input they may behave wildly.  For example,
```latex
\tl_map_inline:npn { a b c } #1 #2 { (#1) }
```
will get in an infinite loop.

I mean, yes, proper usage is supposed to be something like
```latex
\tl_map_inline:npn { { a b } { c d } } #1 #2 { (#1) }
```
but both are equally valid token lists in other functions, so I feel like the first case shouldn't break so badly.  They look a bit too specific to be generic, I think.

What I'd expect (I think) from an interface like this would be for an input like
```
\tl_map_inline:npn { a b { c d } e f } #1 #2 { (#1/#2) }
```
to produce `(a/b)(cd/e)(f/)`... maybe (coming up with new interfaces is hard, I know :).

Skillmon:
I have some small thing to mock: `keyval` is a prefix reserved for `l3kernel`, so if your package uses that code as is, please rename it (and register the new prefix). A list of registered prefixes can be found [here](https://github.com/latex3/latex3/blob/master/l3kernel/doc/l3prefixes.csv).

bb010g:
On the `keyval` example, it's a dumb monkey-patch thing for personal experimentation, which I know never goes on CPAN. These macros were prototyped the same way.

blefloch:
(Not tested but) It seem `\tl_map_inline:npn { { a b } { c d } } #1 #2 { (#1/#2) }` is equivalent to

     \cs_new_protected:Npn \__mypkg_tmp:nn #1#2 { (#1/#2) }
     \tl_map_tokens:nn { { a b } { c d } } { \use:nn \__mypkg_tmp:nn }

which is slightly clearer to me, but I haven't thought about it much.

bb010g:
@PhelypeOleinik

> First of all, thank you for your contribution! :D

Thank you! I'm amazed that I came up with anything that had a chance of making it into l3kernel.

> What follows is my _personal_ opinion on these functions (thus others may have different views):
> They feel a bit unpredictable. I mean that in the sense that depending on the input they may behave wildly. For example,
> 
> ```
> \tl_map_inline:npn { a b c } #1 #2 { (#1) }
> ```
> 
> will get in an infinite loop.

This is a fair concern, and something that should be accounted for. Especially with the `\tl_map_inline:Npn` & `\tl_map_inline:cpn` variants, the user should receive an error about missing parameters and/or the missing parameters should fall back to a sensible default.

> I mean, yes, proper usage is supposed to be something like
> 
> ```
> \tl_map_inline:npn { { a b } { c d } } #1 #2 { (#1) }
> ```
> 
> but both are equally valid token lists in other functions, so I feel like the first case shouldn't break so badly. They look a bit too specific to be generic, I think.

Could you explain what you mean by «They look a bit too specific to be generic» more?

> What I'd expect (I think) from an interface like this would be for an input like
> 
> ```
> \tl_map_inline:npn { a b { c d } e f } #1 #2 { (#1/#2) }
> ```
> 
> to produce `(a/b)(cd/e)(f/)`... maybe (coming up with new interfaces is hard, I know :).

Falling back to an empty token list for missing parameters makes sense. Is there a way to do that with quarks at the end of the parameter set, such that all parameters being gobbled lets the quark be removed, but attempted use makes the quark an empty group and then itself in preparation to be potentially gobbled as the next argument?

---

@blefloch

> (Not tested but) It seem `\tl_map_inline:npn { { a b } { c d } } #1 #2 { (#1/#2) }` is equivalent to
> 
> ```
>  \cs_new_protected:Npn \__mypkg_tmp:nn #1#2 { (#1/#2) }
>  \tl_map_tokens:nn { { a b } { c d } } { \use:nn \__mypkg_tmp:nn }
> ```
> 
> which is slightly clearer to me, but I haven't thought about it much.

That still only scales up to `\use:nnnn`, which is getting up there for inline mappings, but it's still a kinda arbitrary limit? It also means you have to do extra token list gymnastics for an expansion like `\tl_map_inline:npn { { a b } { c d } } #1#2 { (#2/#1) }` or `\tl_map_inline:npn { { a b } { c d } } #1#2 { (#1/#2/#1) }`.

`\tl_map_inline:nn` exists to make simple repetitions less of a pain & cleaner, but requiring a `\use:nn` series function means you're back to creating private, scratch functions. `\tl_map_inline:nn` *already* creates a (very private & isolated) scratch function, but it doesn't let you control that function's parameter list. `\tl_map_inline:npn` exposes that in a natural way.

car222222:
@bb010g wrote: That still only scales up to \use:nnnn

Why?  You can easily define the 'extra n versions' if you want?

So I think @blefloch 's idea is worth pursuing.  

At least use a name more like his suggestion, rather than 'map_inline' which for the name must run over all single items (i.e. with inlynin3 parameter (#1) being an item.  With this name it must not over consecutive pairs (with possibly empty implicit items) or anything else fancy.

blefloch:
No need to go beyond \use:nn.  What happens if the tl contains { abcdefghi } {
jklmnopqr } etc is

\use:nn \__mypkg_tmp:nnnnnnnnn { abcdefghi }

which gives

\__mypkg_tmp:nnnnnnnnn abcdefghi

as needed.

PhelypeOleinik:
> > I mean, yes, proper usage is supposed to be something like
> > ```
> > \tl_map_inline:npn { { a b } { c d } } #1 #2 { (#1) }
> > ```
> > but both are equally valid token lists in other functions, so I feel like the first case shouldn't break so badly. They look a bit too specific to be generic, I think.
> 
> Could you explain what you mean by «They look a bit too specific to be generic» more?

I mean they are `tl` functions, so in principle they should operate on arbitrary token lists, but instead they require the token lists to have a very specific structure (the case that doesn't require this special structure is when the parameter is `#1`, which is what is already in the kernel).  As you said, this should sanitise the input and warn/error the user about missing items in the token list (or at least provide fallback values, but definitely not an infinite loop), so I'd expect this level of parsing to belong to a specific package, rather than something generic in the kernel.  I hope that's clear (and that it makes sense :)

> > What I'd expect (I think) from an interface like this would be for an input like
> > ```
> > \tl_map_inline:npn { a b { c d } e f } #1 #2 { (#1/#2) }
> > ```
> > 
> > 
> > to produce `(a/b)(cd/e)(f/)`... maybe (coming up with new interfaces is hard, I know :).
> 
> Falling back to an empty token list for missing parameters makes sense. Is there a way to do that with quarks at the end of the parameter set, such that all parameters being gobbled lets the quark be removed, but attempted use makes the quark an empty group and then itself in preparation to be potentially gobbled as the next argument?

The falllback value could be anything reasonable (empty, `\q_no_value`, `\c_novalue_tl`, etc.).  Here's a possible implementation of that (didn't put too much thought in it, so it probably has room for improvement).  The idea is roughly the same as in your implementation, with a few tweaks.  It first sets the `\__tl_map_<n>:w` to run the `<inline function>` _and_ call the looping auxiliary, so that `\__tl_map_<n>:w` is responsible for executing the code and resuming the loop (it would be much more effort, otherwise, to pre-grab the tokens for the `<inline function>`).

Then the code computes `<number of parameters> - \int_mod:nn { <number of items supplied> } { <number of parameters> }`, and inserts that number of empty items at the end of the `<supplied items>`, so that the `<inline function>` never gets fewer parameters than it was expecting.  A `\q__tl_recursion_tail` is inserted at the end.

Finally, at each iteration, `\__tl_map_inline_aux:Nn` checks the first item of the list to be grabbed; if it is `\q__tl_recursion_tail`, it stops, otherwise it expands `\__tl_map_<n>:w` to do its thing.

```latex
\ExplSyntaxOn

\cs_new_protected:Npn \tl_map_inline:npn #1#2#
  { \__tl_map_inline:nnn {#1} {#2} }
\cs_new_protected:Npn \__tl_map_inline:nnn #1#2#3
  {
    \int_gincr:N \g__kernel_prg_map_int
    \cs_set_protected:Npx \__tl_tmp:w ##1 ##2
      {
        \cs_gset_protected:cpx
          { __tl_map_ \int_use:N \g__kernel_prg_map_int :w } ##1
          {
            ##2
            \exp_not:N \__tl_map_inline_aux:Nn
              \exp_not:c { __tl_map_ \int_use:N \g__kernel_prg_map_int :w }
          }
      }
    \__tl_tmp:w {#2} {#3}
    \cs_set:Npn \__tl_tmp:w ##1 ##2
      {
        \int_compare:nNnF { \int_mod:nn {##1} {##2} } = { 0 }
          { \prg_replicate:nn { ##2 - \int_mod:nn {##1} {##2} } { { } } }
      }
    \exp_args:Nx \use_ii_i:nn
      {
        \exp_args:Nff \__tl_tmp:w
          { \tl_count:n {#1} } { \tl_count:n {#2} / 2 }
        \exp_not:N \q__tl_recursion_tail
      }
      {
        \exp_args:Nc \__tl_map_inline_aux:Nn
          { __tl_map_ \int_use:N \g__kernel_prg_map_int :w } #1
      }
    \prg_break_point:Nn \tl_map_break:
      { \int_gdecr:N \g__kernel_prg_map_int }
  }
\cs_new_protected:Npn \__tl_map_inline_aux:Nn #1 #2
  {
    \__tl_if_recursion_tail_break:nN {#2} \tl_map_break:
    #1 {#2}
  }
\cs_new_protected:Npn \tl_map_inline:Npn
  { \exp_args:No \tl_map_inline:npn }
\cs_generate_variant:Nn \tl_map_inline:Npn { c }

% \tl_map_inline:npn { a b { c d } e f }
%   { \typeout{()} } % BOOM

\tl_map_inline:npn { a b { c d } e f }
  #1 { \typeout{(#1)} }

\tl_map_inline:npn { a b { c d } e f }
  #1 #2 { \typeout{(#1/#2)} }

\tl_map_inline:npn { a b { c d } e f }
  #1 #2 #3 { \typeout{(#1/#2/#3)} }

\tl_map_inline:npn { a b { c d } e f }
  #1 #2 #3 #4 { \typeout{(#1/#2/#3/#4)} }

\tl_map_inline:npn { a b { c d } e f }
  #1 #2 #3 #4 #5 { \typeout{(#1/#2/#3/#4/#5)} }

\tl_map_inline:npn { a b { c d } e f }
  #1 #2 #3 #4 #5 #6 { \typeout{(#1/#2/#3/#4/#5/#6)} }

\stop
```

But this has lots of pitfalls, for example:
```latex
\tl_map_inline:npn { a b { c d } e b f }
  #1 b #2 { \typeout{(#1/#2)} }
```
runs correctly in the code above by chance, because the code relies in counting the number of tokens in the `<parameter text>` and in the `<token list>`, so adding an explicit delimiter there will throw a monkey wrench in the counting (even spaces are dangerous there if the interface is somehow exposed to the user).  Maybe a less unstable interface would be
```latex

**Comment by car222222** [comment]
@PhelypeOleinik wrote:
I mean they are tl functions, so in principle they should operate on arbitrary token lists

But also, if they continue to have names \tl_map... , then they should do just that: map over the list, just one item at a time, not two or three or . . . , as this is what 'mapping over a list'  means.

Best to use a successor function (which every list should have!) to pick up more than one successive item.

**Comment by PhelypeOleinik** [comment]
> But also, if they continue to have names \tl_map... , then they should do just that: map over the list, just one item at a time, not two or three or . . . , as this is what 'mapping over a list' means.

Ah, that's a good point too!

> Best to use a successor function (which every list should have!) to pick up more than one successive item.

@car222222 Sorry, what's a successor function?

**Comment by car222222** [comment]
Well, what I mean (for a tl using source3 conventions) is a function:

\tl_successor:nn { 〈 token list 〉 } { 〈 integer expression 〉 }

-->

\tl_item:nn { #1 } { #2 + 1 }

Maybe \tl_nextitem:nn  would be a better name?

**Comment by car222222** [comment]
Which leads to the question of what is (should be) returned when #2 is out of range in \tl_item:nn ?

**Comment by PhelypeOleinik** [comment]
Ah, got it. Thanks!

> Which leads to the question of what is (should be) returned when #2 is out of range in \tl_item:nn ?

It expands to nothing. The documentation of `\tl_item:nn` says:

>  If the index is out of bounds, then the function expands to nothing.

**Comment by car222222** [comment]
The documentation of \tl_item:nn says:

If the index is out of bounds, then the function expands to nothing.

Yes, I found that now.  Not sure that this is optimal as an item being selected could well both exist and be empty!  It should give some kind of sentinel 'out-of-range' value for testing: otherwise it is as if every list is 'doubly infinite/unbounded' with an infinite number of empty items! 

Needs an issue or two, I guess.

**Comment by Skillmon** [comment]
Either it could provide `\q_no_value` or `\c_novalue_tl` (I'd suggest the second, as that one doesn't lead to an infinite loop), or provide empty values for those out of bounds. The issue with providing a marker is, that there is quite some overhead to test for the marker for every (except the first) value (making the inner code slow), and the issue with providing none is that one can't distinguish `{}` from the end of the list this way.

Both have some serious drawbacks, which is why I'd advocate for either don't provide it as part of `l3tl` (it is quite special code), or provide both versions.

**Comment by Skillmon** [comment]
Also, as long as the function should loop over `n`-type arguments, the testing overhead is small, but what about delimited arguments? If an `npn`-signature function is provided, the testing overhead becomes much bigger.

**Comment by eg9** [comment]
> The documentation of \tl_item:nn says:
> 
> If the index is out of bounds, then the function expands to nothing.
> 
> Yes, I found that now. Not sure that this is optimal as an item being selected could well both exist and be empty! It should give some kind of sentinel 'out-of-range' value for testing: otherwise it is as if every list is 'doubly infinite/unbounded' with an infinite number of empty items!
> 
> Needs an issue or two, I guess.

If one needs the check, it can be implemented:
```
\cs_new:Nn \tl_item_check:nn
 {
  \int_compare:nTF { \tl_count:n { #1 } >= \int_abs:n { #2 } }
   {
    \tl_item:nn { #1 } { #2 }
   }
   {
    BOOM
   }
 }
```

blefloch:
I'm closing this because it seems more like a convenience macro than an important improvement.  However, it led to interesting discussions, for which I thank you, and I've opened two issues as a consequence.

bb010g:
Thank you; I haven't been using LaTeX3 enough lately to feel qualified to further develop this.

PR #823, opened 2020-11-15
Prop to keyval

Skillmon:
This PR adds the function `\prop_to_keyval:N` which expandably turns a `prop` variable in a key=value list (so basically is the reverse of `\prop_set_from_keyval:Nn`).

This could be used to manage the options later given to a key=value interface using `l3prop`. Each value will be returned inside of braces with a space before the opening brace (so as `<key>= {<value>}`). This is done so that most key=value interfaces parse the pair correctly (as some strip braces too early). I tested this variant and it only failed for two of the 9 tested key=value packages and then only in one of the two test cases.


FrankMittelbach:
One possible problem that I see with the idea is that property lists have no order, but keyval lists often do have an order.
Of course, the same is true for the command for the opposite direction, but there is no problem if you only convert and then read from the prop but not alter it. In this direction, however,  you expect alterations done to the prop prior to being converted to a keyval list so the resulting list is essentially random, isn't it?  Don't necessarily means that this is a bad idea but it could be carefully documented as a restriction in my opinion.

Skillmon:
You're right, I thought that since `prop`s are documented to be unordered, I wouldn't have to document this (and I even talked with Ulrike beforehand about just that point). I can include a warning in the documentation.

Skillmon:
I should add that the different implementations behave differently, the one making use of `\expanded` will output the list in the current order of the `prop`, while the one without `\expanded` will invert the order. If this is unwanted I can make the second variant behave like the first (at the cost of performance, but since that variant is only there to support old engine versions which don't have `\expanded` yet, I guess that would be fine).

blefloch:
On 11/16/20 9:58 PM, Skillmon wrote:
> I should add that the different implementations behave differently, the one
> making use of |\expanded| will output the list in the current order of the
> |prop|, while the one without |\expanded| will invert the order. If this is
> unwanted I can make the second variant behave like the first (at the cost of
> performance, but since that variant is only there to support old engine versions
> which don't have |\expanded| yet, I guess that would be fine).

About reversing: check out code comments around the implementation of
\seq_reverse:N and check with >5000 item lists whether you run into an overflow.
 Maybe it is irrelevant if the code is too slow for such a use case, but maybe
not, in which case a slower approach that grabs the whole list at each step is
better (then reversing is not helpful speed-wise).

Skillmon:
@blefloch thanks for pointing out, I wasn't aware of this. I've tested the new variants for 10000 items and both seem fine now.

Skillmon:
There is, however, a small issue with the test suite. The variant without `\expanded` will not be tested at all. I *think* it behaves good, and of course I tested it, but it will not be checked whether it breaks in the future.

blefloch:
I think this could be a nice addition.  However, keys of property lists can include commas and can start/end with spaces, so presumably you need to output `{<key>}= {<value>}` at least in these cases?

blefloch:
I'll be happy to merge this if you manage to deal properly with keys that start/end with spaces or contain commas or equal signs.  Presumably bracing the key in all cases is not great, because one may expect simple cases to give `a={text a},b={text b}` and so on, or even `a=text a,b=text b`

Skillmon:
@blefloch Thanks for the prod, I totally forgot about the remaining open points. I'll first run a few tests to see how many of the key=val parsers are fine with braced key names.

About simplifying the output when bracing isn't needed: This could be done, but I don't really see the advantage. This is about programmatically turning data into a format that is understood without errors by other parsers. As long as the other parsers get the contents correct everything is fine, slowing down the processing by checking whether equals, commas, or spaces on either end are present seems wasteful to me (more wasteful than including 3 maybe unneeded tokens per entry). But I'll check how efficiently this can be done.

blefloch:
Don't waste time finding a quick way to detect space/comma/equal, unless too
many keyval parsers break with ,~{#1}=~{#2} or similar.

Skillmon:
Ok, I ran a few tests (finally). All the major key=val packages seem to cope reasonably well with braced keys (`pgfkeys` sometimes reparses a key=value pair depending on the used handler, but nothing we can do there anyway). For normal keys all tested packages parse `~{key}=~{value}` correctly (no fancy tests on the used value, just looked for key name parsing here). Also "fancy" key names containing equals or commas worked in all of the big packages, however they failed in `options` and `yax` (while `options` tried to run the correct key names, just defining didn't work; I didn't take a closer look on `yax`).

Bottom line: I'm going for the `,~{#1}=~{#2}` implementation.

blefloch:
Great, thanks! Let me know when to look at it and merge.

Skillmon:
I'm still unsure whether I should include the conditional code to simplify the output of the keys, a quick test would need two or three steps of expansion and simply work by argument gobbling logic. For the values the situation is more complicated and I'd not include such tests, here we'd have to test 4 things: single group, starts/ends with space, contains unhidden `=`, contains unhidden `,`. And the test would be slower as we have arbitrary input (so we'd have to use comparatively slow functions like `\tl_if_head_is_group:nTF`).

Without the simplification of key-name output I think this is ready.

blefloch:
Let's do the simple variant.  If people request it we can add extra tests for the keys, but you're right that testing the value is most likely a waste of time.

PR #824, opened 2020-11-24
Register prefix (prelim for prelim2e)

TeXhackse:
It's not a big package, but since this is to avoid duplicates, you get a registration. I hope you are fine with this, otherwise I can also stick to L2 Syntax for the internal macros.


TeXhackse:
I didn't expect this to break something and the build system is confusing me. So if anyone could tell me if it's my fault, that would be great. Thank you. - I will wait with the CTAN upload for a reply here.

PhelypeOleinik:
@TeXhackse No, the failure is due to font caching on Travis; not your fault :)

FrankMittelbach:
I wonder if we can (and should) mask that out in the test if that could happen more often
```
! (load cache: extra_lmromanslant10-regular.luc)

! (load cache: ../extra_lmromanslant10-regular.lua.gz)(load cache: ../extra_lmromanslant10-regular.luc)
```
What do you think @josephwright ?  Doesn't seem to have sensible pattern hough.

josephwright:
@FrankMittelbach  I did try but it's tricky, and a simple 're-run when the cache changes' approach _does_ work

PR #826, opened 2020-11-27
Newprefixes

u-fischer:


u-fischer:
I'm not sure to which "packages" the latex2e prefixes actually belong (and which date to use ...)

FrankMittelbach:
maybe we should make a global change here and change LaTeX3 Project to LaTeX Project in the file.

PhelypeOleinik:
It's missing the `hook` prefix from `lthooks`.  I think that as “package” the file name is okay (for example `lthooks` introduces itself as “The `lthooks` package”), but it might be confusing... Maybe “LaTeX2e - `lthooks`”?

FrankMittelbach:
or maybe just "LaTeX format"  for all those that are inside the format?

car222222:
@u-fischer I guess there is some reason not to use 'hyper'?
'hyp' looks a bit strange to me, but better than hype:-)!

u-fischer:
@car222222  no real reason, mostly because I thought three letters are enough. hyperref itself use `\Hy` as prefix, sometimes with some additions like `\HyPsd`, but I didn't want to use uppercase, and hy looked to much like a greeting. But perhaps it would be really better to use `hyper`.

PR #828, opened 2020-12-05
[.travis.yml] Set `os`, `dist` and remove `on_start` and `skip_cleanup`

SamuelMarks:
Used official linter from travis 1.10.0 (https://github.com/travis-ci/travis.rb#lint) to find the last two issues, `skip_cleanup: true` is now default as per https://blog.travis-ci.com/2019-08-27-deployment-tooling-dpl-v2-preview-release#cleaning-up-the-git-working-directory. `on_start` isn't present for email https://docs.travis-ci.com/user/notifications/#configuring-email-notifications. Explicitly setting the `os` and `dist` settings, as that's general good practice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+os: linux
+dist: xenial
+
 install:
  - source ./support/texlive.sh
 
@@ -25,7 +28,6 @@ notifications:
       - latex3-commits@tug.org
     on_success: change
     on_failure: always
-    on_start:   never
 
 deploy:
   - provider: releases
@@ -34,7 +36,6 @@ deploy:
     file_glob: true
     file:
       - ./*.zip
-    skip_cleanup: true
     on:
       tags: true
     prerelease: false


josephwright:
I'm wary of forcing any particular Linux distro: doesn't that just mean we have to alter the YAML from time-to-time for no good reason? In many ways the same applies to OS: we don't really care if it's Linux or Mac or Windows (WSL or whatever).

SamuelMarks:
@josephwright In that case you can test on multiple operating systems using Travis CI, like so:
https://docs.travis-ci.com/user/multi-os

Happy to send a PR for that. In terms of locking down to one OS distribution, the advantage is that your builds won't randomly fail for no apparent reason (assuming there's an issue with a new OS version). Whereas setting it explicitly ensure a certain level of stability; caeteris paribus.

josephwright:
As it's likely not a downside, I'll merge.

The point about OS selection is that for us, there really should be no effect as we are using 'vanilla' TeX Live, and it *really* shouldn't give OS-dependent results :)

josephwright:
Hmm, something's not right now: cleanup still happens, so I'm building a release locally :(

SamuelMarks:
Maybe it's a one-off? - Otherwise I suppose you can `cleanup: false`, i.e., what should be the default [now]…

josephwright:
I think I tracked down the issue: the new API is still in beta (over 1 year!). I've added the necessary opt-in, I hope.

Issue #24, opened 2011-04-29
Expandable `\clist_if_in_numrange:nn`


blefloch:
In the current `l3candiates`, there is a non-expandable `\clist_if_in_numrange:nn`, with a comment that it could be made expandable. The topic came up on LaTeX-L, so I copy the code here. I have the feeling that it may be better to use a double dash as a range delimiter: `--2,4--7,10--`. Two reasons: practical, so that negative numbers are easy to use in this context, and typographical: the correct way to write a number range (`1989--2010`) in a text uses an en-dash (I believe?).

    % \begin{macro}[pTF]{\clist_if_in_numrange:nn}
    %   Expandable version of what's now in |l3candidates.dtx|.
    %   A `numrange' is like |-2,5-8,17-| (can be unsorted).
    %
    %   The following possibilities are supported for the numrange:
    %   \begin{itemize}
    %   \item |4|, a simple integer
    %   \item |5-8|, a range of integers
    %   \item |-2| or |17-| for open-ended ranges.
    %   \item |-| for accepting everything (?).
    %   \end{itemize}
    %   Numeric expression (or negative integers) can be enclosed within braces.
    %   However, this may not work super well yet. In particular, right now, you
    %   will have to use two pairs of braces when giving a single negative number,
    %   since grabbing the element of a clist loses one pair of braces, and the
    %   minus sign would then be interpreted as a delimiter for an open-ended
    %   range.
    %
    %    \begin{macrocode}
    \prg_set_conditional:Npnn \clist_if_in_numrange:nn  #1 #2 {p,TF,T,F}
       {
          \exp_args:Nf \clist_numrange_recurse:nw
             { \int_eval:n {#1} }
             #2,\q_recursion_tail,\q_recursion_stop
       }
    \cs_new:Npn \clist_numrange_recurse:nw #1 #2,
       {
          \quark_if_recursion_tail_stop_do:nn {#2} {\prg_return_false:}
          \clist_numrange_parse:nwww {#1} #2 - #2 - \q_stop
          \clist_numrange_recurse:nw {#1}
       }
    \cs_set:Npn \clist_numrange_parse:nwww #1 #2 - #3 - #4 \q_stop
       {
           \tl_if_empty:nF {#2}{
              \int_compare:nNnT {#1}<{#2} {\use_none_delimit_by_q_stop:w}
            }
           \tl_if_empty:nF {#3}{
[...]

PR #831, opened 2020-12-16
L3bitset

u-fischer:
This defines a small set of command for bitsets (base commands and what is needed for PDF dictionaries).
A bitset is a simple string of 0 and 1. The index is zero based (as it is in bitset). It can grow and the length is not delimited, but it can be output as a decimal only if it doesn't exceed 2**32-1.


blefloch:
I think the bound should be 2^31-1 since TeX integers are signed.

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+%   { \bitset_set:Nn, \bitset_set:cn, \bitset_gset:Nn, \bitset_gset:cn  }
+%   \begin{syntax}
+%     \cs{bitset_set:Nn}   \meta{bitset var}  \Arg{integer expression}\\
+%     \cs{bitset_gset:Nn}  \meta{bitset var}  \Arg{integer expression}
+%   \end{syntax}
+% Evaluates the \meta{integer expression} and sets the bit of the
+% index position it represents to 1. Index position are zero based and set from
+% right to left (so one can view the bitset as a binary number).
+% If needed the length of the bit vector is enlarged.
+% If the \meta{integer expression} evaluates to a negative number, nothing happens.
```
zauguin, line 105: Is there a reason why we want to allow this? It seems to me that a negative index should trigger an error instead. (The same applies to `_[g]clear:`)

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+             }
+         }
+     }
+ }
+
+\cs_new_protected:Npn \bitset_gclear:Nn #1 #2  %#1 name, #2 index (zero based)
+ {
+   \int_compare:nNnT {#2 } > { -1 }
+     {
+       \int_compare:nNnT {\str_count:N #1 } > { #2 }
```
PhelypeOleinik, line 360: If `\bitset_clear:Nn` and `\bitset_set:Nn` are roughly the same function, they could be generalised to an internal `\__bitset_set:NNn` that takes `1` or `0` as its first argument.  However `\bitset_clear:Nn` doesn't seem to extend the `bitset var` in case the index `#2` is out of the range.  Is this intentional?

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+% output is needed.
+%
+% \section{Creating bitsets}
+% \begin{function}[added = 2020-12-13]
+%   {\bitset_new:N,\bitset_new:c}
+%   \begin{syntax}
+%     \cs{bitset_new:N}  \meta{bitset var} \\
+%   \end{syntax}
+% Creates a new \meta{bitset var} or raises an error if the name is already taken.
+% The declaration is global. The \meta{bitset var} is initially empty.
```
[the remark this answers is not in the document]
u-fischer: It would be okay to initialize it to zero (and I had it at the start), the main problem is that one has to check if one has to do it globally or locally by looking at the variable name, and to decide what should happen if people use a "wrong name" not starting with l or g.  So leaving it empty is easier (and should imho work fine too).

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+%   \begin{syntax}
+%     \cs{bitset_clear:Nn}   \meta{bitset var}  \Arg{integer expression}\\
+%     \cs{bitset_gclear:Nn}  \meta{bitset var}  \Arg{integer expression}
+%   \end{syntax}
+% Evaluates the \meta{integer expression} and clears the bit of the
+% index position it represents (sets it to 0).
+% The index is zero based. If the \meta{integer expression} evaluates to a
+% negative number or is larger than the current length of the bit vector
+% nothing happens. If the leading (left most) bit is cleared,
+% zeros are not trimmed but stay in the bit vector.
```
zauguin, line 119: Since we have variable-sized bitsets anyway and most of the code seems to work faster with shorter bitsets than with longer ones, I would trim them here. This would also give them a more predictable length and allow to temporarily use high bits without permanently altering the bitset.

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+% \end{function}
```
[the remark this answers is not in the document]
u-fischer: the problem with on/off is that sound quite odd in the global case: `gon/goff`. Similar with true/false (gtrue/gfalse).  An alternative could be `\bitset_put`/\bibset_pop` but  I mostly  use set/clear as it is also used in the bitset package, and set is also used in other programming languages.

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+%
+% \section{Using bitsets}
+%
+% \begin{function}[EXP,added = 2020-12-14]
+%   { \bitset_index:Nn, \bitset_index:cn }
```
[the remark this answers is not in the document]
u-fischer: wouldn't that be confusing as the index is zero based?

l3kernel/l3bitset.dtx
u-fischer: Hm, one could even consider a public \bitset_set:NNn, then one would have to find only one name ;-). It is intentional that zeros don't extent the var--imho that isn't needed--but could be easily changed.

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+% \end{function}
+% \begin{function}[EXP,added = 2020-12-13]
+%   {\bitset_to_arabic:N, \bitset_to_arabic:c}
+%   \begin{syntax}
+%     \cs{bitset_to_arabic:N} \meta{bitset var}
+%   \end{syntax}
+% This leaves the current value of the bitset expressed as
+% a decimal number in the input stream. If no bit has been set yet,
+% the output is zero. The function uses \cs{int_from_bin:n}, so it
+% can only be used with a bitset whose value is smaller than $2^{31}-1$.
```
zauguin, line 150: In PDF, we sometimes need bitsets with bit 31 set (represented as 32bit signed two-complement integer), so this restriction might be too restrictive. (Of course it could be argued that this is a special case and generic bitset functionality should only provide the unsigned case, but then we might need a separate function to get the PDF output)

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
FrankMittelbach: I think @PhelypeOleinik has a point with clear and the convention used elsewhere for it. Maybe the right way would be to follow what what the bool module does: `\bitset_gset_true:Nn` not `gtrue`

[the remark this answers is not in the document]
PhelypeOleinik: `gon`/`goff` sounds awful indeed!  `set` and `clear` are not _bad_ on themselves, but they seem (to me) to convey the wrong idea in an `expl3` module.

`put` and `pop` don't look too good either, as they seem to add and remove a bit from the variable (which they don't necessarily do).

`gtrue` and `gfalse` are too bad, I think, but there's the problem that we don't have any other `true` or `false` as actions in `expl3`.  What about `(g)set_true` and `(g)set_false`, as `bool` variables?

Also (maybe) we could have `\bitset_toggle:N` or `\bitset_switch:N` that sets to 1 if it is 0 and vice-versa

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+    \int_compare:nNnT {#2 } > { -1 }
+      {
+        \int_compare:nNnTF {\str_count:N #1 } < { #2 + 1 }
+          {
+            %extend the str
+            \int_step_inline:nn { #2 - \str_count:N #1 }
+              {
+                \str_put_left:Nn #1 { 0 }
+              }
+            \str_put_left:Nn #1 { 1 }
```
zauguin, line 288: I haven't measured this, but something like

```
\str_put_left:Nn {
  1
  \prg_replicate:nn { #2 - \str_count:N #1 }
}
```
feels like it should be faster and is IMO more readable.

l3kernel/l3bitset.dtx
FrankMittelbach: again based on general bitset usage or on the need for pdf?

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+%    \begin{macrocode}
+\cs_new:Npn \bitset_to_arabic:N #1
+  {
+    \exp_args:Ne \int_from_bin:n { #1 }
+  }
+
+\cs_new:Npn \bitset_to_bin:N #1
+  {
+    #1
+  }
```
zauguin, line 394: Especially when showing a bitset to the user, in my experience you either want a trimmed bitset or a bitset of a predefined constant width. So maybe we could add a `\bitset_to_bin:Nn` which extends the bitset to at least `#2` digits?

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,551 @@
...
+   \int_eval:n { \str_item:Nn #1 { -1 - (#2)} +0 }
+ }
+\cs_generate_variant:Nn \bitset_index:Nn {cn}
+%    \end{macrocode}
+% \end{macro}
+
+% \begin{macro}
+%   {
+%     \bitset_show:N, \bitset_show:c,
+%     \bitset_log:N,  \bitset_log:c
```
zauguin, line 416: As for `\bitset_to_bin`, I would like a fixed length (or maybe minimum length) `:Nn` version for these.

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
PhelypeOleinik: Ah, there's that... Okay, so `index` might be better

Though being zero-based is confusing by itself.  Does it have a strong reason for being so? (other than bits being _always_ zero-based everywhere :)

l3kernel/l3bitset.dtx
PhelypeOleinik: > It is intentional that zeros don't extent the var

I think either both extend, or both don't.  Mixing just makes it confusing, I think

zauguin (commented):
Some smaller suggestions. (I don't feel particularly strongly about any of these, so feel free to ignore them if you disagree.)

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
FrankMittelbach: conceptually I think `(g)set_true/false` + `g(toggle)` not a bad choice (and add the latter also to the `bool` module).

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
zauguin: > Though being zero-based is confusing by itself. Does it have a strong reason for being so? (other than bits being _always_ zero-based everywhere :)

While I never seen one-based bitsets in another context, the PDF specification uses them, so using zero-based indexing here means that in our primary usecase everything has to be shifted.

l3kernel/l3bitset.dtx
u-fischer: I thought that trimming and extending the bitset  everytime the leading bit is switched (assuming that most uses have a more less fix length) would be slower than simply leaving it as it is. But I have no strong mind about this, and I don't know which uses cases need the high bits.

l3kernel/l3bitset.dtx
u-fischer: error would be okay with me too.

l3kernel/l3bitset.dtx
u-fischer: @FrankMittelbach  as far as I got from the bitset package, most uses of bitsets handles them as "numbers" (binary or arabic), and with numbers leading zeros aren't significant.

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
u-fischer: oh yes, zero based is confusing. But as it it represents a (binary) number and the first (right) position is  for 2^0 there is imho not much one can do.

l3kernel/l3bitset.dtx
u-fischer: we will probably need a bigint module at some time to handle larger integers and then extend the functions (the bigset package uses bigintcalc).

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
u-fischer: ok, sounds like there is some consens here. ;-)  So I will change to (g)set_true/false.

l3kernel/l3bitset.dtx
u-fischer: Probably  `\str_put_left:Ne` or Nx? but beside this you are quite right, it looks more readable.

l3kernel/l3bitset.dtx
u-fischer: I have no real opinion what a user expect or need for binary representation, (and if someone actually need it) I mostly used it to check if the index were correctly set. Do you know real use cases here?

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
u-fischer: @zauguin  oh my god, I have completly pushed that out of my mind. I'm so used to the numbering in hyperref and the bitset package, that I could have sworn the numering in the pdf reference starts with zero too. But I guess we still can't make it 1-based, or can we?

car222222:
why can't you lot ever listen to me on anything?

car222222:
And read plus discuss everything I write?

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
josephwright: I think we can go with indexing from 1. It's what we do for everything else, and it's not like people are (largely) going to do `2^(\bitset_index:n {position} )`. If we go with a zero index, we are almost certain to get confused ourselves as at least the 'always 1' plan is self-consistent.

On `index` vs `item`, I thought the point is that unlike say a `prop`, you don't have 'items' here as each entry is a flag. So `index` seemed closer to the idea we are dealing with. I think we need _some_ word there, we can't just have `\bitset:Nn`.

l3kernel/l3bitset.dtx
josephwright: As this is a one-off function, we likely don't need all of bigint` to fix it: we could merge the PR (after agreement) then look to code just that conversion up 'by hand'. (I don't think there are other places `int` functions are used?)

l3kernel/l3bitset.dtx
josephwright: I think the 'defined width' part is a more general point. Should the set have a pre-defined fixed length? This would address the question about what a the 'cleared' set looks like. Might also be useful as at present one can't predict how long the set will be if you dump it as a binary, which would be 'fun' if it was going somewhere you then wanted to pull apart on a bit basis.

l3kernel/testfiles/m3bitset001.lvt
```diff
@@ -0,0 +1,241 @@
+% !Mode:: "TeX:UTF-8:Main"
+\input{regression-test}
+\RequirePackage[enable-debug,check-declarations]{expl3}
+\documentclass{article}
+
+
+
+\begin{document}
+\ExplSyntaxOn
+\START\showoutput
```
josephwright, line 10: Using `\showoutput` here is a quick way to make a test, but almost all of the `expl3` tests use the `\TEST { }` or `\TESTEXP { } { }` set up with deliberate `\<thing>_show:N` or similar. That helps with programming stuff as we don't get the engine-dependent output that using typeset output can lead to.

I think before we merge, the tests need to be tidied up to be more similar to those in the other 'variable' modules.

l3kernel/l3bitset.dtx
josephwright, line 74: As I've said in another comment, we might want to think carefully about whether bitsets should have a pre-defined length/size. That would make things like the binary representation clearer.

josephwright (changes requested):

l3kernel/l3bitset.dtx
u-fischer: well I don't think about bitset in this way. For me it is a set of "set of indexed bits". The length would be for me the highest bit set. Zeros are for me "implementation detail", and  000, 00, 0 and "empty" are the same bitset, only with a different representation.

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
FrankMittelbach: if there is not a huge disadvantage in code I would go with indexing from 1 because it is consistent  with everything else we do. as to "item" or "index" as we support named positions I think item is again more consistent.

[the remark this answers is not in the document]
FrankMittelbach: > While I never seen one-based bitsets in another context, the PDF specification uses them, so using zero-based indexing here means that in our primary usecase everything has to be shifted.

one remark on that: we do have bithandling in LaTeX already (for the float specifiers), would be interesting to make sure that - at least in theory - all that is needed there could be properly expressed in that bitset code even if we do not change to it at this stage.

blefloch:
Replying at top-level to ask a question about the name of the module: we currently have `intarray` and `fparray`, which are data structures storing a fixed-length collection of integers or floating points.  This would suggest that the natural name for a fixed-length collection of booleans should be `boolarray`.  From this point of view the most natural name for a variable-length collection of booleans (as implemented in this pull request) would be `boolseq`, and `boolprop` for a collection of booleans accessed by name.  (I don't like these names very much.)

The big drawback of `bitset` is as follows.  We may want to define `intset` at some point (some collections of page numbers for instance, or when using numeric citations, see #24 ), which would be subsets of the set of integers, whereas the current `bitset` is of course not subsets of `{0, 1}`.  I guess I'm getting closer to Chris' suggestion to change the name to `bitvec` or similar.

l3kernel/l3bitset.dtx
josephwright: I think we agreed at the team meeting that this is likely best left 'open ended'.

car222222:
Also, a string of tokens is not in any way at all anything whatsoever to do with bits.  

bitsets ate literally what it says: actual bits (possibly grouped into bytes).  They are used only in languages that give access reasonably directly to near-hardware for efficient operations on individual 'actual bits' (such as C? well, it used to?, or Assembler)

They are useless as an abstraction in other languages.

And emulating with strings of character tokens hone many bytes per bit is that?) is, well, farcical!  

Finally, there are AFAIK zero used-cases for emulation.

An array of flags should remain an array of flags: booleans, integers or whatever -- my recommendation for the pdf case is powers of two, but that is just to make the final calculation easier.  

I also think that having the suggested 32 integer constants predefined may have many uses.

This implementation of arrays of flags also extends nicely to other 'composite numbers such as Leslie's float-specs, mathocodes, mathdekcides etc etc.

So please get real and simply drop these useless strings of char tokens. 

And DO NOT ignore me so rudely, ever again!  (And freedom for \shipout, the primitive).

car222222:
And do not feel free to ignore me!  

If you disagree, start a discussion but not do not use specious arguments such as 'every one agrees with me! Nah Nah!'  

Or simply ignore the large amount of previous discussion and produce a frankly useless and dreadfully coded module like this one! 

And do not make primitives into hooks!  Ever!  Got it?

u-fischer:
@blefloch 

> From this point of view the most natural name for a variable-length collection of booleans (as implemented in this pull request) would be `boolseq`, and `boolprop` for a collection of booleans accessed by name. (I don't like these names very much.)

I don't think that something with `bool` is right. That would imho require that you can use a bit in a boolean test.  `bitseq` could work, as can `bitvec`.  Personally I have nothing against another name, I choose bitset for the first implementation as Heiko's package is named like this and some research showed that the name is not unknown in C, java, python, lua. 

>  We may want to define intset at some point 

Well if it is for things like page ranges, then doesn't intrange as suggested in the old issue too fit better anyway?

car222222:
"I don't think that something with bool is right. 

That would imho require that you can use a bit in a boolean test."

You can use these 0s and 1s in in a boolean test so that is OK.
But they are not 'bits', they are character tokens (containing many bytes).

car222222:
@u-fischer I know a lot about bitsets in C so 
I shall try again to explain what the word means there.  
They are a very, very specialized datatype that really does give quite direct access to real bits (via Assembler and really close the hardware): the 'bit' in bitset means just that, not a boolean represented by a character token!

I shall check the other languages you cited: it could be that if we actually ever need the superbly efficient access to 'real bits' as in C, then we should use lua.

But please do not pretend to be implementing anything like (or called) a bitsets using token strings (even if Java does this!:-).

Just adding this explanation again for completeness: flag sets are isomorphic to boolean arrays (however they are implemented) so this is what we should implement (without any strings of tokens).  

The only extra method we need is the conversion/coercion of 'the value of a small array of flags/booleans into a decimal number'.

car222222:
@blefloch On the name:

What we really need here is a module called flag that contains arrays of flags and the particular methods they require.

But the word 'flag' has long ago been taken for some strange(to me:-)  expandable entity with arcane usage. 

So maybe we should use boolarray (with subtitle 'really flags').

But please: no strings of character tokens.

zauguin:
> I don't think that something with `bool` is right. That would imho require that you can use a bit in a boolean test.

I think that a function to test if the value of a specific bit is 1/true/on/... would be a good addition to the code, independent from naming. Testing is a specific flag is set sounds like a typical usecase.

@car222222 I get that you don't like the name "bitset" when the implementation is not using native bitsets, but what are your arguments against "strings of character tokens"? Representing an array of booleans as a string of tokens, each representing one boolean seems like a pretty reasonable design choice, especially since we need to support at least 32 flags so storing them in integers isn't really an option in TeX.

car222222:
@zauguin Yes sure!

what are your arguments against "strings of character tokens"? 

Probably just that they are the antithesis of the major use-case for bitsets:-).   A character string (one byte per O or 1) would not be too bad, but chracter tokens can be much bigger than that! Also, bytes are really integers so why not use integers.

Representing an array of booleans as a string of tokens, each representing one boolean seems like a pretty reasonable design choice, especially since 

"we need to support at least 32 flags"
At most, I think?  Because we want to convert the whole array to a single integer.

"so storing them in integers isn't really an option in TeX."

I am not clear what the problem is here, but maybe 
I am misunderstanding what you are saying?

My understanding is that integers (in count registers) are a very good way of storing information in TeX, much better than token lists.  Are not integer registers are pretty much just 4 bytes each? 32 times 4 is not large.  And this representation is very time-efficient (this is a general point about using integers: in any language, even TeX, they are highly optimised).

Even if we do ever need larger arrays, using a sequence (array) of arrays of integer registers sounds very efficient to me (in any language that does not give access to bits rather than bytes).

But maybe I am deluded about how TeX works with tokens and integer registers.  

That is why I had asked questions about how best to implement the small collection of powers of 2 that are needed.  One possibility would be as the logs (0 to 31) but I think that there is no difference between storing these and storing the actual powers (2^0 to 2^31): both need 32*4 bytes.

A large time advantage of storing these powers (as global integer constants) is that they are used a lot and save 32 multiplication operations on every use for for arrays.

All this is unconfirmed, so please help me to get it right!

car222222:
Thinking aloud (or allowed?):

The general elephant lurking here is how best, in general, to implement 'small arrays' in TeX. 

Maybe emulation of  something more like lua tables perhaps offers a workable solution to this?

The idea is that a TeX lua-table would be simply a 'collection of names' that point to csnames whose (one-level) expansion contains the value of that table entry.  This immediately gives named-index arrays, in which the 'index' (as an ordinal number) can either be part of the value of its entry or (part of) the name.  To get more natural looking 'classically indexed arrays', just use 'index1' etc as the names.

The values need not be explicit: for example, integer values could be more efficiently stored in registers.

This data-type is similar to property lists except that, for efficient access, all that is stored is the collection of control sequences (containing the key as part of their name) and (of course:-) the current value of each (why do attributes and their values keep turning up like this!:-).  They should probably be kept in some kind of list so that some operations involving the index order are very efficient: just execute the list!  

But (importantly for arrays in general) note that this order is never used to access an element of the array as this is very inefficient: it is an important to the essence if an array that such access to a single element does not depend on the ordering (this is often expressed by saying that access time is O(1) ).

car222222:
Real bitsets and a few of the necessary methods (far beyond what I had imagined):

Turn "11011000111101..." into bytes, (padded left or right, 0 or 1,) and vice versa.
Slice ranges of bits.
Rotate bits, addressed by the bit. That is, say: "rotate bits 13-17, wrapping around the edges," or, "rotate bits 13-17, lose bits on the one side, set all new bits to 0."
Similarly, revert regions of bits, apply logic to regions of bits, etc.,.
Switch Endianness, with different block sizes.
Apply operations in block groupings: ex: apply XOR 10101 (5 bits) repeatedly across a field.

(OK, so this is really byte-and-bitsets, rather than pure bits.)

car222222:
A very little research on bitsets in Java, Python and lua.

 All of these three languages/systems are very close to C, but in subtly different ways.  Thus in all cases, any low-level stuff is ‘essentially using  C-style access to Assembler code'.

One difference is that Java (at least some versions) provide a bitset as a fundamental data-type, whereas in Python and in lua (by definition:-) there is nothing in the base language itself but there are libraries(?) etc that add such things.: I have not looked at how they access bits and bytes (very likely it is essentially via C?).

One important thing to note (and this is made very clear in modern documentation of C and C++) is that a bitset is not a data abstraction but more a 'data concretization', allowing slightly higher level access down close to the metal/silicon/whatever.  They are used for efficiency of storage rather than ease of programming.

——

﻿A little bit about bitsets in Java (as I expected, these are low-level bits, stored eight to a byte):

Array of Bits

To store and manipulate arrays of bits, one might argue that we should use boolean[] as our data structure. At first glance, that might seem a reasonable suggestion.

However, each boolean member in a boolean[] usually consumes one byte instead of just one bit. So when we have tight memory requirements, or we're just aiming for a reduced memory footprint, boolean[] are far from being ideal.

. . .

For the sake of simplicity, let's suppose we're going to represent eight flags with one byte.

josephwright:
@car222222 

> "we need to support at least 32 flags"
> At most, I think? Because we want to convert the whole array to a single integer.
> 
> "so storing them in integers isn't really an option in TeX."
> 
> I am not clear what the problem is here, but maybe
> I am misunderstanding what you are saying?

The point is that there are use cases where we need to go beyond 31 entries in a set/array of bits. Using a macro to implement the data storage, that's not an issue as the internal representation is not limited in size. The only issue comes in conversion to an decimal representation at 'point of output'. That can be handled doing things 'by hand' in a 'big integer' implementation.

> My understanding is that integers (in count registers) are a very good way of storing information in TeX, much better than token lists. Are not integer registers are pretty much just 4 bytes each? 32 times 4 is not large. And this representation is very time-efficient (this is a general point about using integers: in any language, even TeX, they are highly optimised).

That's efficient in terms of memory usage and 'fundamental' access times, but not when you start thinking about how many expansions are needed to get data in and out. The use cases we have for bitsets involved setting various entries, possibly doing some Boolean logic with them, then finally 'passing on' to some place that needs a decimal representation. That last step is the only one where using a decimal representation helps us. So as we are implementing in TeX, with no 'native' access to binary data, we will spend a lot more time converting to/from the decimal to a binary-like form, compared to how much time we spend actually 'doing stuff'. Memory use is also not really a big issue: with the current implementation, we are using one csname per set, whereas with a count-register approach we might end up needing more than one. We essentially never see an issue with the number of tokens we store in macros in a modern TeX system. (Have you seen how much data I have to load for Unicode case changing!)

> Even if we do ever need larger arrays, using a sequence (array) of arrays of integer registers sounds very efficient to me (in any language that does not give access to bits rather than bytes).
> 
> But maybe I am deluded about how TeX works with tokens and integer registers.
> 
> That is why I had asked questions about how best to implement the small collection of powers of 2 that are needed. One possibility would be as the logs (0 to 31) but I think that there is no difference between storing these and storing the actual powers (2^0 to 2^31): both need 32*4 bytes.

The ways I could think of implementing:

- Using one macro per set, with each bit as a `0` or `1` or some other binary-like datapoint
- Using one macro per set, with the value as a decimal number
- Using one or more count registers per set: we'd need a way to allow dynamic growth, so something like
   a 'main' name plus a dynamic lookup for `\<name>_ext_1_int` where this holds bits 32-62, `\<name>_ext_2_int` for
   63-93, etc.
- Doing a similar thing to the previous entry but using chardef/mathchardef storage
- Using one macro, count or (math)chardef per bit
- Using one fontdimen per set
- Using one fontdimen plus one macro, the latter tracking which entries are in which set

The first approach keeps the data structure simple (_cf._ count/chardef approaches), allows global and local data (_cf._ fontdimens), and means we are only converting to/from decimals when we need to.

josephwright:
> Thinking aloud (or allowed?):
> 
> The general elephant lurking here is how best, in general, to implement 'small arrays' in TeX.
> 
> Maybe emulation of something more like lua tables perhaps offers a workable solution to this?
> 
> The idea is that a TeX lua-table would be simply a 'collection of names' that point to csnames whose (one-level) expansion contains the value of that table entry. This immediately gives named-index arrays, in which the 'index' (as an ordinal number) can either be part of the value of its entry or (part of) the name. To get more natural looking 'classically indexed arrays', just use 'index1' etc as the names.
> 
> The values need not be explicit: for example, integer values could be more efficiently stored in registers.
> 
> This data-type is similar to property lists except that, for efficient access, all that is stored is the collection of control sequences (containing the key as part of their name) and (of course:-) the current value of each (why do attributes and their values keep turning up like this!:-). They should probably be kept in some kind of list so that some operations involving the index order are very efficient: just execute the list!
> 
> But (importantly for arrays in general) note that this order is never used to access an element of the array as this is very inefficient: it is an important to the essence if an array that such access to a single element does not depend on the ordering (this is often expressed by saying that access time is O(1) ).

Isn't this all the fact that if you want fast random access it's always best to use an array using the hash table? We have `prop` and `seq` for places where we can save csnames and/or want convenient copying/mapping. (See _e.g._ https://tex.stackexchange.com/questions/147966/how-to-implement-low-level-arrays-in-tex which covers how naming in such cases affects hash table performance.)

josephwright:
@car222222 

> One important thing to note (and this is made very clear in modern documentation of C and C++) is that a bitset is not a data abstraction but more a 'data concretization', allowing slightly higher level access down close to the metal/silicon/whatever. They are used for efficiency of storage rather than ease of programming.

All true for C but we are in a different position. There are places, most notable PDF data structures, where we have to 'provide' an integer representation that represents a series of booleans on a bit basis. Thus a bitset for us is not primarily about efficiency or easy of storage but about necessity.

> Array of Bits
> 
> To store and manipulate arrays of bits, one might argue that we should use boolean[] as our data structure. At first glance, that might seem a reasonable suggestion.
> 
> However, each boolean member in a boolean[] usually consumes one byte instead of just one bit. So when we have tight memory requirements, or we're just aiming for a reduced memory footprint, boolean[] are far from being ideal.

As I've said in another comment, for us, the worry is typically csname usage. The difference between one register or one chardef or one macro really is neither here nor there.

l3kernel/testfiles/m3bitset001.tlg
```diff
@@ -1,2605 +1,194 @@
...
+============================================================
+============================================================
+TEST 15: show index 8 global
+============================================================
+0
+============================================================
+============================================================
+TEST 16: add index 8 with sum
+============================================================
+! Arithmetic overflow.
```
josephwright, line 122: Are these overflows deliberate?

l3kernel/testfiles/m3bitset001.tlg
```diff
@@ -1,2605 +1,194 @@
...
+I can't evaluate this expression,
+since the result is out of range.
+10000000000000000000000000000000111010101
+469
+1
+============================================================
+============================================================
+TEST 17: add index 8 with sum global
+============================================================
+! Arithmetic overflow.
```
josephwright, line 134: Same here: I'm not sure if this is right.

josephwright:
> Similarly, revert regions of bits, apply logic to regions of bits, etc.,.

These are implemented by Heiko's code, although I'm not sure anyone has ever used them 'in anger'. More widely, I think we want to start with only such operations as we need or which fit into the idea of creating a data type/module. (So for example we want `\bitset_set_eq:NN` and so on as that's 'what we do' for a data type.)

josephwright:
> Real bitsets and a few of the necessary methods (far beyond what I had imagined):
> 
> Turn "11011000111101..." into bytes, (padded left or right, 0 or 1,) and vice versa.
> Slice ranges of bits.
> Rotate bits, addressed by the bit. That is, say: "rotate bits 13-17, wrapping around the edges," or, "rotate bits 13-17, lose bits on the one side, set all new bits to 0."
> Similarly, revert regions of bits, apply logic to regions of bits, etc.,.
> Switch Endianness, with different block sizes.
> Apply operations in block groupings: ex: apply XOR 10101 (5 bits) repeatedly across a field.
> 
> (OK, so this is really byte-and-bitsets, rather than pure bits.)

These are all definitely 'bit' not 'bool' operations, BTW.

l3kernel/testfiles/m3bitset001.tlg
u-fischer: yes, at least I didn't overlooked them. It happens as `\bitset_to_arabic:N` can't handle larger bitset (yet). As this would change if the internal implemention is switched to something else, I thought having a test which would react to such a change would be sensible.

car222222:
@josephwright A very good point.  There is a recent tendency to advertise bitsets as 'just for efficiency of storage' rather than more essential features such as bit-fiddling or even set/logical operations.  So it is good to see these more eigen uses being mentioned.

u-fischer:
> "I don't think that something with bool is right.
> 
> That would imho require that you can use a bit in a boolean test."
> 
> You can use these 0s and 1s in in a boolean test so that is OK.

What I meant is that if we name them "bool" then something like 

   ~~~~
  \bool_if:nTF { \bitset_index:Nn\l_my_bitset {3} }{true}{false}
  ~~~~

should work, but it doesn't (The output is `truefalse`). In some places or combinations it may work, but this it then by accident and not by design. 

> But they are not 'bits', they are character tokens (containing many bytes).

What matters it not the internal implementation (which could even change in future), but the documented interface to the user.

car222222:
@josephwright No! "There are places, most notable PDF data structures, where we have to 'provide' an integer representation that represents a series of booleans on a bit basis."

PDF always uses only the integer (at most 2^31-1], so we do not need any bits at all: any representation of the array of booleans/flags is fine.  

This is why I, some time ago, suggested representing flag at index ~n by the boolean with values 0 and 2^n (or rather those integer constants, n=0,...,31).  Then calculating the integer needed by the PDF (the current value of the array) is a simple sum over the values of the flags.  
The bit values and their significance are both neatly hidden away in this very slick representation.  
Testing is integer comparison, either /=0 or >0.
Setting needs to use the index, but must be available to access which flag us being set so there is no overhead.
But note: only works for small arrays!  Where 32 is very small.

Repeated for emphasis: This is all that is needed for the PDF.

car222222:
@josephwright I only mentioned and investigated other languages because they were used as explanation for continuing to refer flag sets/arrays as bitsets.

car222222:
@josephwright "Isn't this all the fact that if you want fast random access it's always best to use an array using the hash table?"

Is this a fact about TeX, or about the (real) world?
I am not at all sure about hashing and use of names in TeX.

Outside TeX, direct (even fixed-size) arrays are of course superior to hash tables, but TeX does not provide them so that is not an option.  I am unsure if lua does provide such simple, direct arrays; I never investigated whether lua tables are really just arrays of pointers with extra bells and whistles (whose inefficiencies are avoided by correct usage).

For space efficiency, hashing is not an obvious choice but many things are not actually much worse if optimised hashing is used.
Ironic fact: Knuth wrote the definitive work on 'optimal hashing' (it is random) but did not think that csname storage needed it:-).

u-fischer:
> That is why I had asked questions about how best to implement the small collection of powers of 2 that are needed.

Well Joseph has already quite nicely summed up the options. But nevertheless: I know that there are various options here, a number of them came up every time I asked about bitset during the year. When I decided to implement something so that I can finish the pdfresource code (whose release we announced "in a few weeks") I used the following list of requirements:
* length longer than 32 (as already mentioned already in pdf we could need this)
* should provide local and global variants
* dynamically growing would be nice but not required
* should cover now the functions I need for pdf (that arabic works only for length <31 is ok for now)
* should have the potential to cover all functions currently offered by the bitset package as probably nothing more is needed in TeX (at least I never saw anyone asking for more)
* should not waste too many csnames or registers
* should not waste too much of my time to program it, so use existing functions if possible. 

The first point excluded using one single integer, the second excluded intarray. Using a bunch of integers or booleans would have meant to use many csnames and would have needed quite some time to write the code to chain them together. But `\str` variable had everything: It uses only one csname, the range-functions are all there, its length is not limited, and as the `bitset` package uses a simple csname too so I knew it will work with the other functions too (Heiko builds the string from left to right, while I decided to do it from right to left to have directly a binary representation and avoid the use of a reverse functions everywhere).

josephwright:
> @josephwright No! "There are places, most notable PDF data structures, where we have to 'provide' an integer representation that represents a series of booleans on a bit basis."
> 
> PDF always uses only the integer (at most 2^31-1], so we do not need any bits at all: any representation of the array of booleans/flags is fine.

The integer we have to output _is_ set up on a bit base is: that's rather the point. Now, yes, we can do what we like to get there, but I don't see the problem with having an internal form where one can see which flags are on and which are off.

> This is why I, some time ago, suggested representing flag at index ~n by the boolean with values 0 and 2^n (or rather those integer constants, n=0,...,31). Then calculating the integer needed by the PDF (the current value of the array) is a simple sum over the values of the flags.

I'm not following here. A Boolean has only two values, TRUE and FALSE, so I don't get 'boolean with values 0 and 2^n'. As you suggesting we have some `<name>_bit_<n>` chardef or similar for holding each entry of the `<name>` set/array? I'm trying to understand how you feel this can be implemented, but I think I could do with a code example [in primitives would be fine :)].

> The bit values and their significance are both neatly hidden away in this very slick representation.
> Testing is integer comparison, either /=0 or >0.
> Setting needs to use the index, but must be available to access which flag us being set so there is no overhead.

Again, I'm obviously missing something here: I don't follow this at all.

car222222:
@josephwright "As I've said in another comment, for us, the worry is typically csname usage." 

I though we did not worry much about the number of csnames any longer.

So would this be a problem? : 32 csnames (for the powers of 2) plus n for every flag array of size n (n < 33).

josephwright:
> @josephwright "Isn't this all the fact that if you want fast random access it's always best to use an array using the hash table?"
> 
> Is this a fact about TeX, or about the (real) world?

About TeX: I don't have a computer science background, so I tend to worry about how things work in TeX, or sometimes how they work in Lua, but not much else.

> I am not at all sure about hashing and use of names in TeX.

The TeX-sx question I linked to covers a good deal of the detail.

> Outside TeX, direct (even fixed-size) arrays are of course superior to hash tables, but TeX does not provide them so that is not an option. I am unsure if lua does provide such simple, direct arrays; I never investigated whether lua tables are really just arrays of pointers with extra bells and whistles (whose inefficiencies are avoided by correct usage).

Lua tables are made up of an array part and a hash part. See for example https://stackoverflow.com/questions/29928379/how-are-lua-tables-handled-in-memory for discussion.

josephwright:
> @josephwright "As I've said in another comment, for us, the worry is typically csname usage."
> 
> I though we did not worry much about the number of csnames any longer.
> 
> So would this be a problem? : 32 csnames (for the powers of 2) plus n for every flag array of size n (n < 33).

I think the key is 'too much': there's a trade-off and we tend to look at the balance between speed and csname usage. For example, in working (again) on `siunitx` v3, I decided to stick to a `prop` (one csname) for decomposing units, as I don't _need_ maximum speed, and would rather keep down csname usage.

There's also the business that operations such as copying are easier with a single csname. That would of course be hidden from end users, but if we use multiple csnames we have to provide ways to do that easily.

car222222:
@josephwright "The point is that there are use cases where we need to go beyond 31 entries in a set/array of bits."

Are there?  Are you sure?  
There definitely is no such need in the flag set use case.

If we really need large (or even medium-sized) arrays of booleans (or anything else) then, yes, different methods are needed (probably adding proper arrays to the engines is then the optimal strategy).

car222222:
@josephwright Yes, the tex.sx ref contained far too much detail for me (right now, at least).

But I think that the efficiency considerations there only apply to data sets of size much, much larger than 32.

So the fundamental question is whether, right now, we are looking at 'efficient array-like structures' for 'big data' or something neat and tiny to get sets of at most 32 flags working well.

car222222:
@josephwright I need to think further about the 'copying efficiency/ease of coding'.  I cannot see it being any problem for my 'small arrays'.  

Large data (arrays) are completely different but I am again asking whether we have any use-cases for implementing these within TeX.  (Horses for courses and all that.)

davidcarlisle:
@car222222 

> So would this be a problem? : 32 csnames (for the powers of 2) plus n for every flag array of size n (n < 33).

if we don't store each set as a single integer it's going to be a list of tokens.

You are suggesting true,false,true,true be stored as

```
\def\myset{\one\zero\four\eight}
```

If I understand you right

whereas this PR is suggesting storing it as

```
\def\myset{1011}
```

There isn't really a lot of difference between the two (not compared to the length of this thread:-)  your suggestion makes it easy to get the integer representation (in the <32 case) as you just sum the values, but most other operations are probably slightly harder as the representation of each bit is different.

car222222:
Lua tables are made up of an array part and a hash part. 
Aha, says it all! . . . except what a 'part of a lua table is';
I think that this might be what I was trying to describe at some stage.  
So I shall look here for further enlightenment:
https://stackoverflow.com/questions/29928379/how-are-lua-tables-handled-in-memory

u-fischer:
> @josephwright "The point is that there are use cases where we need to go beyond 31 entries in a set/array of bits."
> 
> Are there? Are you sure?

From the pdf reference:

> The value of the annotation dictionary’s F entry is an unsigned 32-bit integer containing flags specifying various characteristics of the annotation. Bit positions within the flag word are numbered from 1 (low-order) to 32 (high-order).
> 
> The value of the interactive form dictionary’s SigFlags entry is an unsigned 32-bit integer containing flags ...
> 
> The value of the field dictionary’s Ff entry is an unsigned 32-bit integer containing flags  ...

Currently none of these entries actually use the high bit, which is the reason that it is okay, that the to_arabic-function doesn't handle it yet, but I don't want to setup a module whose need is mostly triggered by the pdf usage in a way that it will fail if some pdf version starts to use the high bit.

car222222:
@josephwright "but I don't see the problem with having an internal form where one can see which flags are on and which are off."

This is a completely new idea, the eyeball interface!

Why would anyone want to do that?  In all but the tiniest cases knowing which bits are on/off is not much use.  Maybe you can remember the meanings of about 4 bits, but beyond that it is impossible.  That is why Ulrike supports naming the individual flags.

A string of 32 0s and 1s is virtually useless as a visual representation of anything.  Unless of course you can do binary arithmetic and other such games in your eyes/brain!  Even if you group them into 4s then 8s, most of us will still have great difficulty telling whether bits 5, 23 and 13 all have the same value.

car222222:
@josephwright "length longer than 32 (as already mentioned already in pdf we could need this)"

As already established, this is not the case.  I do not think that pdf has any other uses of flag sets (or other sets of booleans).  Thus pdf needs only max size 32 (many cases are much smaller than this, often 2, with one value fixed:-); in pdf flag sets are always represented by decimal numbers and in pdf these cannot be big enough for larger arrays.

josephwright:
> @josephwright "length longer than 32 (as already mentioned already in pdf we could need this)"
> 
> As already established, this is not the case. I do not think that pdf has any other uses of flag sets (or other sets of booleans). Thus pdf needs only max size 32 (many cases are much smaller than this, often 2, with one value fixed:-); in pdf flag sets are always represented by decimal numbers and in pdf these cannot be big enough for larger arrays.

Sure, but it's still bigger-than-31, which rules out a 'simple' approach using just integer constants or similar.At which point, having a limit of 32 seems a bit ... odd.

car222222:
@josephwright :

A Boolean has only two values, 
True!

TRUE and FALSE
False: any two values (or ranges) can be used.
Common examples are:  
   0  1
   0  nonzero
   0  positive 
   non-positive positive
   empty non-empty 
I could go on, since any two distinguishable sets will do fine.

car222222:
Sorry, I get confused: for these purposes 
   32 = 31  (approx= log of maxint).

In other words, the restriction on the size of the array is exactly the same as the restriction on the 'size of an integer', not even off-by-one.  There are 32 integers in the 
interval [0, 31].

Yes, there is the statement that "The value of the P entry is an unsigned 32-bit integer" but that is identical (in the bit world) to a "31-bit signed integer".  You just need to add a - sign to the TeX-side value to represent the largest numbers.

Also, I think (not checked) that in all the (very few) cases where pdf appears to use all 32 flags, the first (most significant) bit is mandated to always be 0, thus those pesky negative numbers will never need to be used

u-fischer:
> Yes, there is the statement that "The value of the P entry is an unsigned 32-bit integer" but that is identical (in the bit world) to a "31-bit signed integer". You just need to add a - sign to the TeX-side value to represent the largest numbers.

Sure, but I prefer to avoid special cases if possible. I don't want to add a "if index=32 then" switch everywhere unless there is a very good reason. So if you think it should be implemented this way, please offer a concrete implementation of your suggestion and some comparision tests that demonstrates why you think it is preferable, e.g. some benchmarks.

zauguin:
> Yes, there is the statement that "The value of the P entry is an unsigned 32-bit integer" but that is identical (in the bit world) to a "31-bit signed integer". You just need to add a - sign to the TeX-side value to represent the largest numbers.

The problem here is that TeX does not really support the value which has only the most significant bit set (aka. 0x80000000 or -0x80000000 or -2^32). In any situation where TeX is actually checking the value this one will trigger an overflow. (E.g. `\showthe\numexpr -"80000000\relax` overflows but even a workaround like `\showthe\numexpr -"7FFFFFFF-1\relax` triggers the same overflow.) While we could try to work around this using `\advance` which for some reason doesn't do overflow checking, this would be non-expandable and be incompatible with expl3 which consistently uses the decimal representation of integers.

zauguin:
> Also, I think (not checked) that in all the (very few) cases where pdf appears to use all 32 flags, the first (most significant) bit is mandated to always be 0, thus those pesky negative numbers will never need to be used.

No, there is at least one case (some permission flags which default to one) where the most significant bit is mandated to always be 1, so in this case negative numbers will always be used.

car222222:
@u-fischer But in practice the large numbers never occur, the most significant bit is always fixed at 0, is it not? 
So in reality, no flag ever gets set to a value greater than 2^31-1 = 01...1 .  I should look at how your code deals with these values.  

I think that therefore in my system I just need to make the constant \g_tp_thirtyone_c have the value 0, but if that is cheating then it simply gets the 'correct' TeX value and everything works.  Here 'correct' may depend on how TeX represents negative numbers but I think it is -2147483648 ??? 

Or, more likely, I do not understand signed integers and 2's-complement.

zauguin:
@car222222 As I wrote in my previous two comments, it does occur in practice and assigning the "correct" TeX value (which would in fact be -2147483648) does not work because this value is outside of the valid range of TeX registers (to quote the TeXbook: "each capable of containing integers between −2147483647 and +2147483647, inclusive;").

car222222:
@zauguin "it does occur in practice".  I shall check again, but it does not seem to occur in practice, only in theory. 

At least I got the theoretical number correct:-).  So there is just this one binary string/number that can be used in pdf (unsigned integers) but not in TeX ('almost' signed integers)!  

Not nice at all: added to my list of Knuth's Nasties.
Do you know of any other languages with a signed integer data-type that misses out this very useful number?

I must check if he has some sneaky efficiency that uses this 'missing value' for an integer register: maybe Frank knows about such things?

zauguin:
> @zauguin "it does occur in practice". I shall check again, but it does not seem to occur in practice, only in theory.

Look at the "User access permissions" (in PDF 1.7 that's section 7.6.3.2). There undefined bits (like bit 32) have to be set to 1.

> 
> At least I got the theoretical number correct:-). So there is just this one binary string/number that can be used in pdf (unsigned integers) but not in TeX ('almost' signed integers)!
> 
> Not nice at all: added to my list of Knuth's Nasties.
> Do you know of any other languages with a signed integer data-type that misses out this very useful number?
> 
> I must check if he has some sneaky efficiency that uses this 'missing value' for an integer register: maybe Frank knows about such things?

It's not an efficiency question it's a compatibility one: This range allows TeX to work correctly on systems which do not use twos complement for signed integers. Architectures with signed magnitude or ones-complement allow a signed zero instead of supporting the one additional negative value.
Regarding languages: For example C does not require -2147483648 to be supported for 32bit values (for the same reason). I would expect most other languages to have similar rules.

car222222:
@zauguin Yes, that makes sense.  But it is still a very nasty hole in TeX, at least for combinatorial applications. 
In the case of C there is in practice likely to be a type around that gives you all 2^n values.

RSN I shall check whether size 32 arrays of flags are actually needed in practice in pdf.

car222222:
"Look at the "User access permissions" (in PDF 1.7 that's section 7.6.3.2). There undefined bits (like bit 32) have to be set to 1."

I was hoping you would not know about that one:-).  It is most likely not a case that we shall use, so maybe we can get away with size 31 arrays?? Or use some special trick if we ever need just that last flag up, with all others down.  

in fact it is OK even in this case because it is not permitted to have only bit 32 being 1, others must also be 1, so an negative TeX number that actually exists will result.  But of course, the calculation of that number would need to be hacked, eg by adding -2^30 twice.

car222222:
More on those access permissions: this would be best handled by an flag array of size 12, then add the (negative) constant for all those 20 'high 1s' (are these like high-fives?) to get the 'pdf number'.

josephwright:
> I was hoping you would not know about that one:-). It is most likely not a case that we shall use, so maybe we can get away with size 31 arrays?? Or use some special trick if we ever need just that last flag up, with all others down.

But we don't need to 'get away' with anything. The existing implementation will work for storing the data, and we only then need to do the 'big int' conversion in one place.

davidcarlisle:
This has been a useful (if long) discussion but I agree with @josephwright  here. I think of the possible implementation strategies discussed, the version in the PR with a 0-1 token list offers the simplest tex implementation and the easiest extension to 32 or more flags.  Storing as an integer value or (as suggested by Chris) storing as a list of power-of-two tokens require slightly more complicated access to individual flags and more complicated workarounds around 32 bits. We also have no actual implementation suggestions for the alternative strategies.

As this is needed for the PDF uses, I would suggest that we go with this implementation now, the underlying implementation could be changed later if there turns out to be issues with this.

car222222:
Searching for "32-bit" found some other interesting arithmetic exercises, such as under

"Entries in a Mac OS file information dictionary"

But back to the flag arrays:

The largest of these are:
Field flags 26
Font flags  19

All others have size under 16.
Note: this may have changed in 2.0.

So the very special case of the User Access Permissions is the only one that is, due to its different conventions, maxed out by all those high 1s.  In fact this use of a 32-bit integer is not described as anything to do with flags.  So we can just ignore it completely.

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,586 @@
...
+%   {\bitset_to_bin:N, \bitset_to_bin:c}
+%   \begin{syntax}
+%     \cs{bitset_to_bin:N} \meta{bitset var}
+%   \end{syntax}
+% This leaves the current value of the bitset expressed as
+% a binary (string) number in the input stream.
+% If no bit has been set yet, the output is zero.
+% \end{function}
+% \begin{function}[EXP,added = 2020-12-13]
+%   {\bitset_to_arabic:N, \bitset_to_arabic:c}
```
josephwright, line 142: I was wondering about the name here. I guess we can't use `int` as that is explicitly limited in size. I thought about `bigint`, or `dec`, but I guess they are no clearer. Not asking for a change, just want to make sure the question is asked.

l3kernel/l3bitset.dtx
```diff
@@ -0,0 +1,586 @@
...
+%   }
+%   The naming of the commands follow the names in the \texttt{int} module.
+%   \cs{bitset_to_arabic:N} uses \cs{int_from_bin:n} and so can only handle small
+%   numbers. There is no test for the string length, so that long bitset with
+%   many leading zeros work nevertheless.
+%    \begin{macrocode}
+\cs_new:Npn \bitset_to_arabic:N #1
+  {
+    \exp_args:Ne \int_from_bin:n { #1 }
+  }
```
josephwright, line 402: Should we have a 'size' check here, with at present an error of the form 'Please discuss this with the team' or similar. We likely need to integrate `bigint` or at least a subset; perhaps Bruno can comment on that aspect. (I think division was the outstanding function.)

l3kernel/l3bitset.dtx
blefloch: In `l3fp` I made `\fp_to_int:n { 1e20 }` give `100000000000000000000`, so I think `\bitset_to_int:N` would be a better name.

l3kernel/l3bitset.dtx
blefloch: It's been a while.  Yes, division was still outstanding, but more importantly the code could be drastically simplified now that we assume `\expanded` is available.  I had some pretty elaborate but now useless code to convert big integers to some kind of tree to speed up all the operations, but it makes the logic extremely complicated, which stopped me from implementing division.

car222222:
I surveyed document languages and processing software and there are no cases where integers are larger than 8-bit -- sometimes signed, sometimes unsigned (with or without the missing number).

So please do not add these to latex without a lot of further thought.  

The case of the missing number (-2^31) is also extremely unlikely to be a problem.  See upcoming email.

It is possible that high-resolution graphics may need large numbers, but even then it is not clear that they should ever be processed within latex.  Note also that these use-cases typically use/need 'arbitrary precision arithmetic' rather than 'big but still fixed-size integers'. 

Please, please show some restraint.  

Note also that there is not much relationship between arithmetic and sets/arrays of flags/booleans or anything else (unless the arrays have size larger than 2^31 of course:-).

car222222:
@davidcarlisle @zauguin 

Many thanks for your valuable help to me in getting to,grips with this stuff, esp3ciallybon the arithmetic aspects.

I am happy with David's suggestions, provided that they are, as he indicated, 'just to something working somehow for now'.

But what is needed now for at most 26 flags is, well, very little. 

Best to continue discussion of the details by email or an
issue elsewhere.  It may also be wise to develop elsewhere the minimal code that is needed for pdf flags.

Then there us the name of the concept! See elsewhere.

car222222:
@car222222 wrote: "More on those access permissions: ...".

No responses!  Why?

Are we simply not covering that case? I suggested that it will probably never be needed.  It is not, in pdf terms, a flag collection, just the generic notation for permissions stored in the canonical 'pdf way'.

car222222:
@u-fischer wrote:  "... some pdf version starts to use 
the high bit."

So 2.0 does not, I guess? 

I boldly predict that this will never happen for pdf flag sets, simply because those in charge know full well that it will cause problems for both producer and consumer apps, like TeX and a lot of other older software, that use (or can be used on systems) 'deficient signed integers' 

I am also reasonably sure that we can get the pdf engineers and standards people (the latter actually being Matthew:-) to guarantee that this will never happen, and probably even to make an addendum saying something like 'It will always be safe to use only signed 8-bits, even if deficient' to the implementation notes in the appendices, mainly as a warning to future generations of engineers:-).

We should definitely contact them so that they are aware of the existence of pdf producing software having this constraint.

Finally, note that just 'using the high bit' is no problem as these pdf numbers just correspond to negative TeX numbers.  The only problem is the 'missing number'.

car222222:
@blefloch @zauguin : 

About the name! As ever:-).

Although, as written "... it is true that the name is not unknown in C, java, python, lua.", 
the name in these languages (and throughout CS/Informatics/The World) does not in any sense refer to strings or arrays (or anything else) of characters, even if there are only two characters involved which happen to be named 0 and 1.  Nor does it refer to sequences/arrays of the integers 0 and 1.  

'bits' are totally different beasts so that this part of the name is completely wrong in our context.  

Yay!, using 'bits' for anything anywhere near to TeX is even more misleading, and wrong, than using 'FPUnit' for a 'FPLibrary'.  It is almost as bad as writing 'map to (a list etc.)' rather than the correct verb 'map over . . .'.

l3kernel/l3bitset.dtx
u-fischer: Imho the to-functions  are like the `\int_to`- functions: they create a formatted representation. \bitset_to_arabic doesn't create an integer variable or converts the bitset into one. Its output is not like \fp_to_int meant as something that is then usable by the int module.  Instead it prints digits.

car222222:
p@blefloch On modules: "a question about the name of the module"

A few extra points about this file, its contents not its name.
It may not be needed as a file at all, then we do not have to worry about not using 'the name' for it!

At the level of modules (or submodules) this file is currently a bit of mixture of distantly related, more fundamental, bits (puns intended!), roughly as follows:

flags (just one) -- although these are isomorphic to booleans (despite not being called true and false, with or without capitals), they maybe need their own niche somewhere.
  btw: sadly the name 'flag' in l3 is already taken for some weird expandable thingy:-)

very small arrays (almost certainly of size is slightly less than 31) of anything (is there any advantage in not having something uniform for all types of entries?) -- currently there is very little on arrays of any size (presumably because they are essentially non-expandable:-); they fit best into int since that is how they are indexed; maybe small arrays are, as @zauguin was perhaps suggesting, always just a special type of 'character token string' with a defined character set but it is not clear to me how to use this idea for arrays of non-trivial types.

small(?) binary numbers (integers only, or also fp?) - represented using characters 0, 1, -, +, ., ,, ??  ; plus maybe some arithmetic, and transformations such as \int_to_binarystr ; probably best add to some arithmetic module (probably for anything longer than about 4 binary digits a separator character is essential).

What did I miss? 

and NO bits (or even anything as small as a single byte).

l3kernel/l3bitset.dtx
u-fischer: I wonder if we can/should use fp for now for  larger bitsets. It isn't superfast for large bitsets (but probably the speeding-up-experts can do something here), but at least it doesn't error:

~~~~
\documentclass{article}
\begin{document}
\ExplSyntaxOn
\bitset_new:N \l_my_bitset
\bitset_set_true:Nn \l_my_bitset {1}
\bitset_set_true:Nn \l_my_bitset {25}
\bitset_set_true:Nn \l_my_bitset {14}

\cs_new:Npn \__bitset_bin_factor:n #1 { + \str_item:Nn\l_my_bitset{-#1} * 2**(#1-1) }

\cs_set:Npn \bitset_to_arabic:N #1
  {
    \int_compare:nNnTF {\str_count:N #1} < {32} %or 31 or 33??
       { \exp_args:Ne \int_from_bin:n { #1 } }
       {
         \fp_eval:n { \int_step_function:nN {\str_count:N #1 }\__bitset_bin_factor:n  }
       }
  }     
 
 
\bitset_to_arabic:N \l_my_bitset \par

\bitset_set_true:Nn \l_my_bitset {45}

\bitset_to_arabic:N \l_my_bitset
\ExplSyntaxOff
\end{document}
~~~~

davidcarlisle:
@car222222 

> I surveyed document languages and processing software and there are no cases where integers are larger than 8-bit

did you mean to write 32-bit rather than 8 there (otherwise I don't understand the comment at all)

C for example doesn't constrain the size of int, just has some constraints on the relative sizes of int and long, int has a _minimum_ size of 16 bits but may be larger.


python's default integer type is arbitrary size (it gets called long but that wouldn't be a good name in tex)


```
z=64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64*64
print(type(z))
print(z)
```

produces

```
<type 'long'>
1427247692705959881058285969449495136382746624
```


Closer to a document related use, XML Schema  when specifying integer types,  doesn't have bit but does have byte, and short and int, but these are specified by restricting the range from a base integer type not by restricting the underlying representation. An XSD byte is restricted  to the range -128 to 127, but there is no requirement that it takes less storage than an unrestricted integer (and is typically encoded as a sequence of characters with the decimal value, so taking much more than a byte.

As such I don't see anything surprising about encoding a sequence of bits as a sequence of characters 01.

car222222:
@davidcarlisle "As such I don't see anything surprising about encoding a sequence of bits as a sequence of characters 01."

Sure, but no one is encoding bits here, are they? We are dealing with sequences/arrays of flag values, which happen to be the character tokens 0 and 1.  'bits' are near-hardware things that are totally inaccessible from TeX and are not needed, even to represent anything here as we do not need the space efficiency of using bits, or even single bytes, to store anything. 

Or maybe you are not suggesting we use the term 'bit' here but you are talking about representing arrays as sequences of values?

car222222:
Yes, only out by a factor of 4!

car222222:
C etc are irrelevant here.  We dealt with programming languages earlier.  This is about document processing (also not for graphics).

car222222:
"python's default integer type is arbitrary size" does not feel much like an integer (or a data type come to that!) ??  

But good for python if it has native arbitrary precision rational arithmetic, every home should have one!

car222222:
Yes, I looked at the XML schema specs, since it explicitly deals in such things.  Not sure that it is document processing:-).
Maybe meta-doc-proc?? 

Anyways, nothing anywhere near to TeX will need anything arbitrary/big (for about 20 years or so, I predict!).

davidcarlisle:
@car222222 

> Sure, but no one is encoding bits here, are they? 

I don't think the name matters much (certainly doesn't matter as much as the length of this thread suggests:-)

You could call it `sequence_of_boolean_flags` but it's not a very catchy module name, and since the motivating use case is to fiddle the bits to construct an integer passed to PDF for the PDF processor to split up again, probably using C bitset functions, calling it bitset seems fine to me. The equivalent code in the Oberdiek package used `bitset` as the name for years without anything bad happening.


> Anyways, nothing anywhere near to TeX will need anything arbitrary/big (for about 20 years or so, I predict!).

I don't see that at all. For the "bitset" use case of generating an integer possibly not (but if allowing bigger cases simplifies the coding then no need to restrict)  But for the use case of storing a sequence of boolean flags, then an unrestricted length is perfectly natural.

As has been stated several times, if you think a different representation would simplify the tex coding here, make a PR then we can review that just as we are reviewing this.

car222222:
Yes, sure, I meant bigger integers not bigger arrays.  !!!  Not obvious? 

I shall write elsewhere about the iniquity (or inadvisability) of continuing to use quirky language such as 'map to' in the near future. 

But this philosophy of 'Heiko Good, New bad' had to go!

But since this file will probably not survive as an entity, its name dies not matter much.

car222222:
Nothing to do with any representation of anything.  Bruno was talking about the name.  Only Heiko has ever used the name 'bitset' and he was talking about something much more like an emulation of bit-twiddling stuff that we are not doing here.  

These 0s and 1s are just binary digits, not bits at all, so the use of 'bit' is misleading, unnecessary and . . . wrong:-)!  Not harmful, no.  But sill, yes!

l3kernel/l3bitset.dtx
blefloch: That's a good idea, but the implementation should better be something like

    \cs_set:Npn \bitset_to_int:N #1 % this is really a much better name than arabic
      {
        \int_compare:nNnTF { \str_count:N #1 } < {32}
          { \exp_args:Ne \int_from_bin:n {#1} } % why "Ne" and not just "No" ??
          {
            \exp_after:wN \@@_to_int:nN \exp_after:wN 0
            #1 \q_recursion_tail \q_recursion_stop
          }
      }
    \cs_new:Npn \@@_to_int:nN #1#2
      {
        \quark_if_recursion_tail_stop_do:Nn #2 {#1}
        \exp_args:Nf \@@_to_int:nN { \fp_eval:n { #1 * 2 + #2 } }
      }

Not tested but hopefully clear.

u-fischer: works fine (and the Ne is either a remainder from some older code or a typo).  But I don't agree really about the int. For me `int` has no intrinsic base. We have already `\bitset_to_bin`, and perhaps there will be a `\bitset_to_hex`.  `\bitset_to_int`  doesn't fit well (and we don't have an `\int_to_int:n`) .

l3kernel/l3bitset.dtx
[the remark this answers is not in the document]
u-fischer: the index is now 1-based, so it remains the question if the command should be `_index` or `item`  (this also affects the `named_index`  command).

PR #832, opened 2020-12-19
allow n-type arguments for keyval_parse

Skillmon:
In https://tex.stackexchange.com/questions/573812/ Ulrich asked why `\keyval_parse:NNn` is restricted to N-type arguments, so I've implemented the n-type arguments.

I've also ported a few minor performance tweaks from `expkv`. Netting a better performance in total (but the performance would be better without the n-type support, I'd estimate 3% to 5%).

------

Benchmarking results with `\keyval_parse:NNn \use_none:n \use_none:nn` and the following token list:

<code>,<sub>12</sub>,<sub>12</sub>key=<sub>13</sub>val,<sub>13</sub>key=<sub>12</sub>val,<sub>12</sub>key</code>

I've replicated that token list 1, 10, 100, and 1000 times. In the case of 1 the first comma is missing. The ratios shown below are the multiple of the run times needed; so in the case `\mykeyval_parse:nnn / \keyval_parse:NNn` with 0.963, this means that the time `\keyval_parse:NNn` takes times 0.963 is the time `\mykeyval_parse:nnn` takes.

`\keyval_parse:nnn` is the current implementation in the kernel, `\mykeyval_parse:nnn` is the implementation of this PR.

The results without the active commas and equal signs look similar (though the difference is smaller by 1% to 2% then.

```
1
\mykeyval_parse:nnn / \keyval_parse:NNn
0.963
0.963
0.963
\keyval_parse:NNn / \mykeyval_parse:nnn
1.033
1.033
1.033
10
\mykeyval_parse:nnn / \keyval_parse:NNn
0.968
0.968
0.968
\keyval_parse:NNn / \mykeyval_parse:nnn
1.037
1.037
1.037
100
\mykeyval_parse:nnn / \keyval_parse:NNn
0.972
0.972
0.972
\keyval_parse:NNn / \mykeyval_parse:nnn
1.026
1.026
1.026
1000
\mykeyval_parse:nnn / \keyval_parse:NNn
0.972
0.972
0.972
\keyval_parse:NNn / \mykeyval_parse:nnn
1.052
1.052
1.052
```

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,14 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+## [2020-12-19]
+
+### Added
+- `\keyval_parse:nnn`
+
+### Changed
+- `\keyval_parse:NNn` is set equal to `\keyval_parse:nnn`
+
 ## [2020-12-07]
 
 ### Fixed
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -1350,7 +1350,6 @@
             \@@_trim_auxiii:w
             #1 \s_@@_nil
             \@@_trim_auxiv:w
-          \s_@@_stop
         }
       \cs_new:Npn \@@_trim_auxi:w ##1 \s_@@_mark #1 ##2 \s_@@_mark ##3
         {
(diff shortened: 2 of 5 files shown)


josephwright:
Looks fine to me, other than we want an entry in the `CHANGELOG` ;)

Skillmon:
Whether this PR is accepted or not, I noticed that I made a mistake in the documentation of `\keyval_parse:NNn` earlier. It is currently marked as `EXP`, but it should be `rEXP`. I'll open an issue so this isn't lost (though the issue could be closed if this PR is accepted).

Skillmon:
> Looks fine to me, other than we want an entry in the `CHANGELOG` ;)

Plus tests for `\keyval_parse:nnn`, which I haven't done yet.

PR #835, opened 2020-12-27
L3bitset

u-fischer:
This installs l3bitset as an experimental package in l3experimental.


l3experimental/l3bitset/l3bitset.dtx
```diff
@@ -0,0 +1,637 @@
...
+%   \begin{syntax}
+%     \cs{bitset_addto_named_index:Nn} \meta{bitset var}
+%      \{
+%         \meta{name1} |=| \meta{index1} |,|
+%         \meta{name2} |=| \meta{index2} |,| \ldots{}
+%      \}
+%   \end{syntax}
+% This adds the name--index pairs given in the second
+% argument to an internal, global property list. If the
+% property list didn't exist yet, it is created.
```
FrankMittelbach, line 195: The fact that this is implemented as a property list seems irrelvant for documentation. By all means it could be done as csnames or what have you. Important on that level is only that this attaches a name to the index position and that this correspondance is global.

l3experimental/l3bitset/l3bitset.dtx
```diff
@@ -0,0 +1,637 @@
...
+%      ToggleNoView   = 9,
+%      LockedContents = 10
+%    }
+%  \end{verbatim}
+%  it is possible to set bit 3 by using
+%  |\bitset_set_true_key:Nn \l_pdfannot_F_bitset {Print}|
+%  If a key is unknown a warning is issued.
+% \begin{function}[added = 2020-12-13]
+%   {\bitset_set_true_key:Nn, \bitset_set_true_key:cn,
+%    \bitset_gset_true_key:Nn,\bitset_gset_true_key:cn }
```
FrankMittelbach, line 224: Not sure about the use of "key" yet, but assuming we use "key" then I hink those would be all better called `\bitset_key_set_true:Nn` etc, i.e., keep `bitset_key` together

l3experimental/l3bitset/l3bitset.dtx
```diff
@@ -0,0 +1,637 @@
...
+%   \begin{syntax}
+%     \cs{bitset_set_false_key:Nn}   \meta{bitset var}  \Arg{key}\\
+%     \cs{bitset_gset_false_key:Nn}  \meta{bitset var}  \Arg{key}
+%   \end{syntax}
+% This sets the bit of the position described by key to 0.
+% If the leading (left most) bit is unset, zeros are not
+% trimmed but stay in the bit vector.
+% \end{function}
+% \begin{function}[EXP,added = 2020-12-14]
+%   {\bitset_key:Nn}
```
FrankMittelbach, line 245: That would then be `\bitset_key_item:Nn`

l3experimental/l3bitset/l3bitset.dtx
```diff
@@ -0,0 +1,637 @@
...
+% \end{function}
+
+%
+% \section{Setting and unsetting bits}
+%
+% \begin{function}[added = 2020-12-13]
+%   { \bitset_set_true:Nn, \bitset_set_true:cn, \bitset_gset_true:Nn, \bitset_gset_true:cn  }
+%   \begin{syntax}
+%     \cs{bitset_set_true:Nn}   \meta{bitset var}  \Arg{integer expression}\\
+%     \cs{bitset_gset_true:Nn}  \meta{bitset var}  \Arg{integer expression}
```
FrankMittelbach, line 100: Is there a practical use case for supporting "integer expressions" as the second argument or is this just the case because it happens to be technically the case?

Because if not you could think of "numerical-index" and "named-index" just being 2 ways of refering to an index position and internally handle the difference, e.g. if position is unknown, check if it is a number and if so extend otherwise complain. Then there is no need to have different commands for named or position indexes. Just thinkingout loud, it may be less efficient compared to keeping them separate, but if most of not all actual use cases use named positions because that is much more readable) it may not make any difference in efficiency.

l3experimental/l3bitset/l3bitset.dtx
```diff
@@ -0,0 +1,637 @@
...
+% unlimited, only restricted by \TeX-memory. But the export function
+% \cs{bitset_to_arabic:N} can use functions from the \texttt{int} module only if
+% the largest index used for this bitset is smaller then 32, for longer
+% bitsets \texttt{fp} is used and this is slower.
+%
+% \section{Creating bitsets}
+% \begin{function}[added = 2020-12-13]
+%   {\bitset_new:N,\bitset_new:c}
+%   \begin{syntax}
+%     \cs{bitset_new:N}  \meta{bitset var} \\
```
FrankMittelbach, line 71: Not yet sure if it is better or worse to think of the named index positions as secondary class citizens. In my opinion (assuming the predominately use is named positions after all it might be better to have `\bitset_new:Nn` and set up the "named-index" by default together with the variable instead of doing this in a separate step.

l3experimental/l3bitset/l3bitset.dtx
u-fischer: Well I support integer expression as all other functions which take an integer as argument support them and I didn't see a reason why not.  I can't say how many of all use cases want named positions. The examples on tex.sx I found all used simple numbers, for example https://tex.stackexchange.com/a/110886/2388.  But imho the main question about your idea to merge numbers and names is the "check if it is a number" part. How fast is it? And can it also be done for the (expandable) `\bitset_item:Nn` ?

l3experimental/l3bitset/l3bitset.dtx
u-fischer: The question is then if one needs a command to extend the names, e.g. if someone doesn't like the name I choose and wants more readable ones.

l3experimental/l3bitset/l3bitset.dtx
u-fischer: ok. I will reword it.

l3experimental/l3bitset/l3bitset.dtx
FrankMittelbach: well that would depend on the implementation I guess. if the argument is just a key into a prop list, for example, you wouldn't need to check you would just use it (like you used the named keys, e.g. if in the named key prop you have 1=1 2=2 in addition to foo=1.  Clearly that is one step slower as it uses 2 indirection regardless of whether you use a name or a number (being now considered a name).  So I'm not saying one or the other is better, I was just offering alternatives to think about and my question was is the expression stuff ever being used because if then unifying is not an option as it means dropping the expression ability.
On the other hand, what you gain is a simpler more unified interface where names and numbers have the same status.

l3experimental/l3bitset/l3bitset.dtx
FrankMittelbach: I wasn't suggesting to drop the command for extending I was suggesting to offer a default list of names already on declaration.

l3experimental/l3bitset/l3bitset.dtx
FrankMittelbach: just to explain my thinking a bit more (and making it easier to tell me where I went the wrong turn): assuming we have in the named prop 1=1 2=2 ... for all bitset positions that are already in use, then your current `\bitset_key:Nn` would handle both named and numbered positions. Now when you set a position to true or false you would need to check if it is a known "name" (which you currently do for named keys) and if not, only then you would need to see if your name is a number and if so  extend your bitset and add further num=num to the name prop. That would then collapes all interface commands to what you currently use with "key" in the command name and make those that currently doesn't have "key" in the command name internal ie not showing up in user/developer documentation (you would still need them to do the actual work depending on your implementation).

u-fischer: Yes I understand the gain, and I like the idea in general and I'm thinking about it ;-).  But I wonder how to handle the "dynamic" growth. The bitset has no predefined length,  I don't know which 1=1, 2=2 will be needed. So one either has to drop the dynamic part and request a fix length at definition time (which imho would work too) or one has to decide what should happen if someone does `\bitset_set_true:N\l_my_bitset {42}` if the bitset itself has until now only 10 positions. How to check that 42 is an allowed position? Should one then fill the prop with 11=11, ..., 42=42?

FrankMittelbach: can't claim I have thought this through, but my initial though was that in the named prop you have 1=1 initially and whenever `\bitset_set_true:Nn` in your current implementation would require to extend the bitset I would add the corresponding num=num as well. So 
```
\bitset_new:N \l_foo_bitset                 % only 1=1
\bitset_set_trueNn\l_foo_bitset {3}    % add 2=2 and 3=3      (or maybe just 3=3? might be even better)
```

FrankMittelbach: by the way, checking for being a number really means checking for being a positive integer so that can be done very efficiently I guess.

u-fischer: Well someone from the efficiency league will have to tell me a suitable test, as the only expl3 test I have is with regex: `\regex_match:nnTF { ^[\d]+$ } {#1}` (but probably it would be okay, there won't be too many tests needed).  But I think if one do it this way, then one should have a public int for a minimal length *n* (with e.g. 32 as default value), and then create the names 1=1, ... ,*n*=*n* from the start on. Should one forbid that people define names like `10=1` or must this be handled?

FrankMittelbach: Was thinking somehting low-level such as
```
\newcount\foo
\def\test#1{\afterassignment\testi\foo=0#1 \testi}
\def\testi#1\testi{\if X\detokenize{#1}X%
                   \else \foo=0 \fi
  \ifnum \foo >0  \typeout{Number}\else\typeout{no number}\fi
}
  
  \test{5}
  \test{5a}
  \test{0}
  \test{foo}
  \test{ }
  \test{1}
```
As to the other questions: I would probably not create all the immediate ones if not used and I don't think that guarding against 10=1 is needed:  (a) it would still work depending on the implementation (only be rather strange) but more importantly (b) to me this is like guarding `\int_add:Nn` having and integer expression in the second arg... if I put in "foo" there it just blows badly.

u-fischer: I implemented your suggestion and imho it works fine.  Currently it uses the regex for the number test, but I will replace it by your more low-level code later.

PhelypeOleinik:
On Mon, Dec 28, 2020 at 2:17 PM Ulrike Fischer <notifications@github.com> wrote:
>
> the only expl3 test I have is with regex: \regex_match:nnTF { ^[\d]+$ } {#1}

I am against this, at the moment.  Nowhere else in expl3 we guess what
the argument means by parsing the input, instead use different functions
with different names (or signatures), and I see no good reason to break
that rule and have that here.  I think there should be two functions
instead, one to access the the bitset by name and another by index.

Phelype

PR #837, opened 2021-01-03
add pdf_destination commands to l3pdf

u-fischer:
These are two direct wrappers around the backend commands which are currently in pdfresources. Imho name and function are fine, and they can be moved to l3pdf.


l3experimental/l3pdf/l3pdf.dtx
```diff
@@ -174,6 +174,27 @@
 %   This function may only be used up to the point where the PDF file is
 %   initialised.
 % \end{function}
+% \subsection{Destinations}
+% % \begin{function}[added = 2021-01-03]
+%   {\pdf_destination:nn}
+%   \begin{syntax}
+%     \cs{pdf_destination:nn} \Arg{name} \Arg{type or integer}
```
josephwright, line 181: The 'keyword or scale' interface makes sense at the backend level, as it's essentially how the backends themselves work. However, I'm not so sure about this as a code-level interface.

l3experimental/l3pdf/l3pdf.dtx
```diff
@@ -174,6 +174,27 @@
...
+%   |fitv|, |fitb|, |fitbh|, |fitbv|, |fitr|, |xyz|
+%   or an integer representing a  scale factor in percent.
+%   The backend code defines |fitr| so that it will with pdflatex and
+%   lualatex use the coordinates of the surrounding box,
+%   with dvips and dvipdfmx it falls back to |fit|.
+% \end{function}
+% \begin{function}[added = 2020-01-03]
+%   {\pdf_destination_box:nn}
+%   \begin{syntax}
+%     \cs{pdf_destination_box:nn} \Arg{name} \Arg{content}
```
josephwright, line 193: As we box up here, is this a `pdf` function or is it an `hbox` one? If it's the former, do we like the name: again, the backend level name is a lot more 'flexible' than the code level.

l3experimental/l3pdf/l3pdf.dtx
```diff
@@ -174,6 +174,27 @@
...
 %   initialised.
 % \end{function}
+% \subsection{Destinations}
+% % \begin{function}[added = 2021-01-03]
+%   {\pdf_destination:nn}
+%   \begin{syntax}
+%     \cs{pdf_destination:nn} \Arg{name} \Arg{type or integer}
+%   \end{syntax}
+%   This creates a destination. \Arg{type or Integer} can be one of |fit|, |fith|,
+%   |fitv|, |fitb|, |fitbh|, |fitbv|, |fitr|, |xyz|
```
josephwright, line 184: We need to describe the meaning of the keywords, or at least reference the PDF spec, here.

l3experimental/l3pdf/l3pdf.dtx
u-fischer: Imho it is fine. The hyperref interface is more complicated, but actually only because it tries to have similar options for pdfview, pdfstartview and pdfremotestartview. If I had only `pdfview` to consider (and wouldn't have to support an existing interface) I would  use 'keyword or scale' too.  It is a command quite near to the backend, but I don't see a need for more abstraction here.

l3experimental/l3pdf/l3pdf.dtx
u-fischer: Well the main point is to set a destination, so I would say it is a pdf function. The question is if it has to typeset the content. And alternative here would be a 

      \pdf_destination_box:nnnn{name}{width}{height}{depth} 

which then gives a box of size 0 with a destination area spreading from there.

l3experimental/l3pdf/l3pdf.dtx
u-fischer: no problem, I can add it.

l3experimental/l3pdf/l3pdf.dtx
josephwright: We could have both interfaces, I guess?

u-fischer: FitR is the least used case (and imho the least useful). Having two commands for it alone, is overdoing. 

I wondered if one can have only one command, but it is not easy to find a sensible interface--I don't think that one want arguments that need splitting like  `\pdf_destination:nn{name}{zoom 1}` and `\pdf_destination:nn{fitr width height depth}`. What could work is 
`\pdf_destination:nn{fitr}` along with three external lengths, `\l_pdf_destination_fitr_width_dim`.

u-fischer: > What could work is \pdf_destination:nn{fitr} ...

That is actually nonsense, as `\pdf_destination:nn{fitr}` already exists, one would need a new option, e.g. `\pdf_destination:nn{fitr-box}`

josephwright: It occurs to me that the reason for the `box` backend function was as `FitR` didn't seem to work in the same way as other keywords. As there is now a `fitr` option for `\@@_backend_destination:nn`, and so for `\pdf_destination:nn`, do we need a separate `box`/`FitR` function at all?

u-fischer: We need it for the dvi backends and also for real control on the coordinates for pdflatex/lualatex. The `fitr` option of `@@_backend_destination:nn`  gives with pdflatex/lualatex the coordinates of the surrounding box and falls back to `fit` for the dvidrivers. That isn't bad, but not the same.

josephwright: We should document that aspect then

u-fischer: I added some documentation which hopefully makes this clearer. But  the question remains if the "real fitr" should measure/set a box, or be set with width/height/depth, or if it should be integrated somehow in `\pdf_destination:nn` and if yes how.

PR #842, opened 2021-01-23
Faster expandable sort

Skillmon:
This PR adds a code variant for `\tl_sort:nN` making use of `\expanded` when it's available, as a result speeding up the code considerably (almost 30 times faster for 10000 items, 1.6 times fast for 1 element).

It also adds an expandable variant to sort `clist`s, which is a bit slower because it has to sanitize a `clist` and convert the sorting result from a sorted token list to a `clist` again. The variant without `\expanded` is pretty simple, the variant with `\expanded` is much more complicated, incorporating the input sanitisation. The code could be reduced by using `\clist_map_function:nN`similar to how it is done for the variant without `\expanded`.

### DIFF
--- a/l3kernel/l3clist.dtx
+++ b/l3kernel/l3clist.dtx
@@ -351,6 +351,23 @@
 %   described in Section~\ref{sec:l3sort:mech}.
 % \end{function}
 %
+% \begin{function}[added = 2021-01-23, EXP]{\clist_sort:NN, \clist_sort:nN}
+%   \begin{syntax}
+%     \cs{clist_sort:NN} \meta{clist var} \meta{conditional}
+%   \end{syntax}
+%   Sorts the items in the \meta{clist var}, using the \meta{conditional} to
+%   compare the items, and leaves the result in the input stream as a |clist|.
+%   The \meta{conditional} should have signature |:nnTF|, and return
+%   \texttt{true} if the two items being compared should be left in the same
+%   order, and \texttt{false} if the items should be swapped. The details of
+%   sorting comparison are described in Section~\ref{sec:l3sort:mech}.
+%   \begin{texnote}
+%     The result is returned within \cs{exp_not:n}, which means that the
+%     token list does not expand further when appearing in an
+%     \texttt{x}-type or \texttt{e}-type argument expansion.
+%   \end{texnote}
+% \end{function}
+%
 % \section{Comma list conditionals}
 %
 % \begin{function}[EXP,pTF]{\clist_if_empty:N, \clist_if_empty:c}
@@ -1517,7 +1534,10 @@
 % \end{macro}
 %
 % \begin{macro}
-%   {\clist_sort:Nn, \clist_sort:cn, \clist_gsort:Nn, \clist_gsort:cn}
+%   {
+%     \clist_sort:Nn, \clist_sort:cn, \clist_gsort:Nn, \clist_gsort:cn
+%     \clist_sort:nN, \clist_sort:NN
+%   }
 %   Implemented in \pkg{l3sort}.
 % \end{macro}
 %
--- a/l3kernel/l3sort.dtx
+++ b/l3kernel/l3sort.dtx
@@ -1321,6 +1321,335 @@
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}[EXP]{\clist_sort:nN,\clist_sort:NN}
+%   Sorting a |clist| can be done in the same way we sort a |tl|, we just have
+%   to convert the |clist| into the same representation used by \cs{tl_sort:nN}
+%   internally during our first bottom up pass and convert back to a |clist|
+%   after we're done. The \cs{clist_sort:NN} variant can take a shortcut in
+%   the bottom up pass as a |clist| variable would already be sanitized. For the
+%   |n| variant we place \cs{prg_do_nothing:} infront of the list to avoid
+%   accidental brace stripping.
+%
+%   In the case that \cs{tex_expanded:D} isn't available we fall back by
+%   converting the |clist| into a token list and using the same sorting function
+%   inside an |f|-expansion. The result is then turned into a |clist|. For the
+%   conversion to a token list we use \cs{clist_map_function:NN} for the
+%   |N|-type and \cs{clist_map_function:nN} for the |n|-type variant.
+%    \begin{macrocode}
+\cs_if_exist:NTF \tex_expanded:D
+  {
+    \cs_new:Npn \clist_sort:nN #1#2
+      {
+        \__kernel_exp_not:w
+          \exp_after:wN \@@_merge_exp_after_bottom_up_clist:Nw \exp_after:wN #2
+          \tex_expanded:D
+            {
+              \@@_merge_exp_bottom_up_clist_n:Nw
+                #2 \prg_do_nothing: #1 , \s_@@_merge_exp_mark ,
+            }
+      }
+    \cs_new:Npn \clist_sort:NN #1#2
+      {
+        \__kernel_exp_not:w
+          \exp_after:wN \@@_merge_exp_after_bottom_up_clist:Nw \exp_after:wN #2
+          \tex_expanded:D
+            {
+              \exp_after:wN \@@_merge_exp_bottom_up_clist_N:Nw \exp_after:wN
+                #2 #1 , \s_@@_merge_exp_mark , \s_@@_merge_exp_stop ,
+            }
+      }
+  }
+  {
+    \cs_new:Npn \clist_sort:nN #1#2
+      {
+        \exp_not:f
+          {
+            \clist_map_function:nN {#1} \@@_quick_prepare_clist:nw
+            \@@_quick_prepare_clist_result:nN {} #2
+          }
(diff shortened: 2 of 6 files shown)


l3kernel/l3clist.dtx
```diff
@@ -351,6 +351,23 @@
...
+%   \texttt{true} if the two items being compared should be left in the same
+%   order, and \texttt{false} if the items should be swapped. The details of
+%   sorting comparison are described in Section~\ref{sec:l3sort:mech}.
+%   \begin{texnote}
+%     The result is returned within \cs{exp_not:n}, which means that the
+%     token list does not expand further when appearing in an
+%     \texttt{x}-type or \texttt{e}-type argument expansion.
+%   \end{texnote}
+% \end{function}
+%
```
blefloch, line 370: We have a non-expandable `\tl_reverse:N` modifying a tl variable, and an expandable `\tl_reverse:n` that leaves its result in the input stream.  To do the second (expandable) operation on the contents of a token list variable, we use `\tl_reverse:V`, which is not a synonym of `\tl_reverse:N`.

For the same reason we shouldn't implement an expandable `\clist_sort:NN` (clist var) (comparison) that would be likely to get confused with the non-expandable assignment `\clist_sort:Nn (clist var) (comparison)`.  Instead we should implement the variant `\clist_sort:VN` of `\clist_sort:nN`.

Alternatively, it might be better to have `\clist_sorted` for the expandable version, like Python's `sorted(l)` versus `l.sort()`, but perhaps the ship has sailed already given the expandable `\tl_sort:n` and the expandable `\tl_reverse:n` (which with this logic should be `\tl_reversed:n`).

l3kernel/l3tl.dtx
```diff
@@ -891,6 +891,19 @@
...
+%   Like \cs{tl_sort:nN} but this sorts a \meta{tl var}, leaving the result in
+%   the input stream.
+%   \begin{texnote}
+%     The result is returned within \cs{exp_not:n}, which means that the
+%     token list does not expand further when appearing in an
+%     \texttt{x}-type or \texttt{e}-type argument expansion.
+%   \end{texnote}
+% \end{function}
+%
 % \section{The first token from a token list}
```
blefloch, line 907: Same comment as for the expandable `\clist_sort:NN`, I think this expandable `\tl_sort:NN` is too likely to get confused with the assignment `\tl_sort:Nn`.

l3kernel/testfiles/m3sort002.lvt
```diff
@@ -29,6 +29,16 @@
...
+    \NEWLINE
+    \tl_sort:nN { 815 } \test_compare:nnTF
+    \NEWLINE
+    \tl_sort:nN { {\if_true: 8\fi:}15 } \test_compare:nnTF
+    \NEWLINE
+    \tl_sort:nN { {\if_true: 8\else: 6\fi:}15 } \test_compare:nnTF
+    \NEWLINE
+    \tl_sort:nN { {\if_false: 8\else: 6\fi:}15 } \test_compare:nnTF
+    \NEWLINE
     \tl_sort:nN { 8{+2}3461{-0}2{00}3748 } \test_compare:nnTF
```
blefloch, line 42: It's a good idea to test primitive conditionals there.  Could you replace some of these by unbalanced conditionals (that are still valid integer expressions) such as `\use_none:n \if:w 7` or `0 \use_none:n \fi:`?

l3kernel/testfiles/m3sort002.lvt
```diff
@@ -50,7 +60,91 @@
...
+\clist_new:N \l_test_many_clist
+\clist_new:N \l_test_I_clist
+\clist_new:N \l_test_II_clist
+\clist_new:N \l_test_III_clist
+\clist_new:N \l_test_IV_clist
+\clist_set:Nn \l_test_I_clist { \ZERO }
+\clist_set:Nn \l_test_II_clist { 1,\ZERO }
+\clist_set:Nn \l_test_III_clist { 1,\ZERO,2 }
+\clist_set:Nn \l_test_IV_clist { 1,\ZERO,3,2 }
+\clist_set:Nn \l_test_many_clist { 5,1,7,8,10,{1~} }
```
blefloch, line 116: Could you please add some negative integers, and perhaps more duplicates (e.g., `1,{1~},1,{~1}`) to show whether the sort preserves the order?

l3kernel/l3sort.dtx
```diff
@@ -968,31 +1021,641 @@
...
+%   After the first bottom up pass every sublist will be sorted and followed by
+%   \cs{s_@@_merge_exp_tail}. The sublists are separated by
+%   \cs{s_@@_merge_exp_list}, and the end of sublists is indicated by a sublist
+%   containing only \cs{s_@@_merge_exp_mark} and another one containing
+%   \cs{s_@@_merge_exp_stop}.
+%    \begin{macrocode}
+    \cs_new:Npn \@@_merge_exp_bottom_up_tl:Nnn #1#2#3
+      {
+        \@@_if_stop:w #3 \@@_merge_exp_bottom_up_tl_stop:w \s_@@_merge_exp_stop
+        \@@_if_mark:w #3 \@@_merge_exp_bottom_up_tl_mark:w \s_@@_merge_exp_mark
```
blefloch, line 1095: The usual trick to speed up the loop is to have two end-markers (like `{a}{b}{c}\s_mark \s_mark \s_stop`) and then test whether the `#3` argument is the end-marker.  Then you can refine to distinguish even/odd by testing the `#2` argument.  Actually, if you have two end-markers like this, you should be able to do both tests in one go more cleanly as `\@@_if_mark:w #2 \@@_merge_even:w #3 \@@_merge_odd:w \s_@@_merge_exp_mark`.

I have to say, your use of `\@@_if_mark:w` and similar to end the loop is really nice in comparison to our much slower quark tests in `\tl_map_function:nN`.

l3kernel/l3sort.dtx
```diff
@@ -199,11 +199,26 @@
...
 %    \begin{macrocode}
 \scan_new:N \s_@@_mark
 \scan_new:N \s_@@_stop
+\cs_if_exist:NT \tex_expanded:D
+  {
+    \scan_new:N \s_@@_merge_exp_mark
+    \scan_new:N \s_@@_merge_exp_stop
+    \scan_new:N \s_@@_merge_exp_tail
+    \scan_new:N \s_@@_merge_exp_list
+  }
```
blefloch, line 221: I would prefer slightly shorter names, perhaps dropping `exp`.  Same comment for most of the auxiliaries later in this file, actually.  They don't need to say quite so precisely what they do.  String space is not completely free.

l3kernel/l3sort.dtx
```diff
@@ -968,31 +1021,641 @@
...
+        \s_@@_merge_exp_stop
+        \@@_if_mark:w \s_@@_merge_exp_stop \@@_merge_exp_bottom_up_tl_mark:w
+        \s_@@_merge_exp_mark
+        #1
+        \s_@@_merge_exp_tail \s_@@_merge_exp_list
+        \@@_merge_exp_bottom_up_tl:Nnn #2
+      {
+        \s_@@_merge_exp_mark \s_@@_merge_exp_list
+        \s_@@_merge_exp_stop \s_@@_merge_exp_list
+      }
```
blefloch, line 1112: I think this level of detail in the delimiters of macro parameters is not readable.  Strictly speaking you are right that it is faster for TeX, but I would still go for the slightly slower `\cs_new:Npn \@@_merge_exp_bottom_up_tl_stop:w #1 \s_@@_merge_exp_list #2#3`.  Maybe I'm wrong though.

blefloch (changes requested):
Thanks a lot for the pull request, and sorry that I took infinitely long to get to it.  I haven't yet fully understood the new sorting code, especially haven't looked at the clist sorting yet, but there are already some changes I suggest, that could simplify the code.

l3kernel/l3clist.dtx
Skillmon: The reasoning I had to not call it `\clist_sort:VN` was that it doesn't do a `V`-type expansion, but only works correctly for a `clist`-variable (with `V` one could expect it to work with a `tl` as well, but this would only really do the expected thing if that `tl` was also a correctly sanatised `clist` just somehow stored in a `tl`).

But I'm happy with whatever name you pick.

josephwright:
@Skillmon Just looking back here: I see @blefloch asked for some changes which are pending. Are they likely to happen - if so I think we might then move to merge.

josephwright:
In particular, we now have (finaly) required `\expanded` and I'm resolving https://github.com/latex3/latex3/issues/914, so probably `\clist_sort_use:`, etc., will be sensible/available.

josephwright:
I'm looking back - what's the status here?

Skillmon:
Unchanged; put on hold. I need to get a bit of work on it done. I did restart some work on it in the past, but I fear that's lost to time and my memory is again too shady to salvage what I already did. But it's been only two years, so guess I'll still be in time for a full LaTeX3 release :P

Bad jokes aside, current personal L3-plan of action: Get myself motivated to write the `tl_act`-PR. After that I can tackle this one again.

PR #846, opened 2021-02-03
add dflt as alias for utf8/native to str-convert

u-fischer:
Currently one has to use depending on the engine

    \str_set_convert:Nnnn \l_tmpa_str {Grüße} {utf8} {utf16/hex}
 
or

    \str_set_convert:Nnnn \l_tmpa_str {Grüße} {} {utf16/hex}

to convert utf8. This is quite a nuisance and forces engines tests in various places.

This PR adds an encoding `dflt` which is simply an alias for utf8 with (pdf)latex and for the native encoding with the unicode engines.

(side question why does the test do `\tl_new:N \l_tmpa_str`?)


l3kernel/l3str-convert.dtx
```diff
@@ -317,6 +319,17 @@
...
+\bool_lazy_any:nTF
+  {
+    \sys_if_engine_luatex_p:
+    \sys_if_engine_xetex_p:
+  }
+  {
+    \prop_gput:Nnn \g__str_alias_prop { dflt } {  }
+  }
+  {
+    \prop_gput:Nnn \g__str_alias_prop { dflt } { utf8 }
```
PhelypeOleinik, line 331: `@@` instead of `_str`

PhelypeOleinik:
> (side question why does the test do `\tl_new:N \l_tmpa_str`?)

Probably was added there before `l3str` and never removed

car222222:
My investigations had also led me to question the idea of 'native encoding' and the meaning of 'empty'.  

This addition may help!

l3kernel/l3str-convert.dtx
u-fischer: grr, that happens if you copy from a test file. Corrected

l3kernel/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+- `dflt` alias to str_convert
```
FrankMittelbach, line 11: Is there any real reason for saving the vocals in that name? I think it is rather strange looking (assuming it stands for "default"

car222222:
Oh I see now:  dflt = default !!  Took me a while to decipher that:-).
Fear of vowels is curable.  
But now I can add simply add it to the list of things I do not understand in the table.

l3kernel/CHANGELOG.md
u-fischer: no, default is fine too, I will change to it. I only started with variants of utf8xx when struggling to find a name and so was in "abbreviation mode" and then the open type `dflt` jumped into my mind.

josephwright (approved):
Looks good to me: I assume we feel that simply making `utf8` the same as `native` on Unicode engines is mislead.

u-fischer:
> Looks good to me: I assume we feel that simply making `utf8` the same as `native` on Unicode engines is mislead.

I thought about it, but it is at first not trivial to do it (simply setting an alias like here didn't work), and it would change the documentated behaviour.

PR #847, opened 2021-02-04
Drop contrib tests

josephwright:


FrankMittelbach (commented):
lots of deletes :-) should make things simpler

PR #851, opened 2021-02-11
Move l3pdf to l3kernel

josephwright:


l3kernel/l3pdf.dtx
```diff
@@ -271,16 +263,20 @@
...
+  { \str_if_eq_p:Vn \fmtname { LaTeX2e } }
+  { \tl_if_exist_p:N \@expl@finalise@setup@@@@ }
   {
-    \AtBeginDocument
-      { \bool_gset_true:N \g_@@_init_bool }
+    \tl_put_right:Nn \@expl@finalise@setup@@@@
+      {
+        \AtBeginDocument
+          { \bool_gset_true:N \g_@@_init_bool }
+      }
```
FrankMittelbach, line 278: We should not use \AdBeginDocument but place all this stuff into kernel hooks next to that public hook.

FrankMittelbach (changes requested):

l3kernel/l3pdf.dtx
josephwright: I've moved to `\@kernel@after@begindocument` - hopefully about the right place (it's supposed to come once the PDF is initialised/structures are fixed).

u-fischer: shouldn't pdf_object_now and pdf_object_write change the flag too?

FrankMittelbach: > hopefully about the right place

where it has to go is something you and Ulrike should understand better than me :-). My point was simply that the kernel code should use kernel hooks and not populate public ones unless there is the need that the public should interact with it (in ordering) which I doubt is the case here. each code chunk will slow down the document processing a bit and we should avoid this.

josephwright:
@car222222  I'm only asking to merge those functions that are presented here: any new material would be subject to separate PRs. In particular, all of the PDF resource management material is entirely separate.

car222222:
OK, I should look at what is 'presented here' then.

l3kernel/l3pdf.dtx
josephwright: I did wonder about `\everypar`, as in some ways we can wait until there is text. But someone is bound to put some `\pdfliteral` straight after `\begin{document}`, so it's a reasonable balance I hope. (@u-fischer?)

FrankMittelbach: You could delay to `\AddToHookNext{para/before}` but it has the danger of being too late (comes after the normal public hook there and as you say some \pdfliteral may get into the way as well) and putting it in `\everypar` would be even worse (and later). You could put it into the kernel hook there, but then you would need to manage it manually as that is not a onetime hook (obviously).

u-fischer: The threshold is not "text" but if something has opened the pdf, at least with pdflatex. beamer e.g. initializes some objects, and this will already give a fatal error with pdflatex (with l3pdf loaded naturally):

~~~~
\documentclass{beamer}
\ExplSyntaxOn
\pdf_version_gset:n{1.7}
\ExplSyntaxOff
~~~~

But the problem is that the engines differ: with xelatex the example works, (but after \begin{document} is too late here). This means that setting the boolean in some begin document hook will catch some cases but not all and marks the latest possible point, so it is okay, but it doesn't pay off to put to much thought considering which hook is the best. I would document for `\pdf_uncompress:` and `\pdf_version_gset:n` that they should be used as early as possible, e.g before \documentclass (or through the new interface, which will be safe ...).

josephwright: > The threshold is not "text" but if something has opened the pdf, at least with pdflatex. beamer e.g. initializes some objects, and this will already give a fatal error with pdflatex (with l3pdf loaded naturally):
> 
> ```
> \documentclass{beamer}
> \ExplSyntaxOn
> \pdf_version_gset:n{1.7}
> \ExplSyntaxOff
> ```
> 
> But the problem is that the engines differ: with xelatex the example works, (but after \begin{document} is too late here). This means that setting the boolean in some begin document hook will catch some cases but not all and marks the latest possible point, so it is okay, but it doesn't pay off to put to much thought considering which hook is the best. I would document for `\pdf_uncompress:` and `\pdf_version_gset:n` that they should be used as early as possible, e.g before \documentclass (or through the new interface, which will be safe ...).

Well of course this is back with the idea that really 'everything' that writes to the PDF should wait until a single defined point, likely somewhere in `\begin{document}`. I guess we are trying to find a balance: if you set th ePDF version in the preamble, it's likely that you accept that it's now fixed. 

We can I think hope to clean up stuff over the medium term, e.g. `beamer`. That's certainly true if everything can go via interfaces that know about setting a 'PDF set up' flag.

u-fischer: A clear border would be useful, but I'm not sure yet, where it should be really be at the end. begin document sounds sensible, but I don't know how many packages beside beamer violate this rule. xobjects or shadings etc imho are problematic too, and these are perhaps stored in the preamble (as we know  people do typesetting tasks there, even if they shouldn't).

FrankMittelbach: > in the preamble (as we know people do typesetting tasks there, even if they shouldn't).

but you can only do typesetting in a box in the preamble. Would that open the pdf file or cause a problem then?

u-fischer: Yes, e.g. a tikz shade would try to write a shading object and so this will error with pdflatex. 

~~~~
\RequirePackage{l3pdf} %if not in this branch
\documentclass{article}
 \usepackage{tikz}
 \setbox0\hbox{\tikz\draw [shade, left color=red, right color=blue] (0,0) rectangle (1,1);}
 \ExplSyntaxOn
 \pdf_version_gset:n{1.7}
 \ExplSyntaxOff
 
 \begin{document}
 
 blub
 \end{document}
~~~~

That is not unsolvable---tikz could delay the writing of the object.

josephwright: We could adjust the PDF backend code such that our stuff at least delays all operations to a 'safe' and known point: that's separate from this PR but is doable.

u-fischer: yes, I thought about that too. But one the other side only two commands have problems: uncompress does nothing and version errors. Both aren't commands that users/packages will use often.

car222222: Is it necessary to provide/expose commands for these two for potential use in the preamble?

Since they are both clearly setting MetaData the only "public/advertised", and safe, interface should be whatever we provide for setting this at (or before) the start of the file (inn the prepreamble:-), such as:

\DeclareDocumentMetadata   ;
A configuration file   ;
A fixed \documentclass command

car222222: Setting these two in code in packages etc. is similarly problematic: should any such setting of the pdf version be allowed to over-ride the default setting (from the format or in a configuration file)?  Should compression level ever be the concern of a class or package?

l3kernel/l3pdf.dtx
```diff
@@ -97,15 +94,15 @@
...
 %   \begin{syntax}
 %     \cs{pdf_object_ref:n} \Arg{object}
 %   \end{syntax}
 %   Inserts the appropriate information to reference the \meta{object}
 %   in for example page resource allocation
 % \end{function}
 %
-% \begin{function}[added = 2019-06-02]
+% \begin{function}[added = 2021-02-10]
 %   {\pdf_object_now:nn, \pdf_object_now:nx}
```
u-fischer, line 106: From today meeting: This function should be  named `\pdf_object_unnamed_write:nn`

l3kernel/l3pdf.dtx
```diff
@@ -123,7 +120,7 @@
 %   \end{itemize}
 % \end{function}
 %
-% \begin{function}[EXP, added = 2019-06-02]{\pdf_object_last:}
+% \begin{function}[EXP, added = 2021-02-10]{\pdf_object_last:}
```
u-fischer, line 123: From todays meeting: this should be named `\pdf_object_ref_last:`

u-fischer:
The command to reference the pageobject can be added too (I can then remove it from pdfresources).

~~~~
\cs_new:Npn \pdf_pageobject_ref:n #1 { \@@_backend_pageobject_ref:n { #1 }}
~~~~

l3kernel/l3pdf.dtx
u-fischer: @car222222 > Is it necessary to provide/expose commands for these two for potential use in the preamble?
Yes. `\pdf_uncompress:` is needed by any developer working on pdf. `\pdf_gset_version:n` will be needed by class writers: some journals or institutions will want to force a version.  I think this part is fine for now..

PR #854, opened 2021-02-17
Split kernel messages by module

josephwright:
Messages that are truly cross-module are left as kernel,
plus those in the cs/prg modules (truly core), plus one
or two at the 'loading' stage where this seems the best description.

This is step one of looking at removing/reworking \__kernel_msg_ support.

### DIFF
--- a/l3kernel/l3clist.dtx
+++ b/l3kernel/l3clist.dtx
@@ -2012,15 +2012,15 @@
 %   Apply the general \cs{__kernel_chk_defined:NT} and
 %   \cs{msg_show:nnnnnn}.
 %    \begin{macrocode}
-\cs_new_protected:Npn \clist_show:N { \@@_show:NN \msg_show:nnxxxx }
+\cs_new_protected:Npn \clist_show:N { \@@_show:NN \__kernel_msg_show:nnxxxx }
 \cs_generate_variant:Nn \clist_show:N { c }
-\cs_new_protected:Npn \clist_log:N { \@@_show:NN \msg_log:nnxxxx }
+\cs_new_protected:Npn \clist_log:N { \@@_show:NN \__kernel_msg_log:nnxxxx }
 \cs_generate_variant:Nn \clist_log:N { c }
 \cs_new_protected:Npn \@@_show:NN #1#2
   {
     \__kernel_chk_defined:NT #2
       {
-        #1 { LaTeX / clist } { show }
+        #1 { clist } { show }
           { \token_to_str:N #2 }
           { \clist_map_function:NN #2 \msg_show_item:n }
           { } { }
@@ -2033,11 +2033,11 @@
 %   A variant of the above: no existence check, empty first argument for
 %   the message.
 %    \begin{macrocode}
-\cs_new_protected:Npn \clist_show:n { \@@_show:Nn \msg_show:nnxxxx }
-\cs_new_protected:Npn \clist_log:n { \@@_show:Nn \msg_log:nnxxxx }
+\cs_new_protected:Npn \clist_show:n { \@@_show:Nn \__kernel_msg_show:nnxxxx }
+\cs_new_protected:Npn \clist_log:n { \@@_show:Nn \__kernel_msg_log:nnxxxx }
 \cs_new_protected:Npn \@@_show:Nn #1#2
   {
-    #1 { LaTeX / clist } { show }
+    #1 { clist } { show }
       { } { \clist_map_function:nN {#2} \msg_show_item:n } { } { }
   }
 %    \end{macrocode}
--- a/l3kernel/l3clist.dtx
+++ b/l3kernel/l3clist.dtx
@@ -2020,7 +2020,7 @@
   {
     \__kernel_chk_defined:NT #2
       {
-        #1 { LaTeX/kernel } { show-clist }
+        #1 { LaTeX / clist } { show }
           { \token_to_str:N #2 }
           { \clist_map_function:NN #2 \msg_show_item:n }
           { } { }
@@ -2037,7 +2037,7 @@
 \cs_new_protected:Npn \clist_log:n { \@@_show:Nn \msg_log:nnxxxx }
 \cs_new_protected:Npn \@@_show:Nn #1#2
   {
-    #1 { LaTeX/kernel } { show-clist }
+    #1 { LaTeX / clist } { show }
       { } { \clist_map_function:nN {#2} \msg_show_item:n } { } { }
   }
 %    \end{macrocode}
--- a/l3kernel/l3coffins.dtx
+++ b/l3kernel/l3coffins.dtx
@@ -566,7 +566,7 @@
     \coffin_if_exist:NTF #1
       { #2 }
       {
-        \__kernel_msg_error:nnx { coffin } { unknown-coffin }
+        \__kernel_msg_error:nnx { coffin } { unknown }
           { \token_to_str:N #1 }
       }
   }
@@ -913,7 +913,7 @@
     \prop_get:cnNF
       { coffin ~ \@@_to_value:N #1 ~ poles } {#2} #3
       {
-        \__kernel_msg_error:nnxx { coffin } { unknown-coffin-pole }
+        \__kernel_msg_error:nnxx { coffin } { unknown-pole }
           { \exp_not:n {#2} } { \token_to_str:N #1 }
         \tl_set:Nn #3 { { 0pt } { 0pt } { 0pt } { 0pt } }
       }
@@ -2463,7 +2463,7 @@
   {
     \@@_if_exist:NT #2
       {
-        #1 { coffin } { show-coffin }
+        #1 { coffin } { show }
           { \token_to_str:N #2 }
           {
             \iow_newline: >~ ht ~=~ \dim_eval:n { \coffin_ht:N #2 }
@@ -2491,16 +2491,16 @@
     but~they~do~not~have~a~unique~meeting~point:~
     the~value~(0pt,~0pt)~will~be~used.
   }
-\__kernel_msg_new:nnnn { coffin } { unknown-coffin }
+\__kernel_msg_new:nnnn { coffin } { unknown }
   { Unknown~coffin~'#1'. }
   { The~coffin~'#1'~was~never~defined. }
-\__kernel_msg_new:nnnn { coffin } { unknown-coffin-pole }
+\__kernel_msg_new:nnnn { coffin } { unknown-pole }
   { Pole~'#1'~unknown~for~coffin~'#2'. }
   {
     LaTeX~was~asked~to~find~a~typesetting~pole~for~a~coffin,~
     but~either~the~coffin~does~not~exist~or~the~pole~name~is~wrong.
   }
-\__kernel_msg_new:nnn { coffin } { show-coffin }
+\__kernel_msg_new:nnn { coffin } { show }
   {
     Size~of~coffin~#1 : #2 \\
     Poles~of~coffin~#1 : #3 .
--- a/l3kernel/l3coffins.dtx
+++ b/l3kernel/l3coffins.dtx
@@ -2454,16 +2454,16 @@
 %   structure then the code complains.
 %    \begin{macrocode}
 \cs_new_protected:Npn \coffin_show_structure:N
-  { \@@_show_structure:NN \msg_show:nnxxxx }
+  { \@@_show_structure:NN \__kernel_msg_show:nnxxxx }
 \cs_generate_variant:Nn \coffin_show_structure:N { c }
 \cs_new_protected:Npn \coffin_log_structure:N
-  { \@@_show_structure:NN \msg_log:nnxxxx }
+  { \@@_show_structure:NN \__kernel_msg_log:nnxxxx }
 \cs_generate_variant:Nn \coffin_log_structure:N { c }
 \cs_new_protected:Npn \@@_show_structure:NN #1#2
   {
     \@@_if_exist:NT #2
       {
-        #1 { LaTeX / coffins } { show-coffin }
+        #1 { coffins } { show-coffin }
           { \token_to_str:N #2 }
           {
             \iow_newline: >~ ht ~=~ \dim_eval:n { \coffin_ht:N #2 }
--- a/l3kernel/l3coffins.dtx
+++ b/l3kernel/l3coffins.dtx
@@ -566,7 +566,7 @@
     \coffin_if_exist:NTF #1
       { #2 }
       {
-        \__kernel_msg_error:nnx { coffins } { unknown-coffin }
+        \__kernel_msg_error:nnx { coffin } { unknown-coffin }
           { \token_to_str:N #1 }
       }
   }
@@ -913,7 +913,7 @@
     \prop_get:cnNF
       { coffin ~ \@@_to_value:N #1 ~ poles } {#2} #3
       {
-        \__kernel_msg_error:nnxx { coffins } { unknown-coffin-pole }
+        \__kernel_msg_error:nnxx { coffin } { unknown-coffin-pole }
           { \exp_not:n {#2} } { \token_to_str:N #1 }
         \tl_set:Nn #3 { { 0pt } { 0pt } { 0pt } { 0pt } }
       }
@@ -1147,7 +1147,7 @@
         \l_@@_pole_a_tl \l_@@_pole_b_tl
     \bool_if:NT \l_@@_error_bool
       {
-        \__kernel_msg_error:nn { coffins } { no-pole-intersection }
+        \__kernel_msg_error:nn { coffin } { no-pole-intersection }
         \dim_zero:N \l_@@_x_dim
         \dim_zero:N \l_@@_y_dim
       }
@@ -2463,7 +2463,7 @@
   {
     \@@_if_exist:NT #2
       {
-        #1 { coffins } { show-coffin }
+        #1 { coffin } { show-coffin }
           { \token_to_str:N #2 }
           {
             \iow_newline: >~ ht ~=~ \dim_eval:n { \coffin_ht:N #2 }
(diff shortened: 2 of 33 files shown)


l3kernel/l3coffins.dtx
```diff
@@ -2463,7 +2463,7 @@
   {
     \@@_if_exist:NT #2
       {
-        #1 { LaTeX / kernel } { show-coffin }
+        #1 { LaTeX / coffins } { show-coffin }
           { \token_to_str:N #2 }
```
FrankMittelbach, line 2467: Still have problems to understand why a few messages are `LaTeX/coffins`or `LaTeX/clist` but most others `\__kernel` ones

l3kernel/l3file.dtx
```diff
@@ -3372,7 +3372,7 @@
       }
     \seq_concat:NNN \l_@@_tmp_seq \l_@@_tmp_seq \g_@@_record_seq
     \seq_remove_duplicates:N \l_@@_tmp_seq
-    #1 { LaTeX/kernel } { file-list }
+    #1 { LaTeX / kernel } { file-list }
       { \seq_map_function:NN \l_@@_tmp_seq \@@_list_aux:n }
```
FrankMittelbach, line 3376: ???

l3kernel/l3intarray.dtx
```diff
@@ -517,7 +517,7 @@
   {
     \__kernel_chk_defined:NT #2
       {
-        #1 { LaTeX/kernel } { show-intarray }
+        #1 { LaTeX / intarray } { show }
           { \token_to_str:N #2 }
```
FrankMittelbach, line 521: ???

l3kernel/l3msg.dtx
```diff
@@ -1736,30 +1736,30 @@
 % Error messages needed to actually implement the message system
 % itself.
 %    \begin{macrocode}
-\__kernel_msg_new:nnnn { kernel } { message-already-defined }
+\__kernel_msg_new:nnnn { msg } { message-already-defined }
   { Message~'#2'~for~module~'#1'~already~defined. }
```
FrankMittelbach, line 1740: elsewhere we have shorter second args, ie only "already-defined" as "message" is already given by the first art "msg". There a few more cases below

l3kernel/l3msg.dtx
```diff
@@ -1879,30 +1917,27 @@
     \iow_indent:n { #1 .<property> } \\ \\
     LaTeX~did~not~find~a~'.'~to~indicate~the~start~of~a~property.
   }
-\__kernel_msg_new:nnnn { kernel } { key-property-boolean-values-only }
+\__kernel_msg_new:nnnn { keys } { key-property-boolean-values-only }
```
FrankMittelbach, line 1920: again, "key-" unnecessary? (and below)

l3kernel/l3msg.dtx
```diff
@@ -1912,16 +1947,16 @@
       { with~signature~'#2',~but~that~signature~ }
     is~not~valid.
   }
-\__kernel_msg_new:nnn { kernel } { invalid-quark }
+\__kernel_msg_new:nnn { quark } { invalid-quark }
   { Invalid~quark~variable~'#1'. }
-\__kernel_msg_new:nnnn { kernel } { scanmark-already-defined }
+\__kernel_msg_new:nnnn { scanmark } { scanmark-already-defined }
```
FrankMittelbach, line 1952: again could be shorter I guess

l3kernel/l3prop.dtx
```diff
@@ -602,7 +602,7 @@
 %   error and removes its argument.
 %    \begin{macrocode}
 \cs_new:Npn \@@_pair:wn #1 \s_@@ #2
-  { \__kernel_msg_expandable_error:nn { kernel } { misused-prop } }
+  { \__kernel_msg_expandable_error:nn { prop } { misused-prop } }
```
FrankMittelbach, line 605: "misused" ?

l3kernel/l3prop.dtx
```diff
@@ -1359,7 +1359,7 @@
   {
     \__kernel_chk_defined:NT #2
       {
-        #1 { LaTeX/kernel } { show-prop }
+        #1 { LaTeX / prop } { show }
```
FrankMittelbach, line 1362: ???

l3kernel/l3quark.dtx
```diff
@@ -641,7 +641,7 @@
 \cs_new_protected:Npn \@@_new_test_aux:Nn #1 #2
   {
     \if_meaning:w \q_nil #2 \q_nil
-      \__kernel_msg_error:nnx { kernel } { invalid-quark-function }
+      \__kernel_msg_error:nnx { quark } { invalid-quark-function }
         { \token_to_str:N #1 }
```
FrankMittelbach, line 645: drop -quark- here and below?

l3kernel/l3quark.dtx
```diff
@@ -961,7 +961,7 @@
   {
     \tl_if_in:NnTF \g_@@_marks_tl { #1 }
       {
-        \__kernel_msg_error:nnx { kernel } { scanmark-already-defined }
+        \__kernel_msg_error:nnx { scanmark } { scanmark-already-defined }
```
FrankMittelbach, line 964: I think that "scanmark" got dropped elsewhere

l3kernel/l3seq.dtx
```diff
@@ -2335,7 +2335,7 @@
   {
     \__kernel_chk_defined:NT #2
       {
-        #1 { LaTeX/kernel } { show-seq }
+        #1 { LaTeX / seq } { show }
```
FrankMittelbach, line 2338: ???

l3kernel/l3seq.dtx
```diff
@@ -1054,7 +1054,7 @@
 %    \begin{macrocode}
 \cs_new:Npn \@@_item:n
   {
-    \__kernel_msg_expandable_error:nn { kernel } { misused-sequence }
+    \__kernel_msg_expandable_error:nn { seq } { misused-sequence }
```
FrankMittelbach, line 1057: drop -sequence?

l3kernel/l3tl-analysis.dtx
```diff
@@ -990,15 +990,15 @@
     \tl_if_exist:NTF #1
       {
         \exp_args:No \@@_analysis:n {#1}
-        \msg_show:nnxxxx { LaTeX / kernel } { show-tl-analysis }
+        \msg_show:nnxxxx { LaTeX / tl } { show-tl-analysis }
```
FrankMittelbach, line 993: ???

l3kernel/l3tl-analysis.dtx
```diff
@@ -990,15 +990,15 @@
...
       { \tl_show:N #1 }
   }
 \cs_new_protected:Npn \tl_analysis_show:n #1
   {
     \@@_analysis:n {#1}
-    \msg_show:nnxxxx { LaTeX / kernel } { show-tl-analysis }
+    \msg_show:nnxxxx { LaTeX / tl } { show-tl-analysis }
       { } { \@@_analysis_show: } { } { }
   }
 %    \end{macrocode}
```
FrankMittelbach, line 1004: why LaTeX/tl and kernel below?

l3kernel/l3tl-analysis.dtx
```diff
@@ -1482,7 +1482,7 @@
 % \end{variable}
 %
 %    \begin{macrocode}
-\__kernel_msg_new:nnn { kernel } { show-tl-analysis }
+\__kernel_msg_new:nnn { tl } { show-tl-analysis }
```
FrankMittelbach, line 1485: see above

l3trial/l3fp-extras/l3fp-symbolic.dtx
```diff
@@ -153,7 +153,7 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_symbolic_chk:w #1,#2#3;
   {
-    \__kernel_msg_error:nnx { kernel } { misused-fp }
+    \__kernel_msg_error:nnx { fp } { misused-fp }
```
FrankMittelbach, line 156: just "misused" ?

l3trial/l3kernel-extras/l3kernel-extras.dtx
```diff
@@ -1026,7 +1026,7 @@
       \prg_return_false:
     \fi:
   }
-\__kernel_msg_new:nnn { kernel } { show-char }
+\__kernel_msg_new:nnn { char } { show-char }
```
FrankMittelbach, line 1029: but here as kernel

l3kernel/l3regex.dtx
```diff
@@ -5814,7 +5814,7 @@
   {
     \@@_compile:n {#1}
     \@@_show:N \l_@@_internal_regex
-    \msg_show:nnxxxx { LaTeX / kernel } { show-regex }
+    \msg_show:nnxxxx { LaTeX / regex } { show }
```
FrankMittelbach, line 5817: ???

l3kernel/l3clist.dtx
```diff
@@ -2037,7 +2037,7 @@
 \cs_new_protected:Npn \clist_log:n { \@@_show:Nn \msg_log:nnxxxx }
 \cs_new_protected:Npn \@@_show:Nn #1#2
   {
-    #1 { LaTeX/kernel } { show-clist }
+    #1 { LaTeX / clist } { show }
```
FrankMittelbach, line 2040: why LaTeX / clist here?

FrankMittelbach (commented):
bunch of queries and suggestions

FrankMittelbach:
> This is step one of looking at removing/reworking __kernel_msg_ support.

My feeling remains that we should keep `\__kernel`  as the command even if internal it expans to `\msg ...{LaTeX / #1}`.
For one thing, while the messages might get redirected, no kernel code should not use them to generate package errors and warning so keeping it `\__` seems appropriate instead of showing `{ LaTeX/ xxx}`which gives the message that this is the way to do it elsewhere too (which it isn't).

l3kernel/l3coffins.dtx
josephwright: These ones were I think done by @blefloch or @PhelypeOleinik, not me. I'll tidy them up so we are self-consistent.

l3kernel/l3file.dtx
josephwright: I was tidying up the formatting to make search-and-replace easier. I've changed here to just `{ kernel }`  with `#1` here including the `__kernel_` prefix.

l3kernel/l3coffins.dtx
josephwright: Turned out we didn't have `\__kernel_mg_show:nnnn`, which I'll sort first with a change on master then with one here.

l3kernel/l3msg.dtx
josephwright: Changed

l3kernel/l3msg.dtx
josephwright: Shortened where possible

l3kernel/l3intarray.dtx
josephwright: I think the two views seem to be at odds here ...

PhelypeOleinik: My suggestion is just to make browsing the code easier, so feel free to ignore

FrankMittelbach: I think your comment is worth thinking about, but on the whole I think the shorter versions are cleaner and faster and if used well then searching for them shouldn't be (much) more complicated - though I must confess I'm not sure why
`{ foo } { bar }` in these places should be better than `{foo} {bar}`   --- but I'm fine with either convention, we should just stay consistent I think.

FrankMittelbach: > I think the two views seem to be at odds here ...

a bit :-).   Mine was triggered from what was done in the past elsewhere. Also I really think we should consider making use of those arguments in the error message in the future, e.g.
```
   LaTeX (#1) error:
```
as a  minimum, perhaps also integrating `#2` somehow and then, not doubling the info  makes sense.

l3kernel/l3file.dtx
FrankMittelbach: my comment was not about the space but about not using \__kernel here :-)

josephwright: As I said elsewhere, there wasn't a `\__kernel_msg_` function for the `show` operation: I've fixed that.

l3kernel/l3intarray.dtx
PhelypeOleinik: > I think your comment is worth thinking about, but on the whole I think the shorter versions are cleaner and faster

Both true, but the speed improvement is likely negligible compared to the rest of the `msg` code.

> though I must confess I'm not sure why `{ foo } { bar }` in these places should be better than `{foo} {bar}`

Better I don't know, but that's what `l3styleguide` says:

> With the exception of simple runs of parameter (`{#1}`, `#1#2`, etc.), everything should be divided up using spaces to make the code more readable.

and that's the convention used throughout `l3kernel`, so it should be (for consistency) with spaces (and to ensure that doing ``\catcode`\ =9`` was worth it :-)

FrankMittelbach: > > I think your comment is worth thinking about, but on the whole I think the shorter versions are cleaner and faster
> 
> Both true, but the speed improvement is likely negligible compared to the rest of the `msg` code.

I don't care about speed when an error occurs :-) I  was hthinking about bypassing many additional tokens  when no error occurs, it is also minimal but unnecessary

> Better I don't know, but that's what `l3styleguide` says:

does it? must be ok in that case :-)

> and that's the convention used throughout `l3kernel`, so it should be (for consistency) with spaces (and to ensure that doing `` \catcode`\ =9 `` was worth it :-)

that was worth it in any case

josephwright: > ```
>    LaTeX (#1) error:
> ```
> 
> as a minimum, perhaps also integrating `#2` somehow and then, not doubling the info makes sense.

I'll work one that.

You'll remember that we used to show `#2` but it was felt to be too wordy. At some stage I do wonder if we want to have an option to show the message details, either in parallel with or as an alternative to the human-readable text. At some stage if we do that we should document the messages (including how many arguments they take).

l3kernel/l3tl-analysis.dtx
josephwright: Back with the `show` business, which is fixed

josephwright:
> > This is step one of looking at removing/reworking __kernel_msg_ support.
> 
> My feeling remains that we should keep `\__kernel` as the command even if internal it expans to `\msg ...{LaTeX / #1}`.
> For one thing, while the messages might get redirected, no kernel code should not use them to generate package errors and warning so keeping it `\__` seems appropriate instead of showing `{ LaTeX/ xxx}`which gives the message that this is the way to do it elsewhere too (which it isn't).

As I've indicated, we could just use `\msg_error:nn(nnnn)`, etc.; it's only a question of adding the modules to the 'mapping' data

FrankMittelbach:
> > > This is step one of looking at removing/reworking __kernel_msg_ support.
> > 
> > 
> > My feeling remains that we should keep `\__kernel` as the command even if internal it expans to `\msg ...{LaTeX / #1}`.
> > For one thing, while the messages might get redirected, no kernel code should not use them to generate package errors and warning so keeping it `\__` seems appropriate instead of showing `{ LaTeX/ xxx}`which gives the message that this is the way to do it elsewhere too (which it isn't).
> 
> As I've indicated, we could just use `\msg_error:nn(nnnn)`, etc.; it's only a question of adding the modules to the 'mapping' data

right we could I understand that. my point is that then people look at the code and see that, they are likely to  copy it (including "LaTeX / ...")  change "..." and think they do the right thing.  That's why I prefer `\__kernel...` (and also because it is shorter and less tokens to pass over :-)

PhelypeOleinik:
I prefer `\__kernel_...` too. Makes it easier later to change behaviour if we want without hunting down every occurrence in the code

josephwright:
Does this look OK? Further comment?

PhelypeOleinik (approved):

FrankMittelbach:
looks fine to me, but I wonder if it is really sensible to push that now in front of the freeze as this does affect the main LaTeX kernel and it is just an internal adjustment that may be better delayed and not show up on the DVD (just in case there is a typo or some mistake somewhere that is not yet caught). Is there any reason why it needs pushing now (other than getting it out?

josephwright:
@FrankMittelbach  No rush, but I was not expecting to release for this, just to merge into master - I don't think we will do an upload for a while.

FrankMittelbach:
Am 24.02.21 um 18:38 schrieb Joseph Wright:
> @FrankMittelbach <https://github.com/FrankMittelbach> No rush, but I was 
> not expecting to release for this, just to merge into master - I don't 
> think we will do an upload for a while.

then fine ... code looks good as far as I can tell.

FrankMittelbach:
Am 18.02.21 um 09:39 schrieb Joseph Wright:
> *@josephwright* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In l3kernel/l3intarray.dtx 
> <https://github.com/latex3/latex3/pull/854#discussion_r578226394>:
> 
>>        {
> -        #1 { LaTeX/kernel } { show-intarray }
> +        #1 { LaTeX / intarray } { show }
>             { \token_to_str:N #2 }
> 
>     |LaTeX (#1) error: |
> 
>     as a minimum, perhaps also integrating |#2| somehow and then, not
>     doubling the info makes sense.
> 
> I'll work one that.
> 
> You'll remember that we used to show |#2| but it was felt to be too 
> wordy. At some stage I do wonder if we want to have an option to show 
> the message details, either in parallel with or as an alternative to the 
> human-readable text. At some stage if we do that we should document the 
> messages (including how many arguments they take).

yes I remember that. And I didn't mean to reintroduce what we did 
initially, but it might make sense to make use of this info somehow so 
that one can more easily follow an error back (nothing urgent)

But I think the "LaTeX (module) error:" makes some sense and can be helpful.

josephwright:
@FrankMittelbach  There's no hurry: I mainly wanted to know where we were up to. I'll leave this one 'parked' until after the TL'21 DVD is cut.

PR #855, opened 2021-02-21
Add ICC-based color profile support

josephwright:
At present only PDF-based routes work; I'm not sure at all how to handle PostScript, and can't work out if SVG supports this concept either.

### DIFF
--- a/l3kernel/l3color.dtx
+++ b/l3kernel/l3color.dtx
@@ -2107,9 +2107,9 @@
     \@@_model_devicen_convert:nnn {#4} {#2} {#3}
   }
 %    \end{macrocode}
-%   For short lists of DeviceN colors, wee can use hand-tuned parsing. This
+%   For short lists of DeviceN colors, we can use hand-tuned parsing. This
 %   lines up with other models, where we allow for up to four components. For
-%   larger spaces,. rather than limit artificially, we use a somewhat slow
+%   larger spaces, rather than limit artificially, we use a somewhat slow
 %   approach based on open-ended commas-lists.
 %    \begin{macrocode}
 \cs_new_protected:cpn { @@_model_devicen_parse_1:nn } #1#2
--- a/l3kernel/l3color.dtx
+++ b/l3kernel/l3color.dtx
@@ -388,6 +388,7 @@
 %   The latter should be one of
 %   \begin{itemize}
 %     \item \texttt{DeviceN}
+%     \item \texttt{ICCBased}
 %     \item \texttt{Separation}
 %   \end{itemize}
 %   (The \meta{family} may be given in mixed case as-in the PDF reference:
@@ -429,6 +430,13 @@
 %   \item \texttt{names} The names of the components of the \texttt{DeviceN}
 %   space. Each should be either the \meta{name} of a \texttt{Separation} model,
 %   a process color name (\texttt{cyan}, etc.) or the special name \texttt{none}.
+% \end{itemize}
+%
+% For a \texttt{ICCBased} space, there are two \emph{compulsory} key.
+% \begin{itemize}
+%   \item \texttt{file} The name of the file containing the profile.
+%   \item \texttt{components} The number of components the profile requires:
+%     must be one of $1$, $3$ or $4$.
 % \end{itemize}
 %
 % \end{documentation}
@@ -2485,6 +2493,69 @@
 % \end{macro}
 % \end{macro}
 %
+% \begin{macro}{\@@_model_iccbased:n}
+% \begin{macro}{\@@_model_iccbased:nn}
+% \begin{macro}{\@@_model_iccbased:nnn, \@@_model_iccbased_aux:nnn}
+%   For an ICC profile, we need a file name and a number of components. The
+%   file name is processed here so the backend can treat it as a string.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_model_iccbased:n #1
+  {
+    \prop_get:NnNTF \l_@@_internal_prop { file }
+      \l_@@_internal_tl
+      {
+        \exp_args:NV \@@_model_iccbased:nn
+          \l_@@_internal_tl {#1}
+      }
+      {
+        \__kernel_msg_error:nnn { color }
+          { ICCBased-requires-file } {#1}
+      }
+  }
+\cs_new_protected:Npn \@@_model_iccbased:nn #1#2
+  {
+    \prop_get:NnNTF \l_@@_internal_prop { components }
+      \l_@@_internal_tl
+      {
+        \exp_args:NV \@@_model_iccbased:nnn
+          \l_@@_internal_tl {#2} {#1}
+      }
+      {
+        \__kernel_msg_error:nnn { color }
+          { ICCBased-requires-components } {#2}
+      }
+  }
+\cs_new_protected:Npn \@@_model_iccbased:nnn #1#2#3
+  {
+    \bool_lazy_any:nTF
+      {
+        { \int_compare_p:nNn {#1} = 1 }
+        { \int_compare_p:nNn {#1} = 3 }
+        { \int_compare_p:nNn {#1} = 4 }
+      }
+      { \@@_model_iccbased_aux:nnn {#1} {#2} {#3} }
+      {
+        \__kernel_msg_error:nnn { color }
+          { ICCBased-component-number } {#2}
+      }
+  }
+%    \end{macrocode}
+%   Here, we can use the same internals as for DeviceN approach as we know the
+%   number of components. No conversion is possible, so there is no need
+%   to worry about that at all.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_model_iccbased_aux:nnn #1#2#3
+  {
+    \@@_model_init:nnn {#2} {#1} { iccbased }
+    \use:c { @@_model_devicen_parse_ #1 :nn } {#2} {#1}
+    \exp_args:Nx \@@_backend_devicen_init:nnn
+      { \file_full_name:n {#3} } {#1} {#2}
+  }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+% \end{macro}
+%
 % \subsection{Diagnostics}
 %
 % \begin{macro}{\color_show:n}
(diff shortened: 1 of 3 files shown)


car222222:
Not sure exactly what you mean here by 'SVG supports . . .'

The latest drafts for SVG 2 indicate that everything in this area now refers to CSS.

But  maybe you are asking about what 'our SVG producer' supports? 

I also found this older (now obsolete) statement:

SVG 2 content may specify color using an ICC profile (see [ICC42]); an sRGB fallback must still be provided.

In: https://www.w3.org/TR/2013/WD-SVG2-20130409/color.html

And a pseudo-example:
<color-profile name="acmecmyk" 
   href="http://printers.example.com/acmecorp/model1234"/>

<circle fill="#CD853F 
   icc-color(acmecmyk, 0.11, 0.48, 0.83, 0.00)"/>

Note that sRGB is itself an ICC profile, and it is ubiquitous in SVG 2.

josephwright:
@car222222 Well I found e.g. https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/color-profile, which says `color-profile` is deprecated.  I see the link you found suggests some CSS alternative, but I've no idea how one arranges that: we can only add to the SVG file itself.

davidcarlisle:
@josephwright  I think that you should be able to use css in an svg `<style>` element which works like the same named element in html, more or less.

josephwright:
OK, so profiles are doable in SVG but they want some ID rather than a file name: I'll have to work out how you know what that is!

(Aside: there is also mention of CIELAB, which presumably should therefore be workable for separations, but not _as_ separation: I suppose I need to think about how that splits out behind the scenes.)

car222222:
Thanks, David.  I did not go into the details but yes, in general, SVG and HTML+CSS are highly integrated in this kind of way, forming a single hierarchy.

josephwright:
Looking at the wider concept, I suspect this is not quite the right approach. The existing `colorspace` package maps /DeviceRGB to the /ICCbased, and that has seems to fit much better with SVG. I think when you start working through what you can do with ICC spaces, they look more sensible as modifying existing colours: avoids needing to redefine everything.

u-fischer:
Well it works now, and imho it is one allowed and documented way to set it up. Perhaps you could add an key to switch between this "standalone" and the mapping to DeviceRGB variant (if the second fits better to SVG it should probably the default). 

Whatever you do: don't change the object name ` @@_icc_ #1` , I just changed l3pdfmeta to use this too, so that the profile is embedded only once ...

josephwright:
> Well it works now, and imho it is one allowed and documented way to set it up. Perhaps you could add an key to switch between this "standalone" and the mapping to DeviceRGB variant (if the second fits better to SVG it should probably the default).

I guess looking back my concern is that if you make a separate colour space, you have to worry about standard colours like `red`, `green`, `blue`, etc., whereas if you map the device colour spaces to a profiled one, that's not an issue. 

> Whatever you do: don't change the object name ` @@_icc_ #1` , I just changed l3pdfmeta to use this too, so that the profile is embedded only once ...

That will definitely stay: it matches the pattern for other colour-related objects.

josephwright:
I'm going to sort a parallel PR to address the 'just apply the profile' approach: we can then consider pros and cons.

car222222:
"they want some ID rather than a file name: I'll have to work out how you know what that is!"

Not sure who "they" are? :-)

Do you mean by 'some ID' the "acmecmyk" here?  

icc-color(acmecmyk, 0.11, 0.48, 0.83, 0.00)

This is just a 'local name' that is linked to a file by a  declaration in a 'style element', for example:

<style>
@color-profile {
  name: acmecmyk;
  src: url(http://printers.example.com/acmecorp/model1234);
}
</style>

they want some ID rather than a file name: I'll have to work out how you know what that is!

But there may also be some pre-defined icc profile names, such as "sRGB".

car222222:
"The existing colorspace package maps /DeviceRGB to 
the /ICCbased, . . . "

Maybe it does, but that package is concerned only with the following:

• ICC based (or “tagged”) default CMYK, RGB and Gray spaces.

and not with icc profiles in general.

". . . that has seems to fit much better with SVG."

Well, SVG is pretty much only concerned with RGB and the icc profile 'sRGB', so that fit may well be better for /DeviceRGB.

"I think when you start working through what you can do with ICC spaces, they look more sensible as modifying existing colours: avoids needing to redefine everything."

Not quite sure what you mean by 'modifying' here.
And what are these 'existing colors'?  Are they just colors that already got a name somehow?

josephwright:
> "they want some ID rather than a file name: I'll have to work out how you know what that is!"
> 
> Not sure who "they" are? :-)

https://www.w3.org/TR/SVG11/color.html#ColorProfileAtRule

> Do you mean by 'some ID' the "acmecmyk" here?

No idea! It says: '`<string>` is the profile's unique ID as specified by International Color Consortium'. It also mentions referencing by IRI (~URL), but I don't see anything about a file name.

> This is just a 'local name' that is linked to a file by a declaration in a 'style element', for example:
> <style> @color-profile { name: acmecmyk; src: url(http://printers.example.com/acmecorp/model1234); } </style>

Maybe, but that doesn't address the way we will be presented with profile files: just as a local file.

> But there may also be some pre-defined icc profile names, such as "sRGB".

sRGB is pre-defined certainly.

josephwright:
> "The existing colorspace package maps /DeviceRGB to
> the /ICCbased, . . . "
> 
> Maybe it does, but that package is concerned only with the following:
> 
> • ICC based (or “tagged”) default CMYK, RGB and Gray spaces.
> 
> and not with icc profiles in general.

Very true, but if you read about ICC profiles they can only have 1, 3 or 4 components and fall back on Gray/RGB/CMYK automatically. So I'm not sure they are actually useful for say a DeviceN case of 2 components.

> Well, SVG is pretty much only concerned with RGB and the icc profile 'sRGB', so that fit may well be better for /DeviceRGB.

I think SVG does cover other stuff: as I mentioned, Lab is in the docs for example.

> "I think when you start working through what you can do with ICC spaces, they look more sensible as modifying existing colours: avoids needing to redefine everything."
> 
> Not quite sure what you mean by 'modifying' here.
> And what are these 'existing colors'? Are they just colors that already got a name somehow?

I mean we have for example
```latex
\color_set:nnn { blue }  { rgb } { 0 , 0 , 1 }
```
as a pre-defined ('existing') colour. If we add an ICC profile as a new colour model, it doesn't affect any of those colours. So `blue` will still use un-profiled DeviceRGB. To apply the profile to these colours, you'd therefore need to re-declare the colour
```latex
\color_set:nnn { blue }  { sRGB } { 0 , 0 , 1 }
```
That may be appropriate in some cases, but I suspect the most common use case is to want to apply a profile and have it 'just work'. That could be done using the 'fixed model' code already in `l3color`
```latex
\tl_set:Nn \l_color_fixed_model_tl { sRGB }
```
or by setting the default for (_e.g._) DeviceRGB in the PDF structure. The latter approach is what `colorspace` does and avoids having to do conversions at the TeX side for every colour use. It also means you can be more selective: it only applies to _e.g._ DeviceRGB, whereas setting a fixed model applies to _all_ colour that goes through `l3color`.

I'm going to put in a second PR that shows the alternative approach. That will not preclude the plan here: the 'behind the scenes' data loading is the same. The idea of putting both ideas in is to see which, if either, we want _now_ (@u-fischer wants ICC profiles for PDF/A support) and which, if any, we might want to put in later.

To be clear, I'd like to support as many PDF concepts as possible, so I don't see a major downside having both approaches.

zauguin:
> No idea! It says: '`<string>` is the profile's unique ID as specified by International Color Consortium'. It also mentions referencing by IRI (~URL), but I don't see anything about a file name.

If you have a ICC profile in a file, you can get the major version and "unique ID" using

```
\documentclass{article}
\begin{document}
\ExplSyntaxOn
\typeout{Major~version:~\file_hex_dump:nnn{./sRGB_v4_ICC_preference.icc}{9}{9}}
\typeout{Unique~ID:~\exp_args:Nf\str_foldcase:n{\file_hex_dump:nnn{./sRGB_v4_ICC_preference.icc}{85}{100}}}
\ExplSyntaxOff
\end{document}
```
We probably have to verify that the major version is "04", otherwise the unique ID field doesn't exists. If the unique ID only contains zeros, then it hasn't been set.

But I guess we could also embed the profile using data: URL's, but then we need some base64 encoding support first.

josephwright:
> sRGB_v4_ICC_preference

That will do as a starting point for SVG then :)

u-fischer:
@zauguin I hadn't thought about  the possibility to get data from the .icc directly.  How reliable is this? Which data is always there? I need for the OutputIntent dictionary imho the fields
OutputConditionIdentifier (e.g IEC sRGB), the Info (IEC 61966 2.1 Default RGB colour space - sRGB), the 
RegistryName (http://www.iec.ch) and the Number (3)

zauguin:
> @zauguin I hadn't thought about the possibility to get data from the .icc directly. How reliable is this? Which data is always there? I need for the OutputIntent dictionary imho the fields
> OutputConditionIdentifier (e.g IEC sRGB), the Info (IEC 61966 2.1 Default RGB colour space - sRGB), the
> RegistryName (http://www.iec.ch) and the Number (3)

The output condition information is not included as far as I can tell. There is a profile name, but it's referring to the profile and not the output condition. I guess the 3 refers to the number of components? That's included indirectly: Basically there is a 4 byte color space signature with 26 possible values. For ICCBased colors only four colorspace signatures are valid "GRAY", "RGB ", "CMYK" and "Lab ", so we could test for RGB or Lab and map these to "3" (I don't know if that condition holds for the OutputIntent, but I don't think that we need the number of components there anyway. )

josephwright:
Looking at the SVG/CSS docs, we probably can use a simple relative file name. I've tried that in https://github.com/latex3/latex3/pull/856. I also notice that SVG should allow colour selection by model, so the approach *here* does look viable, although the issue is whether that will work with the color stack.

josephwright:
I'd welcome thoughts on both this PR and https://github.com/latex3/latex3/pull/856 -  the latter is simpler but probably addresses 99% of the use cases, this one is more flexible but I think doesn't immediately cover e.g. PDF/A requirements

l3kernel/l3color.dtx
```diff
@@ -2562,6 +2628,30 @@
...
+    LaTeX~has~been~asked~to~create~an~ICCBased~color~space~with~#1~components,~
+    but~only~1,~3~or~4~components~are~allowed.
+  }
+\__kernel_msg_new:nnnn { color } { ICCBased-requires-components }
+  { ICCBased~color~space~'#1'~require~a~number~of~components. }
+  {
+    LaTeX~has~been~asked~to~create~an~ICCBased~color~space,~but~no~\\ \\
+    \iow_indent:n { compontents~=~<number> }
+    \\ \\
+    key~was~given~with~the~correct~information.~LaTeX~will~use~ignore~this~
```
zauguin, line 2643: s/use/just/ ?

josephwright:
@zauguin suggests automatically getting data out of the profile files: sounds good to me. So if we go with this PR I'd expect to integrate his changes. (I feel we need this and my other PR, as they do slightly different things.)

PR #856, opened 2021-02-23
Support ICC profiles for device colour spaces

josephwright:
Parallel to #855, this approach only covers modifying the 'matching' device
spaces.

### DIFF
--- a/l3backend/l3backend-color.dtx
+++ b/l3backend/l3backend-color.dtx
@@ -913,6 +913,37 @@
 % \end{macro}
 % \end{macro}
 %
+% \begin{macro}{\@@_backend_iccbased_device:nnn}
+%   This is very similar to setting up a color space: the only part we
+%   skip is adding it to the page resources.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_backend_iccbased_device:nnn #1#2#3
+  {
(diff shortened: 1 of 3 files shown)


l3backend/l3backend-color.dtx
```diff
@@ -913,6 +913,37 @@
...
+      }
+    \pdf_object_unnamed_write:nx { array }
+      { /ICCBased ~ \pdf_object_ref:n { @@_icc_ #1 } }
+    \cs_if_exist:NT \pdfmanagement_add:nnn
+      {
+        \use:x
+          {
+            \pdfmanagement_add:nnn
+              { Page / Resources / ColorSpace }
+              { Default #2 }
```
u-fischer, line 963: That should be Device not Default

josephwright: The idea is to parallel `colorspace`:
```
/DefaultRGB 0 <X> R
```
where `<X>` is the object ref. for the ICC colour space. Have I got it in the wrong place?

u-fischer: I'm not sure ;-) I thought it is Device, but perhaps I mixed it up. I need to check,

u-fischer: You are right, it is Default.

PR #864, opened 2021-03-17
Add "pdfcolorstack n current" to (x)dvipdfmx scopes

josephwright:
This mirrors https://github.com/latex3/graphics-def/pull/27


josephwright:
Not going ahead at this stage: revisit later.

PR #877, opened 2021-04-24
Change catcode produced by default by l3regex replacement (fixes #621)

### LINKED ISSUE #621: Regex replacement turns letter tokens into other tokens
From [this](https://tex.stackexchange.com/a/506624/134574) TeX.SX thread. The code:
```latex
\input expl3-generic
\ExplSyntaxOn
\tl_set:Nn \l_tmpa_tl { a }
\regex_replace_once:nnN { a } { \0 b } \l_tmpa_tl
\tl_analysis_show:N \l_tmpa_tl
\use:c { stop }
\bye
```
shows this in the terminal:
```none
The token list \l_tmpa_tl contains the tokens:
>  a (the letter a)
>  b (the character b).
```
with the `b` becoming a catcode 12 token (I believe due to the `\detokenize` applied to the regular expression before processing).

This makes, in the linked thread, `\regexpatchcmd` fail because when retokenized, `b` will have catcode 11 and `\regexpatchcmd` thinks (correctly) that the function was defined under a different catcode regime.

blefloch:
The code now uses the current values of `\catcode` to decide what token to produce.  E.g.,

    \regex_replace_all:nnN { .* } { $a^{b+1}$ } \l_tmpa_tl
    \tl_analysis_show:N \l_tmpa_tl

will now show the catcodes one would see upon rescanning `$a^{b+1}$` with the current catcodes.  This is a breaking change, so I'd like some opinions.  No need to look at the code, I think, but I'd like to know if the doc is clear.

### DIFF
--- a/l3kernel/testfiles/m3regex005.luatex.tlg
+++ b/l3kernel/testfiles/m3regex005.luatex.tlg
@@ -180,9 +180,9 @@ The token list \l_tmpa_tl contains the tokens:
 <recently read> }
 l. ...  }
 The token list \l_tmpa_tl contains the tokens:
->  x (the character x)
+>  x (the letter x)
 >  \c_parameter_token (control sequence=macro parameter character #)
->  x (the character x).
+>  x (the letter x).
 <recently read> }
 l. ...  }
 ============================================================
@@ -281,6 +281,11 @@ TEST 10: Caseless matching and cs
 TEST 11: Braces
 ============================================================
 |\{}|
+The token list \l_tmpa_tl contains the tokens:
+>  \{ (control sequence=\protected macro:->\ifmmode \lbrace \else \textb\ETC.)
+>  } (the letter }).
+<recently read> }
+l. ...  }
 ! LaTeX3 Error: Missing right brace inserted in replacement text.
 For immediate help type H <return>.
  ...                                              
@@ -297,8 +302,8 @@ TEST 12: More tests of cs
 TEST 13: Replaced space catcode
 ============================================================
 blank space  
-the character  
-the character  
+blank space  
+blank space  
 ============================================================
 ============================================================
 TEST 14: Catcode group in replacement
@@ -312,7 +317,7 @@ In a replacement text, the '\c' escape sequence can be followed by one of the
 letters 'ABCDELMOPSTU' representing the character category. Then, a character
 must follow, not '\1'.
 The token list \l_tmpa_tl contains the tokens:
->  q (the character q)
+>  q (the letter q)
 >  e (subscript character e)
 >  t (superscript character t)
 >  a (the letter a)
@@ -321,8 +326,8 @@ The token list \l_tmpa_tl contains the tokens:
 >  1 (math shift character 1)
 >  p (superscript character p)
 >  s (subscript character s)
->  f (the character f)
->  q (the character q)
+>  f (the letter f)
+>  q (the letter q)
 >  e (subscript character e)
 >  t (superscript character t)
 >  a (the character a)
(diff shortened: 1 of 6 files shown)


l3kernel/testfiles/m3regex005.luatex.tlg
```diff
@@ -281,6 +281,11 @@ TEST 10: Caseless matching and cs
 TEST 11: Braces
 ============================================================
 |\{}|
+The token list \l_tmpa_tl contains the tokens:
+>  \{ (control sequence=\protected macro:->\ifmmode \lbrace \else \textb\ETC.)
+>  } (the letter }).
```
FrankMittelbach, line 286: caveat: comment is just based on reading the diffs not the whole code, but why is this `letter`?  if that test is a verbatim context I would have expected `the character }`  otherwise `begingroup character`

FrankMittelbach (approved):

l3kernel/testfiles/m3regex005.luatex.tlg
blefloch: Previously, the test did a regex replacement with replacement text `\c{{}}`.  Since `\c{...}` does not attempt to keep track of brace nesting (do we want to change that?), this produces `\{` and `}`.  With the new changes, `}` would have catcode 2 and the result would be unbalanced.  So I had to change the test somehow, and I opted for the stupid replacement text `\c{{}\cL}`, which makes the closing `}` of catcode letter.

Do we want `\c{...}` to keep track of left/right brace matching, i.e. do we want `\c{ a { } b }` to give a single control sequence?

FrankMittelbach: Don't feel competent to have a serious opinion here, my point was simply that I didn't expect to see "letter" if it has to be rendered harmless I would have expected "character", but  without considering implications I would think that it seems natural that `\c{ ... }`resulting in a single control sequence not a cs followed by some remainder. So either, yes, it should produce `\a{}b` as a single cs or it should error or the documentation for "replacement" would need ot say more than `\c{ 〈 cs name 〉 } inserts a control sequence;`.

FrankMittelbach: Guess I could test but what would `\c{ _ _ foo _ bar : } produce? A single cs?

blefloch: Yes, `\c{__foo_bar:}` gives a single cs, the only issue is when one tries to have braces inside `\c` because `{` are treated like any other characters while `}` end the csname, so `\c{ { } ... }` gives `\{` followed by some characters instead of giving a single control sequence.

FrankMittelbach: ok understood, but having braces in csnames is really weird anyway (other than in `\}` and `\{` so maybe it doesn't really matter either way as long as it is documented. Or you could have some very strange way of generating them explicitly, e.g. `\c{{}} -> \{` and `\c{}{} -> \}` .  Somewhat nicer would be `\c{{}` and `\c{}}` but that would give you unbalanced braces on the outside so not really worth it.

Skillmon: Wouldn't one use `\c{\{something\}}` to match a control sequence containing braces in its name?

blefloch: @Skillmon Yes, which is why the question of what `\c{{}}` does is an unimportant implementation detail.  I'll tweak things by forbidding non-escaped `{` inside `\c{...}` in the next commit.

Issue #590, opened 2019-06-12
Composing regular expressions in l3regex


PhelypeOleinik:
A couple of days ago a question was asked at TeX.SX about composing regexes: https://tex.stackexchange.com/q/494830/134574. In the post, OP wanted an escape sequence similar to `\u{<tl var>}` which, instead of using the literal tokens contained in `\<tl var>`, would parse them as a regular expression. I thought about two ways of doing that:

 1. Having the `\<tl var>` be a compiled regex and inserting in the current regex. The code for this one could be very similar to `\u`, changing only the final step, where for `\u` the `\<tl var>` is read "verbatim", and now it would be read assuming that it is a `\<regex var>` (possibly with a validity verification). The downside of this method is that the `\<regex var>` being used would have to be a complete expression by itself (compiling VAR=`a-z` and `[VAR]+` would result in "any of `a`, `-`, and `z`, one or more times, greedy"). The usage would be like:
```latex
\regex_set:Nn \l_tmpa_regex { [a-z]+ } % Compile \l_tmpa_regex
\regex_set:Nn \l_tmpb_regex { \[ \y{l_tmpa_regex} \] } % Insert it with \y{...}
```

 2. Having the `\<tl var>` be an _arbitrary_ token list which is compiled as a regular expression each time it is used. This method would offer a great deal of flexibility, as there would be no bounds on the contents of the `\<tl var>` as long as, at the very end, the whole thing would make a valid regular expression. However the implementation for this one would be problematic because at the time the `\<tl var>` was read in, the engine would be at the second compilation step and would need to return to the first to scan the `\<tl var>`. If this method were to be used, then perhaps a pre-scanning of the regular expression to expand the `\<tl var>` would be easier (but would be very different from the other steps, and I don't know how problematic that would be). The usage would be like:
```latex
\tl_set:Nn \l_tmpa_tl { [a-z]+ }
\regex_set:Nn \l_tmpb_tl { \[ \y{l_tmpa_tl } \] } % Insert it with \y{...}
```
but could also be:
```latex
\tl_set:Nn \l_tmpa_tl { [a- }
\regex_set:Nn \l_tmpb_tl { \[ \y{l_tmpa_tl }z]+ \] } % Insert it with \y{...}
```

So is this feature still planned? Which of those versions would be implemented (or something else entirely)?

---

The asker also raised the issue of `\regex_show:N` being a tad hard to debug with the many lines of `char code <num>`, which I fully agree. This could easily be improved with something like:
```latex
[...]

PR #878, opened 2021-04-25
Composing regex, improving show/log and documentation (#590 and more)

blefloch:
This is not quite atomic because there were several changes suggested in issue #590.

### DIFF
--- a/l3kernel/l3regex.dtx
+++ b/l3kernel/l3regex.dtx
@@ -499,6 +499,15 @@
 % \end{verbatim}
 % results in \cs{l_my_tl} holding |first,\emph{second},first,first|.
 %
+% Regex replacement is also a convenient way to produce token lists
+% with arbitrary category codes.  For instance
+% \begin{verbatim}
+% \tl_clear:N \l_tmpa_tl
+% \regex_replace_all:nnN { } { \cU\% \cA\~ } \l_tmpa_tl
+% \end{verbatim}
+% results in \cs[no-index]{l_tmpa_tl} containing the percent character
+% with category code~$7$ (superscript) and an active tilde character.
+%
 % \section{Pre-compiling regular expressions}
 %
 % If a regular expression is to be used several times,
--- a/l3kernel/l3token.dtx
+++ b/l3kernel/l3token.dtx
@@ -146,6 +146,8 @@
 %   and other values raise an error. The \meta{charcode} may be any one valid
 %   for the engine in use.
 %   Active characters cannot be generated in older versions of \XeTeX{}.
+%   Another way to build token lists with unusual category codes is
+%   \cs{regex_replace:nnN} |{.*}| \Arg{replacement} \meta{tl~var}.
 %   \begin{texnote}
 %     Exactly two expansions are needed to produce the character.
 %   \end{texnote}


l3kernel/l3regex.dtx
```diff
@@ -3359,19 +3384,37 @@
...
   }
 %    \end{macrocode}
 % \end{macro}
+%
+% \begin{macro}{\@@_compile_ur_end:}
+%   For the |\ur{...}| construction, once we have extracted the
+%   variable's name, we simply insert the compiled regex (which the
+%   variable should be) into a non-capturing group (with no repetition)
+%   to respect the structure of regexes.  It might be possible to omit
+%   this group perhaps.
```
PhelypeOleinik, line 3406: Results may differ if the group is or is not there, for example (if I did the non-group version correctly, that is :)
```latex
\ExplSyntaxOn
\cs_gset_protected:Npn \__regex_compile_ur_end:
  {
    \tl_build_put_right:Nx \l__regex_build_tl
      {
        \exp_not:N \__regex_group_no_capture:nnnN
          { \exp_not:v { \l__regex_internal_a_tl } }
          { 1 } { 0 } \exp_not:N \c_false_bool
      }
  }

\tl_set:Nn \l_tmpa_tl { bc }
\regex_set:Nn \l_tmpa_regex { a|b }
\regex_show:n { ^\ur{l_tmpa_regex}c }
\regex_replace_all:nnN { ^\ur{l_tmpa_regex}c } { [\0] } \l_tmpa_tl
\tl_show:N \l_tmpa_tl

\cs_gset_protected:Npn \__regex_compile_ur_end:
  {
    \tl_build_put_right:Nx \l__regex_build_tl
      { \exp_not:v { \l__regex_internal_a_tl } }
  }

\tl_set:Nn \l_tmpa_tl { bc }
\regex_set:Nn \l_tmpa_regex { a|b }
\regex_show:n { ^\ur{l_tmpa_regex}c }
\regex_replace_all:nnN { ^\ur{l_tmpa_regex}c } { [\0] } \l_tmpa_tl
\tl_show:N \l_tmpa_tl
\stop
```

In the case with the group it matches `bc`, as would the regex if written in full as `^a|bc` (which seems to be correct):
```
> Compiled regex {^\ur {l_tmpa_regex}c}:
+-branch
  assertion: anchor at start (\A)
  ,-group begin (no capture)
  | char code 97 (a)
  +-branch
  | char code 98 (b)
  `-group end
  char code 99 (c).
<recently read> }
                 
l.67 \regex_show:n { ^\ur{l_tmpa_regex}c }
                                          
> \l_tmpa_tl=[bc].
<recently read> }
```

In the case without the group the `\regex_show:n` shows more or less what `\regex_show:n { ^a|bc }` would show, but it only matches `b` which seems wrong (I think?):
```latex
> Compiled regex {^\ur {l_tmpa_regex}c}:
+-branch
  assertion: anchor at start (\A)
+-branch
  char code 97 (a)
+-branch
  char code 98 (b)
  char code 99 (c).
<recently read> }
                 
l.79 \regex_show:n { ^\ur{l_tmpa_regex}c }
                                          
> \l_tmpa_tl=[b]c.
<recently read> }
```

blefloch: 1. Actually the regex `^\ur{l_tmpa_regex}c` is equivalent to `^(a|b)c` but with a non-capturing group.  This is pretty different from `^a|bc`, which is equivalent to `(^a)|(bc)` (but without capturing of course).  Here you didn't see the difference because the `bc` branch ends up being the same.

2. I think it would be confusing if `^ \ur{l_tmpa_regex} c` could match something that does not contain `c`, as would happen for the `^a|bc` interpretation.  So I think it is best to interpret and document `\ur{...}` as introducing a non-capturing group.  My point there was an implementation note: for some simple regexes it is not necessary to introduce the overhead of an actual group.  To be precise, the overhead can be avoided if the inserted regex has a single branch at top level (no "`|`" alternative).  For instance inserting `A(B|C)D` inside some other regex can be done cheaper than it currently is.

3.  The oddity you found when constructing the regex by hand is because the regex is not a valid regex.  (This motivated me just now to log issue #884: `\regex_show:N` should ideally catch that problem.)

PhelypeOleinik: Ah, so my second regex was indeed invalid!  That explains the difference then. Thanks!

PhelypeOleinik (approved):

josephwright (approved):
Looks good to me

l3kernel/l3regex.dtx
blefloch: I should add that if someone wants to get "`^`(something)`c`" to turn into `^a|bc`, then the easiest way is to do

    \tl_set:Nn \l_myreg_tl { a|b }
    \cs_generate_variant:Nn \regex_set:Nn { Nx }
    \regex_set:Nx \l_my_regex { ^ \l_myreg_tl c }

namely to use TeX's standard expansion machinery.  I'll document this momentarily.

blefloch:
I corrected various things in this branch.  The only noteworthy one maybe is that `\ur{...}` now treats any groups in the inserted regex as non-capturing groups.  Otherwise it would be hard to know what group number to expect for `b?` in `a \ur{l_foo_regex} (b?)`.

PhelypeOleinik (approved):

blefloch:
Thanks @PhelypeOleinik .  It's now merged to main.

PR #881, opened 2021-04-25
Avoid hyphenation in the first run of _ in l3doc's \cs (fixes #479)

### LINKED ISSUE #479: l3doc: do not insert hyphen when breaking a cs
In l3doc, if \cs{x_y} breaks at the _, I think it would be much (much much) better not to insert a hyphen as if it were a normal word. That looks confusing o me.

E.g., in wao, "either \cs{@@_problem_identified:n} or" breaks. and produces output:

  ... Either \__fmwao_problem_-
  identified:n or ...

But I propose just

  ... Either \__fmwao_problem_
  identified:n or ...

would be better. Not just \cs I suppose, if there are other commands
that typeset control sequences.

Viable? I'd like to have this possibility for the tugboat publication, anyway ... ?  --thanks, karl.

blefloch:
wspr had already done essentially all the work fixing this, but the regex
wasn't quite right: it didn't correctly omit hyphenation after a backslash
when it was followed by two (or more) underscores.

### DIFF
--- a/l3kernel/l3doc.dtx
+++ b/l3kernel/l3doc.dtx
@@ -1996,8 +1996,8 @@ and all files in that bundle must be distributed together.
             \bool_if:NT \g_@@_cs_break_bool
               {
                 \regex_replace_all:nnN
-                  {([^\\])_([^\_])}
-                  {\1\c{BreakableUnderscore}\2}
+                  { ([^\\\_]\_*) \_ ([^\_]) }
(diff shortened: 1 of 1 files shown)


blefloch:
I wasn't sure whether to add a CHANGELOG entry for this.

FrankMittelbach:
> I wasn't sure whether to add a CHANGELOG entry for this.

I usually think it doesn't hurt :-)

PR #891, opened 2021-04-28
Changing items at specific positions of a sequence (fixes #844)

### LINKED ISSUE #844: l3seq: provide a method to set a term at a specific index
Background: in the [hobby](https://github.com/loopspace/hobby) package there are a fair few calculations on vectors (and matrices).  In order to do the calculations, I need an array-like data structure to hold the vectors.  Specifically, I need to be able to get and set items at a specific index in an ordered list.

When I originally wrote that package, `\seq_item:Nn` didn't exist.  So I made a wrapper around a `prop` which, in essence, added the ordering part.  Now that `\seq_item:Nn` exists, a `seq` is _almost_ the data structure that I need.  What is missing is the ability to _set_ a specific term of the sequence.

Hence my request for a method to set an item at a specific index of a `seq`.

References:
* Issue on hobby repository: https://github.com/loopspace/hobby/issues/8
* A possible implementation on TeX-SX: https://tex.stackexchange.com/questions/497912/how-to-create-array-of-dims

blefloch:
- `\seq_put_item:Nnn` places an item at a certain position in the sequence (removing the item that was previously there), and errors if the position is out of bounds (zero or larger [in absolute value] than the number of items)
- `\seq_put_item:NnnTF` does not error, it returns false if the position is out of bounds
- `\seq_pop_item:NnN(TF)` does not error: it removes the item if possible and stores it in `#3`, and if the position is out of bounds it does not change the sequence and it stores `\q_no_value` in `#3`

Thoughts about the interface are welcome.  Especially what `\seq_put_item:Nnn` should do when the position is out of bounds.  Two useful comparisons: the `\seq_item:Nn` function simply expands to nothing if the position is out of bounds, and the `\prop_put:Nnn` function of course adds the key to the prop.

### DIFF
--- a/l3kernel/l3candidates.dtx
+++ b/l3kernel/l3candidates.dtx
@@ -434,6 +434,24 @@
 %   mappings.
 % \end{function}
 %
+% \begin{function}[added = 2021-04-28, noTF]
+%   {\seq_put_item:Nnn, \seq_put_item:cnn, \seq_gput_item:Nnn, \seq_gput_item:cnn}
+%   \begin{syntax}
+%     \cs{seq_put_item:Nnn} \meta{seq~var} \Arg{intexpr} \Arg{item}
+%     \cs{seq_put_item:NnnTF} \meta{seq~var} \Arg{intexpr} \Arg{item} \Arg{true code} \Arg{false code}
+%   \end{syntax}
+%   Removes the item of \meta{sequence} at the position given by
+%   evaluating the \meta{integer expression} and replaces it by
+%   \meta{item}.  Items are indexed from $1$ on the left/top of the
+%   \meta{sequence}, or from $-1$ on the right/bottom.  If the
+%   \meta{integer expression} is zero or is larger (in absolute value)
+%   than the number of items in the sequence, the \meta{sequence} is not
+%   modified.  In these cases, \cs{seq_put_item:Nnn} raises an error
+%   while \cs{seq_put_item:NnnTF} runs the \meta{false code}.  In cases
+%   where the assignment was successful, \meta{true code} is run
+%   afterwards.
+% \end{function}
+%
 % \section{Additions to \pkg{l3sys}}
 %
 % \begin{variable}[added = 2018-05-02]{\c_sys_engine_version_str}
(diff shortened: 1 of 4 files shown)


l3kernel/l3candidates.dtx
```diff
@@ -434,6 +434,42 @@
...
+%     \cs{seq_put_item:Nnn} \meta{seq~var} \Arg{intexpr} \Arg{item}
+%     \cs{seq_put_item:NnnTF} \meta{seq~var} \Arg{intexpr} \Arg{item} \Arg{true code} \Arg{false code}
+%   \end{syntax}
+%   Removes the item of \meta{sequence} at the position given by
+%   evaluating the \meta{integer expression} and replaces it by
+%   \meta{item}.  Items are indexed from $1$ on the left/top of the
+%   \meta{sequence}, or from $-1$ on the right/bottom.  If the
+%   \meta{integer expression} is zero or is larger (in absolute value)
+%   than the number of items in the sequence, the \meta{sequence} is not
+%   modified.  In these cases, \cs{seq_put_item:Nnn} raises an error
```
FrankMittelbach, line 449: I was pondering a bit if the more often use case is a possible extension of the sequence, i.e., that in case you get past the existing entries all necessary entries are added (as empty values) but it is propably better to keep this separate (after all it is not clear if the default then should be empty or some value or ... so I guess it is fine as it is now.

FrankMittelbach (approved):

PhelypeOleinik (approved):

l3kernel/l3candidates.dtx
blefloch: Fully agreed: producing an error now means that we are free to change and implement more complicated processing later on.

car222222:
\seq_put_item:Nnn 
places an item at a certain position in the sequence (removing the item that was previously there), and errors if the position is out of bounds (zero or larger [in absolute value] than the number of items)

This sounds to me more like:
\seq_change_item:Nnn


\seq_put_item:NnnTF
does not error, it returns false if 
the position is out of bounds

Wrong name, asabove.  
And also I do not think it 'returns' anything, it just interferes by adding some code after the (possible) change.
(This kind of Mixing of 'functional' and 'non-functional' stuff gets really weird.)

\seq_pop_item:NnN(TF) 
I do not think it is meaningful to pop something in the middle of a sequence!  A seq is just like an array, is it not?

PR #893, opened 2021-04-29
Validate internal structure when showing complex datatypes (fixes #884)

### LINKED ISSUE #884: Showing complex datatypes should validate their internal structure
It would be handy if `\seq_show:N \l_my_seq` would validate that `\l_my_seq` has the expected internal structure.  Currently the show/log functions assume that they are handed a correct variable, but sometimes the variable gets corrupted for various reasons (e.g. using a `tl` function on a more complicated datatype) and this is hard to determine without using the primitive `\show` or having a more restrictive `\seq_show:N`.  List of datatypes that could have extra validation: bool, fp, seq, prop, regex,

The "register" types (int, dim, skip, muskip) which simply use `\showthe` could be a bit more picky and check that one doesn't mix and do `\int_show:N \l_tmpa_dim`.

More complicated perhaps: keys, coffin, flag, clist (spaces are correctly trimmed from all entries), str (correct catcodes), perhaps also msg.  I notice that `fparray_show` is missing.

Not worth a separate issue: `\regex_log:N` doesn't exist.

blefloch:
This slows down `show` commands somewhat, especially `\regex_show:N`/`\regex_log:N`, but at most by a factor of 2 or so, so I figured this is ok.

I opted not to fix the fact that \int_show:N, \dim_show:N etc are equivalent (hence accept the wrong type of data) because this is messy to check.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -25,6 +25,7 @@ this project uses date-based 'snapshot' version identifiers.
 - Show printable characters explicitly in `\regex_show:n`
 - Regex replacement now errors when using a submatch (`\1` etc) for which
   the regex has too few groups
+- Showing complex datatypes now validates their internal structure (issue #884)
 
 ### Fixed
 - Evalutate integer constants only once (issue #861)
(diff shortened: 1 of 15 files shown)


josephwright:
Looks good to me: I don't think performance is an issue for `show`.

PR #901, opened 2021-05-04
More map_tokens functions (fixes #900)

### LINKED ISSUE #900: Provide more map_tokens functions
For various datatypes (`tl`, `seq`, `prop`, `clist`, `str`) we have four types of mapping functions:
- expandable `(type)_map_function:NN`, `(type)_map_tokens:Nn`, which simply differ by the fact that the latter allows for multiple tokens instead of a single function
- protected `(type)_map_inline:Nn` and `(type)_map_variable:NNn`, which differ in how the argument is handed to the code.
In this list, we are actually missing `\clist_map_tokens:Nn` and  `\str_map_tokens:Nn` (and n type versions).

(We also have some `\seq_map_indexed` functions but these are rather specific.)

blefloch:


josephwright (changes requested):
The ideas are fine, but I think we are avoiding adding to `l3candidates`: we should just add directly to the main sources and treat these new functions as stable. I'll probably sort that today and then merge if I have time.

car222222:
Please also remove from the documentation the unique concept of 'mapping to' a list.  It should be 'mapping over'.

PR #906, opened 2021-05-08
Faster align safe groups

Skillmon:
This PR changes the `\group_align_safe_begin:` macro to be faster. This is accomplished by using `\romannumeral` for the evaluation of `\`}` instead of an unnecessary `\ifnum` comparison.

The code in the first commit of this PR is faster than that of the second commit, but uses non-standard category codes and might therefore be slightly less stable in edge cases.

The implementation is more than 10% faster than the previous one.

### DIFF
--- a/l3kernel/l3prg.dtx
+++ b/l3kernel/l3prg.dtx
@@ -1637,11 +1637,8 @@
 %   that the successive expansions of \cs{group_align_safe_begin/end:}
 %   are always brace balanced.
 %    \begin{macrocode}
-\group_begin:
-\tex_catcode:D `\^^@ = 2 \exp_stop_f:
 \cs_new:Npn \group_align_safe_begin:
-  { \exp:w \if_false: { \fi: `^^@ \exp_stop_f: }
-\group_end:
+  { \exp:w \if_false: { \fi: -`} \exp_stop_f: }
 \cs_new:Npn \group_align_safe_end:
   { \if_int_compare:w `{ = \c_zero_int } \fi: }
 %    \end{macrocode}
--- a/l3kernel/l3prg.dtx
+++ b/l3kernel/l3prg.dtx
@@ -1620,7 +1620,7 @@
 %   \TeX{}'s alignment structures present many problems. As Knuth says
 %   himself in \emph{\TeX : The Program}: \enquote{It's sort of a miracle
 %   whenever \tn{halign} or \tn{valign} work, [\ldots]} One problem relates
-%   to commands that internally issues a \tn{cr} but also peek ahead for
+%   to commands that internally issue a \tn{cr} but also peek ahead for
 %   the next character for use in, say, an optional argument. If the
 %   next token happens to be a |&| with category code~4 we get some
 %   sort of weird error message because the underlying
@@ -1631,8 +1631,12 @@
 %   special group so that \TeX{} still thinks it's on safe ground but at
 %   the same time we don't want to introduce any brace group that may
 %   find its way to the output. The following functions help with this
-%   by using code documented only in Appendix~D of
+%   by using behaviour documented only in Appendix~D of
 %   \emph{The \TeX{}book}\dots
+%   In short evaluating |`{| and |`}| as numbers will not change the counter
+%   \TeX{} uses to keep track of its state in an alignment, whereas gobbling a
+%   brace using \cs{if_false:} will affect \TeX's state without producing any
+%   real group.
 %   We place the \cs{if_false:} |{| \cs{fi:} part at that place so
 %   that the successive expansions of \cs{group_align_safe_begin/end:}
 %   are always brace balanced.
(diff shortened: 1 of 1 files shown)


blefloch:
I suggest going back to your non-standard catcode implementation, which is around 10% faster than the one that requires an explicit `-`.  In the same vein, I think the following (with catcode0=1) is 10% faster than the current code, please check I didn't mess up the code.

    \cs_new:Npn \my_end: { \exp:w `^^@ \if_false: } \fi: \exp_stop_f: }

If you could add that to your pull request I can simply squash and merge from github; otherwise I can do it on the command line if you prefer, but then I don't know how to make github aware that your pull was accepted.  Some contributors care, maybe you don't.

------

I've just committed a "fix" in `unravel`, which will unconditionally revert the definition of `\group_align_safe_begin:` (and `_end:`) to one without funny catcodes.  It's not perfect but I don't see it causing any issues.

Skillmon:
@blefloch I did consider replacing `\group_align_safe_end:` as well, but according to my benchmarks this is around 4% slower than the current implementation of it. It seems that `\romannumeral` plus `\iffalse` is slower than a single `\ifnum` (if comparing to `\c_zero_int`). Could you please double check your benchmarks compare to the actual current definition of `\group_align_safe_begin:` (iirc, I had a similar wrong benchmark result when comparing to an `\ifnum` with a literal `0`)?

blefloch:
> @blefloch I did consider replacing `\group_align_safe_end:` as well, but according to my benchmarks this is around 4% slower than the current implementation of it. It seems that `\romannumeral` plus `\iffalse` is slower than a single `\ifnum` (if comparing to `\c_zero_int`).

Ah, have you done your benchmarking in LuaTeX by any chance?  Here are my (ugly) benchmarks.  In LuaTeX the `\romannumeral` version is 20% slower (consistent with what you found) but in other engines (pdfTeX, XeTeX, upTeX, pTeX) it is 8-15% faster.  It's not clear what we should do in such cases but I tend towards keeping the old implementation because I don't want to slow down LuaTeX some more.
```
\RequirePackage{l3benchmark}
\ExplSyntaxOn
\long\def\old{\if_int_compare:w `{=\c_zero_int }\fi:}
\ifx\old\group_align_safe_end:\else\ERROR\fi
\catcode0=1\relax
\long\def\new{\exp:w`^^@\if_false:}\fi:\exp_stop_f:}
\benchmark:n { \new\new\new\new\new }
\benchmark:n { \old\old\old\old\old }
\benchmark:n { \new\new\new\new\new }
\benchmark:n { \old\old\old\old\old }
\benchmark:n { \new\new\new\new\new }
\benchmark:n { \old\old\old\old\old }
\benchmark:n { \new\new\new\new\new }
\benchmark:n { \old\old\old\old\old }
\benchmark:n { \new\new\new\new\new }
\benchmark:n { \old\old\old\old\old }
```
By the way, the improvements you claimed for `\group_align_safe_begin:` are even stronger in engines other than LuaTeX.  This is consistent with LuaTeX being a bit slow when it comes to manipulating characters.

Altogether I suggest simply keeping the first of your implementations of `\group_align_safe_begin:` and leaving the `_end:` as it is.

blefloch:
That is decidedly strange.  Running `pdftex` (or `xetex` or `euptex` or `eptex` or the LaTeX formats) on the following file (not adapted for LuaTeX) consistently gives me that `\new` is 10% faster than `\old`.  I guess it depends on lower-level differences.  I'm on TeXlive, Ubuntu, 64bits.  Anyways, it's not worth digging too much, and in any case we agree that the `_end:` should be left as it is so I'm cherry-picking two of your three commits and including them in main.
```
\ifdefined\pdfelapsedtime\let\elapsedtime\pdfelapsedtime\fi
\def\domany#1{%
  #1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1%
  #1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1%
  #1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1%
  #1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1%
  #1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1#1}
\chardef\zero=0\relax
\long\def\old{\ifnum`{=\zero}\fi}
\unexpanded{\let\sptoken= } %
\catcode0=1\relax
\long\def\new{\romannumeral`^^@\iffalse}\fi\sptoken}
\def\test#1{%
  \count0=\elapsedtime
  \domany{\domany{\domany{#1}}}%
  \immediate\write16{\string#1: \the\numexpr\elapsedtime-\count0}%
}
\test\new \test\old
\test\new \test\old
\test\new \test\old
\test\new \test\old
\test\new \test\old
\csname @@end\endcsname\end
```

Skillmon:
Really strange. I'm on Arch Linux, 64bit, vanilla TeX Live, last updated yesterday evening. For me the suggested implementation is only in `eptex` consistently faster, `euptex` is roughly the same (difference below noise level), and in `pdftex` and `xetex` the old implementation is consistently faster. However, I agree to not dig further into this. Thanks for the "merge" :)

PR #908, opened 2021-05-10
Implement \legacy_set_true_if:n and \legacy_set_if:nn (fixes #664)

### LINKED ISSUE #664: [l3legacy] Add setters for legacy booleans
To complement `\legacy_if:n(TF)`, it would be useful to have functions that abstract over the old Boolean setters (e.g. `\@twosidetrue`), including global versions of such functions. Perhaps inverse setters would be useful, too?

Since this would be a small, simple addition, I’d be happy to make a fork and send a pull request.

blefloch:
- Objections about names are welcome.
- I haven't updated the changelog because it would lead to conflicts when merging.


zauguin:
Regarding names, I would prefer to keep `legacy_if` together, so `\legacy_if_set_true:n` and `\legacy_if_set:nn`. To me `\legacy_set_if:nn` sounds as if it is setting some mysterious legacy object only if a condition applies.

blefloch:
@zauguin Done.  Should I merge?

zauguin (approved):

FrankMittelbach (commented):
I'm not convinced about the name. "if" is very much used in a different way, so seeing this my first reaction was that this is a conditional, so I don't quite agree with @zauguin suggestion. Maybe better would be to call the thing `legacy_bool` but this is just at the top of my head and maybe rubbish.

FrankMittelbach (approved):
ok, I take it back, given that it matches with `legacy_if:nTF`

Issue #896, opened 2021-05-02
l3keys isn't align safe


Skillmon:
The underlying `\keyval_parse:NNn` of `l3keys` isn't alignment safe if the key=value list contains an alignment character outside of braces.

```
\documentclass[]{minimal}

\ExplSyntaxOn
\NewDocumentCommand \foo { m } { \keys_set:nn { foo } {#1} }
\keys_define:nn { foo } { a .tl_set:N = \l_tmpa_tl }
\ExplSyntaxOff

\begin{document}
\begin{tabular}{c}
  \foo{a={a&b}} \\ % works
  \foo{a=a&b} \\ % doesn't work
\end{tabular}
\end{document}
```

Skillmon:
There are a few possible fixes for this. The following implements a few of them as an MWE and runs some test to show in which cases the respective fix would actually work.

```
\documentclass[]{article}

\ExplSyntaxOn
\cs_new:Npn \test:nn #1#2
  {
    \typeout{}
    \typeout { while~ searching~ for~ noalign/omit }
    \tex_halign:D { \typeout{0} ## \typeout{1} \tex_cr:D
      \keyval_parse:nnn {#1} {#2} { a, a=b&c }
      \tex_cr:D
    }
    \typeout { with~ relax }
    \tex_halign:D { \typeout{0} ## \typeout{1} \tex_cr:D
      \relax \keyval_parse:nnn {#1} {#2} { a, a=b&c }
      \tex_cr:D
    }
    \typeout{}
  }
\ExplSyntaxOff

\begin{document}
\ExplSyntaxOn

% no fix applied
\test:nn \use_none:n \use_none:nn % fails with relax
\test:nn { \relax \use_none:n } { \relax \use_none:nn } % fails in both cases

% Fix #1: only one pair of group_align_safe...
% This attempted fix will wrap the whole parsing in a single align-safe group
% starting just before the first step of parsing and closing it when the loop is
% done. The `\exp_args:Nno \use:n` construct is used to expand the actual
% meaning of \keyval_parse:nnn so that I don't have to set up the active
% category codes used there.
\group_begin:
\exp_args:Nno \use:n { \cs_set:Npn \keyval_parse:nnn #1#2#3 }
  {
    \exp_after:wN \group_align_safe_begin:
    \keyval_parse:nnn {#1} {#2} {#3}
  }
\cs_set:Npn \__keyval_end_loop_active:w
    \s__keyval_tail
[...]

PR #912, opened 2021-05-12
Implement expanded-fix for keyval

Skillmon:
This PR implements the `\unexpanded\expanded` fix for `\keyval_parse:nnn` to make it alignment safe. As a fall back, fix number 2 from https://github.com/latex3/latex3/issues/896 is used (with proper adjustments to the error recovery for blank key names).

I ran the testsuite locally on the code used if `\expanded` isn't available. One should be aware that the testsuite will henceforth only test the expanded variant (at least without manual modifications).

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -1016,10 +1016,26 @@
 %   The use of \cs{s_@@_mark} here prevents loss of braces from the key
 %   argument.
 %    \begin{macrocode}
-      \cs_new:Npn \keyval_parse:nnn ##1 ##2 ##3
+      \cs_if_exist:NTF \tex_expanded:D
         {
-          \group_align_safe_begin:
-          \@@_loop_active:nnw {##1} {##2} \s_@@_mark ##3 #1 \s_@@_tail #1
+          \cs_new:Npn \keyval_parse:nnn ##1 ##2 ##3
+            {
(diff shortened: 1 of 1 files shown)


blefloch:
I ended up using parts of the ideas in this pull request but with modifications to keep the code more compact and avoid having twice the same code simply with extra `\group_align_safe_begin:`/`_end:`.  Please take a look at commits https://github.com/latex3/latex3/commit/ec1e3f96cc2a2316aa7d6b14fbdc55cbd5d203ce and https://github.com/latex3/latex3/commit/05cc492ea21ad1e039894004388e1f0b55b4fe80 and check if they seem correct.

Skillmon:
I think it looks good, but only took a quick glance on my mobile. This evening I can take another look and give proper feedback.

Skillmon:
@blefloch took another look, all is looking good. Also, I think the definition of `\@@_blank_key_error:w` could be even more simplified, since the error-case doesn't have to be fast, imho. So the argument grabbing optimization there is really unnecessary. We could go with a single definition, namely:

```tex
\cs_new:Npn \@@_blank_key_error:w #1 \@@_loop_other:nnw
  {
    \__kernel_msg_expandable_error:nn { keyval } { blank-key-name }
    \@@_loop_other:nnw
  }
```

PR #918, opened 2021-05-15
l3benchmark: Store results in a variable

Skillmon:
The current output format of `\benchmark:n` is good for humans reading the console output. But I'd often like to further process the results of benchmarks, comparing them with other things, or printing them in a way easy to read by further tools down the road. For this it'd be nice to be able to store the benchmark result inside a variable instead of printing them on screen.

This PR will add a function `\benchmark_set_fp_to_ops:nN` (I'm open for discussions about the name) that does exactly that, it runs the benchmark and instead of displaying the results on the terminal will store the estimated elementary operations inside an `fp` variable provided by the user.


blefloch:
I like it.  Possible improvements (but I haven't thought very much about it).
- Rename to `\benchmark_ops_get_fp:nN`. (Or perhaps just provide `\benchmark_ops_get:nN` expecting a tl variable and giving the result of `\fp_eval:n`?)
- Provide the same function without `_ops` to get the raw time, which may be more interesting to some people?
- Update `l3experimental/CHANGELOG.md`.

Skillmon:
Regarding your points:

- the shorted name looks appealing, but I'd stick to an `fp` variable (further processing using `l3fp` functions should be faster with an `fp` variable, aren't they? I don't see much advantage of the `tl` result over the `fp` result, except that the user doesn't need to do `\fp_to_tl:N` himself)
- I personally don't care too much about the raw times, as they might vary quite a lot from one run to another, depending on general system load, and are completely useless between two different systems, whereas the normalization to estimated ops in general gives good comparable results. However, providing `\benchmark_get_ops:nN`, `\benchmark_get_time:nN` and `\benchmark_get_results:nNN` seems like a good idea to give each user the choice.
- I'll update the changelog :)

blefloch:
We have a public variable `\g_benchmark_duration_target_fp` that affects `\benchmark:n`.  This makes me think we could simply make `\g_@@_time_fp` and `\g_@@_ops_fp` etc internal fp into public fp, and document that `\benchmark:n` simply stores its results into two fp variables.  What do you think?

Skillmon:
Apart from the fact that `\g_@@_ops_fp` doesn't yet exist, afaik, this would be viable as well. In that case some way to disable the normal output of `\benchmark:n` would be needed, though.

blefloch:
Ah, good point about silencing `\benchmark:n` (and `\benchmark_once:n`).  I hesitate between adding `\benchmark_silent:n` and `\benchmark_once_silent:n`, or adding a public boolean `\g_benchmark_silent_bool` controlling that.  If you have a preference, please implement it and I'll merge.  Otherwise maybe we have to ask others to weigh in.

Skillmon:
I think I'd prefer the different names, but I don't really have a strong opinion about that. The boolean variant has only a single csname more, and would only require a small change in `\@@_display:`, so maybe that's the better choice here.

blefloch:
Ok, let's go for two `_silent` names and making the `\g_@@_..._fp` variable public.  Do you want to do it, or should I do it?

Skillmon:
I just implemented the second approach... :) But I can revert and do the other as well.

blefloch (approved):
Thanks @Skillmon for all these changes; I'll squash and merge now.

Issue #913, opened 2021-05-12
Balancing regex replacements without error


blefloch:
Assume that one performs a regex replacement such as `\tl_set:Nn \l_tmpa_tl { a { b } c } \regex_replace:nnN { a . } { \0 \0 } \l_tmpa_tl`.  This should give an unbalanced token list `a{a{b}c`, so l3regex correctly detects the problem and adds a missing closing brace after complaining.  I think we should have a way to specify that a given regex replacement should be balanced if needed, and a way to test after the fact whether rebalancing was needed.  This could be achieved using three variables:

- `\l_regex_auto_balance_bool` should be set before doing a regex replacement, to indicate that l3regex should rebalance without complaining.  This could also be useful in `\peek_regex_replace` to implement #820 (xparse regex arg type).
- `\l_regex_added_group_begin_int` and `\l_regex_added_group_end_int` holding the number of `{` and `}` that were added to the beginning and end of the last regex replacement.

Better names welcome, but more importantly, is it a good idea to have such customizations in a kernel-level command?

josephwright:
Without a clear use case, always having an error here seems reasonable.

Issue #388, opened 2017-08-02
Character class for Greek letters


blefloch:
In l3regex it would be useful to provide some shorthand for matching all Greek letters, both as Unicode code-points and as control sequences (probably two separate shorthands).  This feature request stems from a comment in #261.

Perhaps other similar classes would be useful?  Which ones?

wspr:
A class for the control sequences could be useful. A class for the Unicode codepoints opens a bit of a can of worms — do you include only what you might crudely refer to as “ascii greek”, or do you include the mathematical greek symbols too? What about other Greek letters that aren’t used in maths?

If you match Greek letters, what about Cyrillic, etc. etc.? I guess it’s conceivable you could match to named ranges of Unicode according to the standard subdivisions.

blefloch:
Ah, the correct way for Unicode characters is probably Unicode properties.  The standard notation seems to be `\p{Greek}` (and `\P` instead for any char that does not have the given property).

For control sequences, I can't figure out a good notation.  Perhaps `\pc{Greek}` or `\cp{Greek}` with `c` referring to the `\c` construction and `p` to property?

But perhaps looking for Greek control sequences is not really the right thing to do and the correct question is whether what follows is a mathalpha, mathbin, mathrel, etc?

FrankMittelbach:
> Ah, the correct way for Unicode characters is probably Unicode 
> properties. The standard notation seems to be |\p{Greek}| (and |\P| 
> instead for any char that does not have the given property).

seems a good approach to go in that direction (if at all)

> For control sequences, I can't figure out a good notation. Perhaps 
> |\pc{Greek}| or |\cp{Greek}| with |c| referring to the |\c| construction 
> and |p| to property?
> 
> But perhaps looking for Greek control sequences is not really the right 
> thing to do and the correct question is whether what follows is a 
> mathalpha, mathbin, mathrel, etc?

the latter looks like an interesting idea, but in any case rather than 
implementing I think it really needs a good number of use cases why you 
want "greek cs chars all mathbins or ... why not all integrals or all ...

the possibilities seems endless so I think first step is to collect real 
requests why a class is useful to search for

blefloch:
Frank is right that we need use cases.

See for instance `\l__mhchem_regex_oneGreekLetter_fregex` in `mhchem.sty` (and many others there).  I didn't look at why exactly these letters are allowed and not `\digamma` or `\varkappa` (probably overlooked).

[...]

PR #919, opened 2021-05-15
Provide regex code callout

blefloch:
We are sometimes asked to extend l3regex to have more tests: testing for Greek mathematical letters (#388), testing the meaning of a token, working with grapheme clusters, etc.  Some regular expression engines have ways to call arbitrary code to decide whether to match at a given position.  Tasks:

- [x] Assertions which match or don't match at a given position according to user-provided code: `(?C{my_callout:TF})`

- [ ] Callout groups `(?C{my_callout:TF} explicit-regex)` which match `explicit-regex` then call `\my_callout:TF` with suitable variables being set.  This requires making some intarrays public, or extracting into a variable the text that was matched, but see the balancing issue described at #913.  Precise syntax unclear: can `explicit-regex` have alternatives?

- [ ] Allow regex functions inside a callout: this requires manual intarray and `\toks` bookkeeping.


josephwright:
@blefloch  Could we merge this as-is? The gain looks good to me already.

blefloch:
I strongly suspect there are still important bugs.  Sorry no time atm.

Issue #922, opened 2021-05-16
\vbox_set_top_to_ht?


lvjr:
There are `\vbox_set_top` and `\vbox_set_to_ht` in `l3box`, but not `\vbox_set_top_to_ht`.

blefloch:
This is a duplicate of #552, so I'll close here, but I'm implementing it now.

Issue #552, opened 2019-03-05
Missing \vbox_set_top:Nw


PhelypeOleinik:
I noticed that `expl3` provides a `\vbox_set:Nw`...`\vbox_set_end:` alternative to `\vbox_set:Nn`, but there is no `\vbox_set_top:Nw` counterpart for `\vbox_set_top:Nn`. If this is not intentional, I think it's worth adding :)

Also (I don't know if this one makes sense but) there is a `\vbox_set_to_ht:Nnn` but no `\vbox_set_top_to_ht:Nnn`.

blefloch:
Hum... See pull request #923.  Regarding the `\vbox_set_top_to_ht:Nnn` (specifically requested in #922), I thought it was a trivial `\setbox #1 = \vtop to #2 {#3}`, but actually this primitive construction has a weird effect.  E.g. `\vtop to 30pt{j\par j}` (in plain TeX) gives dimensions `\vbox(6.67859+25.26585)x469.75499`, where `6.67859+25.26585=31.94444` is 30pt plus the baseline of the bottom line.
- First, we need to understand that `\vbox to #2 {#3}` makes a vbox, then sets its height (not height+depth!) to the requested value.
- Then `\vtop to #2 {#3}` makes a box of the same total size as what would be obtained by `\vbox to #2 {#3}`, but with its baseline shifted.  So `#2` ends up being neither the height nor the depth of the resulting box.

blefloch:
@lvjr, could you comment on why you need `\vbox_set_top_to_ht:Nnn`?  It might help clarify what's needed in this direction.

lvjr:
Maybe I have been misunderstanding ` \vtop to #2 {#3}`. In my [`tabularray`](https://ctan.org/pkg/tabularray) package, there are two new column types `h` and `f`, which will align cell text at row head and row foot, respectively. I was thinking that using `\vbox_set_top_to_ht:Nnn \vfil` would correctly align two `h` cells.

dcpurton:
Just to add a vote to this, I'd use `\vbox_set_top:Nw`.

blefloch:
(I'm sorting some branches that I have lying around in my copy of the repository.)  I think we should add `\vbox_set_top:Nw` (and the `cw` variant, and the `gset_top` and the corresponding `end`), which are quite natural.  Regarding a wrapper around `\tex_vtop:D to ...`, it's less clear and we should probably wait to have a fuller rethinking of `l3box` and what tools we want to provide.

[...]

PR #923, opened 2021-05-16
Wrappers around \vtop (see #552 and #922)

blefloch:
Upon request at #922 and #552, I implemented here `\vbox_set_top_to_ht:Nnn` (and some `w` versions that don't need discussion).  This is a wrapper around `\setbox #1 = \vtop to #2 {#3}`.  However, this primitive does very weird things: it first sets the material with a certain distance `#2` between the top of the box and the last baseline (just like `\vbox to`), then it shifts the baseline to be that of the first line.  At the end of the day `#2` is neither the height nor the depth of the resulting box.

Actually, the height of the `\vtop` should always be the height of the first line, otherwise the baselines wouldn't coincide.  So perhaps by symmetry with the `\vbox` case we should provide `\vbox_set_top_to_dp:Nnn` (and not `to_ht`) which sets the depth (namely the distance between the first baseline and the bottom).  For this I need to know: **is it safe to do `\setbox0\vtop{\unvbox0}` or will I lose whatsits or inserts etc?**  I'm thinking of doing the following:
- Set the material to its natural size with `\setbox0\vbox{#3}` to determine the natural total size, and the height of the first line.
- Reset the material in a vtop with `\setbox2\vtop{\unvbox0}` to determine the depth of the last line.
- Then do `\setbox #1=\vtop to \dimexpr #2 + \ht0 - \dp2 {\unvbox2}` (well, actually I need to have saved `\ht0` but you get the gist).

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -15,6 +15,7 @@ this project uses date-based 'snapshot' version identifiers.
   `\ior_show:N`, `\ior_log:N`, `\iow_show:N`, `\iow_log:N`,
   `\tl_log_analysis:N`, `\tl_log_analysis:n`
 - `\legacy_if_set_true:n`, `\legacy_if_set_false:n`, `\legacy_if_set:nn`
+- `\vbox_set_top_to_ht:Nnn`, `\vbox_set_top:Nw`, `\vbox_set_top_to_ht:Nnw` (issue #922)
 
 ### Fixed
 - Checking brace balance in all regex functions (issue #377)
--- a/l3kernel/l3box.dtx
+++ b/l3kernel/l3box.dtx
@@ -513,69 +513,85 @@
 %   and then includes this box in the current list for typesetting.
 % \end{function}
 %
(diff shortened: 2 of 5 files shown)


blefloch:
Relatedly, David Kastrup had an [old BachoTeX contribution](http://www.gust.org.pl/projects/pearls/2008p/david-kastrup/bachotex2008-david-kastrup-pearl1.pdf) in which he build an improved `\vtop` that detects whatsits at the start.  I don't yet understand it because I'm not familiar with `\vsplit`, and I think that it purposefully removes these whatsits.

car222222:
Whatever else is provided, the original `\vtop to …` should still be available, maybe as `\vbox_set_top_to:Nnn`, avoiding the implication that it is the ht that is prescribed.

car222222:
Also (but maybe not now possible) the order in these names should be:
`\vbox_top_set_ … `

car222222:
But now it seems that we do not provide even`\vbox_top_to:nn` (at least I could  not find it).

PR #928, opened 2021-05-16
Tools to match/replace multiple regex concurrently (fixes #433)

### LINKED ISSUE #433: Multiple regex replacements at once
In https://tex.stackexchange.com/questions/407637/ it would have been
convenient to replace all `(` by `\paren{` and all `)` by `}` at once to 
make parenthesis-matching rely on TeX's matching of braces.  (Instead I 
abused `xparse`'s delimiter matching.)  Unfortunately, `l3regex` only 
allows to make one replacement at a time and this would result in an 
unbalanced token list, rightfully refused by `l3regex`.  I propose
`\regex_replace_case_once:nN(TF)` and `\regex_replace_case_all:nN(TF)`
that would perform several replacements at once:

    \regex_replace_case_all:nN
      {
        { \( } { \c{paren} \cB\{ }
        { \) } { \cE\} }
      }
      \l_tmpa_tl

Not sure about the order of words in the function's name.  Another
choice is to provide a replacement command that x-expands the 
replacement parts (after finding the matches):

    \regex_replace_all_x:nnN
      { [()] } % match open or close parenthesis
      {
        \c{if:w} \( \0
          \c{paren} \cB\{
        \c{else:}
          \cE\}
        \c{fi:}
      }

I might be missing some obvious alternatives and other possible interfaces.

blefloch:
This implements `\regex_case_match:nn(TF)`, `\regex_case_replace_once:nN(TF)`, `\regex_case_replace_all:nN(TF)` which can do different code/replacements depending on which of several regex match.  See the documentation in l3regex and issue #433 for details.

I don't expect the code to be reviewed in detail as it's pretty complicated and connects to many parts of l3regex, but if someone else can test beyond the existing test files that would be great.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -10,6 +10,8 @@ this project uses date-based 'snapshot' version identifiers.
 ### Added
 - `\prop_to_keyval:N`
 - Support for validity scope for keys
+- `\regex_match_case:nn(TF)`, `\regex_replace_case_once:nN(TF)`,
+  `\regex_replace_case_all:nN(TF)`
 
 ### Changed
 - Allow indirect conversions between colorspaces though fallback models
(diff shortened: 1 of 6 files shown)


josephwright:
My question here is about the interface: should these be `\regex_match_case`/`\regex_replace_case`? Other than that, I think we should get an and add.

josephwright:
My comment about naming still stands here: anyone want to comment (@eg9, @FrankMittelbach?)

PR #937, opened 2021-05-20
Register pgfmxfp

Skillmon:
Register the `pgfmath-xfp` package that uses the prefix `pgfmxfp`.


Skillmon:
Wait, this was unclean. I'll recreate this.

Skillmon:
Ok, now it should be good to go.

josephwright:
Seems to be a mix of two independent PRs?

blefloch:
@josephwright The history is not clean but the end-result of squashing everything should be a one-line commit.  (See the tab "Files changed" on Github.)

Skillmon:
@josephwright yes, I was creating this branch from the wrong base branch by accident. That's why I reverted the two commits and merged the latest master. @blefloch is right, the result is just a single line added. So after a squash this should look like I did the correct thing from the beginning.

l3kernel/doc/l3prefixes.csv
```diff
@@ -163,6 +163,7 @@ pdfxform,l3pdfmanagement,The LaTeX Project,https://www.latex-project.org/latex3.
 peek,l3kernel,The LaTeX Project,https://www.latex-project.org/latex3.html,https://github.com/latex3/latex3.git,https://github.com/latex3/latex3/issues,2012-09-27,2012-09-27,
 percent,l3kernel,The LaTeX Project,https://www.latex-project.org/latex3.html,https://github.com/latex3/latex3.git,https://github.com/latex3/latex3/issues,2018-05-12,2018-05-12,
 pgf,pgf,The PGF/TikZ Team,https://pgf-tikz.github.io,https://github.com/pgf-tikz/pgf,https://github.com/pgf-tikz/pgf/issues,2020-07-03,2020-07-03,
+pgfmxfp,pgfmath-xfp,Jonathan P. Spratte,https://github.com/Skillmon/ltx_pgfmath-xfp,https://github.com/Skillmon/ltx_pgfmath-xfp,https://github.com/Skillmon/ltx_pgfmath-xfp/issues,2021-05-20,2021-05-19,
```
zauguin, line 166: Your `Last update` field is earlier than `First registered`. That seems wrong.

zauguin (changes requested):

l3kernel/doc/l3prefixes.csv
zauguin: I always assumed that `Last update` refers to the last update to the entry in `l3prefixes` and not the last update of the package.

@josephwright I'm guessing you created the file/column. Any insights regarding the intended meaning?

Skillmon: That would be more reasonable (at least all my other entries don't reflect the latest package version).

FrankMittelbach: As we are not trying to keep track of package versions here, I think the  two columns have to be interpreted as: "when was this entry first registered" and "when was ithe entry last updated".

 the second is an (small) indicator of whether or not the data is likely to be current (in case there seem to be an issue, like not finding the listed home page, for example.

zauguin (approved):

PR #944, opened 2021-06-04
Use Lua for color

josephwright:
The aim here is to use Lua to avoid whatsits, certain for PDF mode and potentially for other backends.


josephwright:
Notice that I've just ripped of Heiko's code: @zauguin has plans about the way to actually add the colour data. I've also tried to keep everything in Lua as far as I can, other than the settings for the macros we require in the backend.

We likely need to think about how to fit transparency in here: the idea is the same but it needs a different, parallel, attribute.

l3backend/l3backend-color.dtx
```diff
@@ -1248,6 +1253,401 @@
...
+%    \begin{macrocode}
+luacmd('tex_pdfxform:D', function()
+  local attr = scan_keyword'attr' and scan_string() or nil
+  local resources = scan_keyword'length' and scan_string() or nil
+  local xtype = scan_keyword'type' and scan_string() or nil
+  local margin = scan_keyword'margin' and scan_string() or nil
+  local n = scan_int() or nil
+  traverse(getbox(n),"")
+  saveboxresources(n,attr,resources,false,xtype,margin)
+end,'global','protected')
```
zauguin, line 1605: The XForm issue...

This breaks *every* use of `\pdf_xform:D` in TeX Live. :laughing:  (Since it doesn't work with `\tex_immediate:D`) Properly supporting `\tex_immediate:D` here is complicated, as soon as we are non-expandable immediate breaks, if we are expandable it doesn't behave like the primitive and supporting immediate is still a pain. The mess in luacolor is the closest solution I can think of, but even that is far from perfect and probably inappropriate for kernel code.

So I think the only reasonable way forward is to change the name and provide a second name for the immediate version. If we use non-expl3 names (or aliases) and define them with also for pdfTeX is should be relatively simple to get most users to change. Since many uses of this happen in generic code, we have to make sure not to depend on LaTeX there, so I would go for just the wrapped primitive.

Anyway, this is a much more general issue than just for colors.

l3backend/l3backend-color.dtx
```diff
@@ -1248,6 +1253,401 @@
...
+%    \begin{macrocode}
+local function color_index(color)
+  local n = map[color]
+  if not n then
+    n = map.n + 1
+    map[n] = color
+    map[color] = n
+  end
+  return n
+end
```
zauguin, line 1311: Not important for the first version, but I think `color_attr` (read-only) and `color_index` should be available for external code. One of the less obvious but still important advantages of the attribute approach is that this can be used with nodes which are generated programmatically.

l3backend/l3backend-color.dtx
```diff
@@ -1248,6 +1253,401 @@
...
+        if n.subtype >= 100 then
+          if n.leader.id == id("rule") then
+            return node_map.color
+          else
+            return node_map.list_leaders
+          end
+        end
+      end
+  }
+%    \end{macrocode}
```
zauguin, line 1427: This is duplicated.

l3backend/l3backend-color.dtx
```diff
@@ -1248,6 +1253,401 @@
...
+%   one-off information in the function that needs it. We also use a single
+%   local \enquote{map} for the node values.
+%    \begin{macrocode}
+local node_map = {
+    list         = 1 ,
+    list_leaders = 2 ,
+    list_disc    = 3 ,
+    color        = 4 ,
+    no_color     = 5
+  }
```
zauguin, line 1360: IMO these should just be separate variables since we never use `node_map` except for accessing an explicitly specified elements.

l3backend/l3backend-color.dtx
```diff
@@ -1248,6 +1253,401 @@
...
+%    \end{macrocode}
+%
+% At present we only set up for \LuaTeX{} in PDF mode: later we will expand
+% on that!
+%
+%    \begin{macrocode}
+%<*luatex>
+%    \end{macrocode}
+%
+% This needs some work: undefine from Lua?
```
zauguin, line 1624: We can't undefine from Lua (we can use runtoks to run `tex_undefine` while Lua is active, but that's not really from Lua. Might be a useful addition to ltx.util though), but `luacmd` could get a parameter to specify that it is allowed to overwrite. We first have to decide if we really want to overwrite it though.

l3backend/l3backend-color.dtx
```diff
@@ -1248,6 +1253,401 @@
...
+end
+%    \end{macrocode}
+% \end{macro}
+%
+%   Local copies of global tables.
+%    \begin{macrocode}
+local node   = node
+local string = string
+local tex    = tex
+local texio  = texio
```
zauguin, line 1320: `texio` isn't used.

josephwright: Fixed in 373c6090b930c0d9c1585c63c84995ccf5448870

l3backend/l3backend-color.dtx
josephwright: I'll come back to this after we deal with the wider question of what exactly we _do_ want to define.

l3backend/l3backend-color.dtx
josephwright: They can be - that's of course what Heiko does. I felt that as this was one 'idea' (mapping from node type to number), it should be in one place. It's not hard to alter - something like `node...` so they have a common pattern?

l3backend/l3backend-color.dtx
josephwright: Fixed in 464b5be2fd5845dc1e0776b9cc74ffd33e93c007

l3backend/l3backend-color.dtx
josephwright: I wasn't sure how `\immediate` interacted with the Lua here - sounds like 'not properly'.

In any case, we don't necessarily want/need to worry about that. I very deliberately only set `\tex_pdfxfom:D`, meaning that the only place I'm expecting this token to be used is from within `l3backend`. At the moment, `l3color` is pretty specialised and I think I'd be happy with a position of

> If you use `l3color` inside an XForm, it's all got to be handled by `expl3`.

So we could then define some `\__kernel_box_xform:w` function(s) and use them  - no messing with other people's code.

l3backend/l3backend-color.dtx
josephwright: I was aiming to keep interfaces to a minimum at least at this stage. I assume you are thinking about using directly from Lua, in which case wouldn't we still keep the `attr` and the `index` private and provide `ltx.color.set` or something?

l3backend/l3backend-color.dtx
josephwright: As in, if we are not setting the token `\tex_pdfxform:D`, this is not an issue.

l3backend/l3backend-color.dtx
zauguin: The problem I see with

> If you use l3color inside an XForm, it's all got to be handled by expl3.

is that you often can't really influence if your code is used in an XForm. E.g. some user using code in a TikZ label shouldn't have to worry about if the code uses l3color or not and if the TikZ document happens to use `expl3` and needs XForms (Or even what XForms are).

Another possible solution to the original issue: We could avoid the problem if we are open to redefining `\immediate` itself. A custom `\immediate` emulation would make `\immediate` compatible with Lua commands, but it's meaning would become `luacall n` and tracing would show an additional `luacall` everytime `\immediate` is used.

l3backend/l3backend-color.dtx
zauguin: > I assume you are thinking about using directly from Lua, in which case wouldn't we still keep the `attr` and the `index` private and provide `ltx.color.set` or something?

I think that having direct access is also useful since it allows e.g. to be more precise when inheriting attributes (when creating a node between two other nodes, it can be useful to only inherit their color attribute if both are colored the same / if at least one is colored etc. to avoid unnecessary color changes, especially when you know that it wouldn't have an effect on your node anyway). Not sure about it though.



> I was aiming to keep interfaces to a minimum at least at this stage.

Fully agreed.

l3backend/l3backend-color.dtx
zauguin: >  felt that as this was one 'idea' (mapping from node type to number), it should be in one place.

Then let's leave it like this. It's probably mostly a style question anyway, any performance impact is probably non-measurable anyway.

l3backend/l3backend-color.dtx
FrankMittelbach: > Another possible solution to the original issue: We could avoid the problem if we are open to redefining `\immediate` itself. A custom `\immediate` emulation would make `\immediate` compatible with Lua commands, but it's meaning would become `luacall n` and tracing would show an additional `luacall` everytime `\immediate` is used.

I wouldn't be surprised at all if that is hitting us in the face. Maybe a `\protected` version would do, but it sounds dangerous.

zauguin: > > Another possible solution to the original issue: We could avoid the problem if we are open to redefining `\immediate` itself. A custom `\immediate` emulation would make `\immediate` compatible with Lua commands, but it's meaning would become `luacall n` and tracing would show an additional `luacall` everytime `\immediate` is used.
> 
> I wouldn't be surprised at all if that is hitting us in the face. Maybe a `\protected` version would do, but it sounds dangerous.

In which way? Do you think that people actually test for `\immediate` being the primitive or do you think that it would break unrelated code since it would behave differently"

josephwright: > The problem I see with
> 
> > If you use l3color inside an XForm, it's all got to be handled by expl3.
> 
> is that you often can't really influence if your code is used in an XForm. E.g. some user using code in a TikZ label shouldn't have to worry about if the code uses l3color or not and if the TikZ document happens to use `expl3` and needs XForms (Or even what XForms are).

At least at present, I'm seeing `l3color` as a definite 'opt in when you know you want it', for things like `l3draw`. As such, I don't expect anything using `l3color` to be used inside an arbitrary colour-dependent context: anything inside `pgf` needs to be using classical `color` I'd say. Even allowing `pgf`, there are not a lot of use cases for XForms that use text: the obvious one is for graphics inclusion, and that's not subject to Lua-based colour.

Of course, we can decide to provide some more abstract interfaces for XForms, but they'll need to be in another place:  if we want arbitrary code to use them, they've got to work in plain (as `pgf` does), and that means not in `expl3` (as plain users might load `graphics.tex` but will likely not go for `expl3` in a routine sense).

l3backend/l3backend-color.dtx
josephwright: I 

> > I assume you are thinking about using directly from Lua, in which case wouldn't we still keep the `attr` and the `index` private and provide `ltx.color.set` or something?
> 
> I think that having direct access is also useful since it allows e.g. to be more precise when inheriting attributes (when creating a node between two other nodes, it can be useful to only inherit their color attribute if both are colored the same / if at least one is colored etc. to avoid unnecessary color changes, especially when you know that it wouldn't have an effect on your node anyway). Not sure about it though.

 I guess what I mean is that they are just numbers, and those should really always be internal representations of abstract concepts. TeX is already bad enough with `\catcode` ....

l3backend/l3backend-color.dtx
josephwright: I'm happy to alter: I'd settled on `id_<name>` ...

l3backend/l3backend-color.dtx
zauguin: Right. 

You're right.

> So we could then define some `\__kernel_box_xform:w` function(s) and use them - no messing with other people's code.

Namewise I think it should include `pdf` since it's a rather PDF specific concept.

l3backend/l3backend-color.dtx
```diff
@@ -1552,22 +1552,32 @@ end
...
 % \end{macro}
 %
-% \begin{macro}{\tex_pdfxform:D}
-%   We redefine the \pkg{expl3} version of the primitive to do the replacement
-%   implicitly. The keywords \texttt{margin} and \texttt{type} are not
-%   documented at the \TeX{} level.
+%<@@=pdfxform>
+%
+% \begin{macro}{pdfxform}
+% \begin{macro}{\@@_backend_now:w, \@@_backend_shipout:w}
```
u-fischer, line 1558: I think the names are wrong in this context. The difference between immediate and non-immediate xform are not related to shipout. An non-immediate xform only reserve an object number and are kept in memory and only moved to the pdf when referenced with `\pdfrefxform`

josephwright: Hmm, to me that suggests that someone (Hans?) made a bit of a mis-step in having it take any notice of `\immediate` in the first place: that's _supposed_ to be about whether to wait for shipout or not.

u-fischer: well it is the same with pdfobjects. There are only written when referenced or if you use \immediate. It is a bit confusing that both use the \immediate command to distinguish the case, which is normally connected to "directly or at shipout", but probably they used it as it was there, and why invent a new command name?

josephwright: @davidcarlisle Suggested that we look in the 2e kernel at a full set of macro interfaces for handling `\immediate` vs 'later' in a way that's easier to integrate with LuaTeX. On that basis, I think I would give macros for 'now or when used' different names from 'now or shipout'. Something like 'firstuse'?

u-fischer: `now` and `delayed`? We could then keep `\pdf_object_write:nn` as the "now" variant and add `\pdf_object_write_delayed:nn` and similar for xform. Imho it would be useful, it happens that one doesn't know if an object will be really needed at the end, for example for namespaces.

zauguin: > add `\pdf_object_write_delayed:nn` and similar for xform.

I think the names for objects and xforms have to reflect that they behave quite differently since objects are "only" written out, while the xform equivalent will also always reference the xform in the the current page stream IIRC.

josephwright: At the moment this is all strictly internal stuff - although I agree we should try to get to a position that can go public

u-fischer: @zauguin objects and xform are different, but I think that is reflected in the start of the name `\pdf_object` and `\pdfxform`,  the behaviour "delayed" part is imho similar enough to justify the same word. In both cases the "object" hangs around in the limbo and is dropped if not used, and when used both behave as if the immediate version has been used there.

l3backend/l3backend-color.dtx
josephwright: I think we are agreed here: we are not touching the primitive name itself, rather a one-level abstraction.

l3backend/l3backend-color.dtx
```diff
@@ -1412,19 +1413,32 @@ end
 %   The mode is hard-coded as $2$: a PDF literal.
 %    \begin{macrocode}
 local function set_color(head, n, new_color, old_color)
-  if new_color == old_color or not new_color then return head end
-  local color_node = node_new("whatsit","pdf_literal")
-  color_node.mode = 2
-  color_node.data = new_color
+  if new_color == old_color then return head end
+  local color_node = node_new("whatsit","pdf_colorstack")
```
josephwright, line 1416: Perhaps I'm missing something, but how does this select the appropriate colour stack?

zauguin: Colors are controlled by colorstack 0 which is the default after `node_new`. We should probably make this more explicit.

josephwright: That's going to need some thought. How does this work in a case where you'd need a separate colour stack in pdfTeX, e.g. footnotes breaking over pages? Are we 'OK' as we are dealing with nodes?

zauguin: As long as the user is actually using `l3color` it's fine since the attribute approach handles footnotes correctly anyway.

But using non attribute colors based on another colorstack interleaved with attribute based colors might reset the wrong stack. We can avoid that by storing the stack number as an additional attribute or by adding a user defined whatsit in order to switch the currently used colorstack. (Which make more sense depends a bit on where the nodes are created. E.g. if a saved box is added in the middle of a footnote it should probably use the footnote stack but that might not be known when the box is created... Then again using classical color commands there would probably use the main colorstack, so using them for luacolors too would be more consistent.)

In any case this is rather easy to adapt later if the issue ever actually appears.

josephwright: OK, that's fine then: I think we look hard at getting all colour in LuaTeX to the attribute-based approach.

josephwright:
(I guess we might not need multiple stacks using attributes, but the data have to get to the right place)

muzimuzhi:
The base branch needs to switch from `main` to `develop`.

This PR was opened so early that at that time there was just the `main` branch.

josephwright:
@zauguin Are we OK here now with Xform objects?

josephwright:
> We need to sort `\mathcolor`

@u-fischer  and I have been looking at this: we wonder if we can save _always_ and then simply always-reset rather than make the current 'pop' a no-op.

josephwright:
> @zauguin Are we OK here now with Xform objects?

@zauguin Ping ...

josephwright:
We really need to move this forward: the current situation is awkward (the `luacolor` package, `pdfmanagement` on or off, ...). I think it's basically right, so it's just a question of getting final agreement.

josephwright:
I've rebased onto the current develop, and looked over things: math color still isn't right here (both leakage and some very odd `.tlg` output when you look at it). That plus the Xforms stuff needs to be addressed, then I really would like to merge.

josephwright:
The math colour business also applies to opacity: both need a reset without a group. We also need to extend to use the stack number to separate out colour and opacity, I guess (unless we can handle in one?).

josephwright:
I've now got the visual output addressed for math mode, but we still have whatsits present ...

l3kernel/l3backend-color.dtx
```diff
@@ -1480,24 +1510,393 @@ if luaotfload and luaotfload.set_transparent_colorstack then
...
+local id_no_color     = 5
+local node_types = {
+    [id("hlist")]   = id_list      ,
+    [id("vlist")]   = id_list      ,
+    [id("rule")]    = id_color     ,
+    [id("glyph")]   = id_color     ,
+    [id("disc")]    = id_list_disc ,
+    [id("whatsit")] = {
+        [subtype("special")]        = id_color ,
+        [subtype("pdf_colorstack")] = id_color ,
```
josephwright, line 1623: I notice that `luacolor` has a different setup here - I'm not sure why we've deviated

josephwright:
I've been trying to work out why we still get whatsits for the math code: really not sure at all. I tried setting the attribute to 0, which can be made to work but I'm not sure is any better than the macro storage. The whatsits seem to be a bit of a random selection of `push` and `set` :-(

For opacity, I note that the `pdf_colorstack` node type has a `stack` value we can set, so generalising is just a question of passing on the stack number and handling it in the colour table (presumably adding an index). But that also needs non-grouped restoration, and I want to understand what the heck is happening!

PR #949, opened 2021-06-14
Fix local assignment to a global variable '\g__sys_backend_tl'

dbitouze:


josephwright:
Could you add a CHANGELOG entry?

dbitouze:
> Could you add a CHANGELOG entry?

Of course. In which one?

PhelypeOleinik:
@dbitouze The one in `l3kernel`: https://github.com/latex3/latex3/blob/main/l3kernel/CHANGELOG.md

PR #956, opened 2021-07-01
Move l3opacity to l3experimental

josephwright:


car222222:
What a thin wrapper!  But looks harmless.

Also, a tiny file.  Could this content be added to l3color?

josephwright:
> What a thin wrapper! But looks harmless.
> 
> Also, a tiny file. Could this content be added to l3color?

Well yes, that was where I started out with the code, but in the main we have separate files for separate modules (now we put everything together with DocStrip, we probably didn't need for example to put `l3file` and `l3io` together). It could certainly go into the same bundle as `l3color` if that's cleaner.

FrankMittelbach (approved):

PR #958, opened 2021-07-04
Speed up string comparison by using \if

Skillmon:
The `\pdfstrcmp` primitive (and equivalent implementations, including the current Lua code used by `expl3` in LuaTeX) doesn't return its result as a count register or similar, but as a tokens of category 12. It is therefore faster to check whether the result is equal to `0` by using the `\if` primitive instead of `\ifnum` (my benchmarks indicate 10% speed difference for the check in pdfTeX).

This changes those usages of `\pdfstrcmp` in which the result is only tested to be equal to zero to use `\if`.

### DIFF
--- a/l3kernel/l3fp-expo.dtx
+++ b/l3kernel/l3fp-expo.dtx
@@ -1000,8 +1000,7 @@
 \cs_new:Npn \@@_pow_normal_o:ww
     \s_@@ \@@_chk:w 1 #1#2#3; \s_@@ \@@_chk:w #4#5
   {
-    \if_int_compare:w \@@_str_if_eq:nn { #2 #3 }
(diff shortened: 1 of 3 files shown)


FrankMittelbach (changes requested):
Imo that kind of change should be documented with short note giving the reason, otherwise it will later be incomprehensible (or changed back by mistake as the speed gain is not necessarily obvious).

Skillmon:
The failing checks aren't my fault, I got "lucky" again with the midyear deprecations.

Skillmon:
Just a prod: How does it look now?

FrankMittelbach (approved):
can't see anything wrong with that but by the end of the day it is @blefloch  who needs to decide and he unavailable during August.

Skillmon:
@FrankMittelbach ok, thanks for taking the time!

Issue #968, opened 2021-07-17
fpexpr → fp expr in interface3.pdf 


jlaurens:
change the`fpexpr` to `fp expr` for better readability and consistency with `fp var`.

blefloch:
Probably also relevant in all other modules (e.g. `intexpr` could be changed to `int expr`).
I have no opinion on whether to do that.

FrankMittelbach:
We are somewhat inconsistent here. Part of the code say `<fpexpr>` or `<intexpr>` part of it says `<floating point expression>` or `<integer expression>`. Not sure if we should normalize on the short or the long form. I updated to the short forms to see how it looks, but maybe always using the long forms is actually better.

josephwright:
I think we did this :)

muzimuzhi:
Seems not.  For example,
```
grep -rn -e '\\\(meta\|Arg\){integer expression}' l3kernel | wc -l
      46
```
and there are a bit more cases in sibling directories other than `l3kernel`.  I will try to provide a PR for this.

josephwright:
> Seems not. For example,
> 
> ```
> grep -rn -e '\\\(meta\|Arg\){integer expression}' l3kernel | wc -l
>       46
> ```
> 
> and there are a bit more cases in sibling directories other than `l3kernel`. I will try to provide a PR for this.

Ah, thanks: I remember @FrankMittelbach doing something, but perhaps it wasn't comprehensive

muzimuzhi:
> I remember @FrankMittelbach doing something, but perhaps it wasn't comprehensive

Yes it was #971.

PR #971, opened 2021-07-28
trial update for #968

FrankMittelbach:
fpexpr  and floating point expression -> fp expr
ditto for intexpr dimexpr skipexpr and muskipexpr

not sure which is better short or long.

There is also an issue with l3doc which I temp corrected in interface3 to see how it comes out. \meta does italic but \Arg uses meta inside \texttt so the two differ next to each other, not sure how this is best handled \Arg to use normalfont itshape or \meta us tt it as I did here?

### DIFF
--- a/l3kernel/doc/interface3.tex
+++ b/l3kernel/doc/interface3.tex
@@ -41,6 +41,14 @@ for those people who are interested.
 
 
 \documentclass[kernel]{l3doc}
+
+% fix for l3doc class
+\ExplSyntaxOn
+\DeclareDocumentCommand \meta { m }
+  { \texttt{ \__codedoc_meta:n {#1} } }
+\ExplSyntaxOff
+
+
 \listfiles
 
 \begin{document}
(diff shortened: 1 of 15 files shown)


Skillmon:
I'd prefer `\meta` to use `\texttt` (and that's what I do in my own documentations which use `l3doc`).

josephwright (commented):
Personally, I like 'intexpr, ,etc., as one word: 'int expr' feels more awkward as it's mixing two abbreviations for no real gain.

FrankMittelbach:
> Personally, I like 'intexpr, ,etc., as one word: 'int expr' feels more awkward as it's mixing two abbreviations for no real gain.

for intexpr I can sort of agree, but dimexpr does read awkward to me. But I don't mind either way, but at least we should be consistent and not sometimes say intexpr and in other places "integer expressions" etc (so if you you want to change back I would still go from my pull request because that normalized the other stuff as well).

PhelypeOleinik (approved):

josephwright (approved):

muzimuzhi:
(just to add a backlink for #968)

PR #975, opened 2021-08-06
implement more robust __tl_act:NNNn

Skillmon:
This PR implements a version of `\__tl_act:NNNn` that can work on arbitrary tokens in its fourth argument (earlier versions had the limitation that the argument could not contain `\s__tl_act_stop` at any group level).

There are still some minor limitations on the first three arguments (those are the same as they currently are): They must not contain `\__tl_act_result:n` and must not expand to `\__tl_act_output:n` or `\__tl_act_reverse_output:n` followed by `\__tl_act_result:n` at the same group level.

The result is considerably slower (small tests indicate a slowdown of at least factor two using `\tl_reverse:n`). And while this is more robust the question whether the limitation (not working on a single token which is only used inside that code, and only in a way it can't remain in the input stream for correct code due to `\exp:w` expansion) is that serious to justify this slowdown.

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -3178,118 +3178,146 @@
 %
 % \subsection{Token by token changes}
 %
-% \begin{variable}{\s_@@_act_stop}
-%   The \cs[no-index]{@@_act_\ldots{}} functions may be applied to any token list.
-%   Hence, we use a private quark, to allow any token, even quarks,
-%   in the token list.
-%   Only \cs{s_@@_act_stop} may not appear in the token lists manipulated by
-%   \cs{@@_act:NNNn} functions.
-%    \begin{macrocode}
-\scan_new:N \s_@@_act_stop
-%    \end{macrocode}
-% \end{variable}
-%
 % \begin{macro}[EXP]{\@@_act:NNNn}
 % \begin{macro}[EXP]{\@@_act_output:n, \@@_act_reverse_output:n}
-% \begin{macro}[EXP]{\@@_act_loop:w}
-% \begin{macro}[EXP]{\@@_act_normal:NwNNN}
-% \begin{macro}[EXP]{\@@_act_group:nwNNN}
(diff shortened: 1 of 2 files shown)


josephwright:
I'll see what @blefloch thinks, but as we are looking at quite a slowdown, perhaps this one is not such a great plan.

Skillmon:
Ran a few more tests with `\__tl_act:NNNn \use_none:n \use_none:n \prg_do_nothing:` on a token list containing 66 `N`-type tokens, 22 spaces, and 11 groups. Factor there is 2.349 times. I guess the most crucial spot for this slow down is the test for the end of the list. In the current implementation this is only done for the `group` case and there it is pretty fast because we just test for one token. With the new implementation that would be on every `N`-type token with `\tl_if_empty:nTF`.

PhelypeOleinik:
@Skillmon Could you please point out the main differences in behaviour between the two versions?

My two cents: I don't think allowing `\s__tl_act_stop` is worth a 2x slowdown (if that's the only thing this change addresses).  `\s__tl_act_stop` is an internal token, and if an user inserts that in the input, they deserve the consequences.

Skillmon:
@PhelypeOleinik yes, that's about it. It allows putting `\s__tl_act_stop` in the argument. When I started this I knew how I could make this work, but didn't know how big the slowdown would be. When I was done I knew, but thought I'd at least give the team the choice.

PhelypeOleinik:
Then I vote for no, sorry.  The whole kernel has lots of internal “forbidden tokens” that serve as markers here and there, so removing one from this list at the cost of a 2x slowdown doesn't seem worth it to me.

FrankMittelbach:
> Then I vote for no, sorry. The whole kernel has lots of internal “forbidden tokens” that serve as markers here and there, so removing one from this list at the cost of a 2x slowdown doesn't seem worth it to me.

I agree with that vote.

Skillmon:
@PhelypeOleinik nothing to be sorry about. It's a proposal I myself would have voted no on :)

PR #976, opened 2021-08-06
Faster tl_if_head_is_space:n

Skillmon:
This PR changes the `\tl_if_head_is_space:n` tests to use `\__kernel_tl_to_str:w`. This way the complex handling to remove the remainder of the list can be avoided, and since `\__kernel_tl_to_str:w` has no effect on spaces doesn't have any drawback, afaics.

My benchmarks indicate around 40% faster execution.

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -3147,31 +3147,23 @@
 %
 % \begin{macro}[EXP,pTF]{\tl_if_head_is_space:n}
 % \begin{macro}[EXP]{\@@_if_head_is_space:w}
-%   The auxiliary's argument is all that is before the first explicit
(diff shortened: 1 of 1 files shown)


blefloch (approved):
I think this is correct and there are definitely some (previously existing) tests, so that's ok.  I'm a little worried about performance for a token list consisting of control sequences with very long names, typical of expl3 code.  Could you benchmark this as a sanity check with argument equal to some typical chunk of code?

Skillmon:
@blefloch good intuition. Indeed if the argument consists out of a few lines of `expl3` code (I tested with the code of the function itself, 145 tokens in the stringified representation) and the current implementation is faster for such a case.

So I guess this shouldn't be merged -> closing.

PR #979, opened 2021-08-11
Automatically create releases from Tags with GH Actions

zauguin:
Questions I got while writing this:

- Do we want a E-Mail notification to be sent to latex3-commits for successful releases? Currently this is done since I would expect that everyone who is interested enough in LaTeX 3 development to subscribe to that list will also want to know about releases. But I would also assume that most readers of that list are aware of releases anyway, so it might be considered redundant.
- Should every pushed Tag trigger a release or should creating a release require explicit confirmation to avoid accidental releases when a tag is pushed? Currently every tag gets a release.


FrankMittelbach:
> * Do we want a E-Mail notification to be sent to latex3-commits for successful releases? Currently this is done since I would expect that everyone who is interested enough in LaTeX 3 development to subscribe to that list will also want to know about releases. But I would also assume that most readers of that list are aware of releases anyway, so it might be considered redundant.

there aren't that many. I wouldn't mind (on the contrary in fact I think the feedback is helpful)

> * Should every pushed Tag trigger a release or should creating a release require explicit confirmation to avoid accidental releases when a tag is pushed? Currently every tag gets a release.

maybe explicit confirmation is better, I don't know. We had one case so far where a release was not wanted but I'm not sure either way.

zauguin:
>     * Should every pushed Tag trigger a release or should creating a release require explicit confirmation to avoid accidental releases when a tag is pushed? Currently every tag gets a release.

I know changed it to require confirmation. We can easily change it later without any code change by changing the settings of the `Release` environment, so it's easy to get rid of if it turns out to be annoying.

PR #987, opened 2021-09-08
Register prefix markdown

Witiko:
I would like to register the `markdown` prefix for [the Markdown package](https://ctan.org/pkg/markdown) as a part of modernization efforts (https://github.com/Witiko/markdown/issues/96, https://github.com/Witiko/markdown/pull/77) that include partial rewrite in expl3.


Witiko:
@josephwright Thank you kindly!

PR #991, opened 2021-09-14
fix box_gresize_to_ht_plus_dp to be global

TeXhackse:
The macro `\box_gresize_to_ht_plus_dp:Nn` currently does not a global but a local assignment. It's as small as a typo but I'd really to love to get this one fixed. 

I was not sure if I also should touch the “updated” date for the macros. Please let me know If should have adjusted that one was well or if this is up to you.

Thanks a lot.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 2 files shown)


josephwright:
No change to the docs (as there is no change to the API), but could you add to the CHANGELOG?

TeXhackse:
Thansk for the quick reply.
I hope I understood that correctly. Added the entry.

PR #1005, opened 2021-10-26
Fallback color model

zauguin:
Replace the existing system of having a fallback color for every color model with the concept of fallback color models:

Whenever direct conversion between two color models isn't supported, try to convert through the fallback color model of one or both of the color models. Since the fallback is always one of the three device models and conversion between all of them is supported, this process will always lead to a valid conversion.

At this point, no functional change is intended. All conversions which didn't exists previously have been replaced by explicit conversions to the previous fallback color. (That being said there is a functional change fixing a IMO wrong conversion to RGB from CIE backed separations since we previously converted to white instead of black)

Currently this doesn't provide useful functionality, but it allows future extensions. Especially we can add a user interface to overwrite the conversion to and from the fallback model and we can add better default fallbacks for ICCBased color models.


josephwright (approved):

l3kernel/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ this project uses date-based 'snapshot' version identifiers.
 ## [Unreleased]
 
 ### Changed
+- Allow indirect conversions between colorspaces though fallback models
```
dbitouze, line 11: "through fallback models"?

l3kernel/l3color.dtx
```diff
@@ -637,31 +637,61 @@
...
 %     \@@_convert_rgb_cmyk:w
 %   }
 %  \begin{macro}[aux, EXP]{\@@_convert_rgb_cmyk:nnnn}
-%    Model conversion is carried out using standard formulae, as described in
-%    the manual for \pkg{xcolor} (see also the \emph{PostScript Language
-%    Reference Manual}).
+%    Model conversion is carried out using standard formulae for base models,
+%    as described in the manual for \pkg{xcolor} (see also the \emph{PostScript
+%    Language Reference Manual}). For other models direct conversion might not
+%    be defined, so we go throught the fallback models if necessary.
```
dbitouze, line 654: "through the fallback models if necessary"

PR #1012, opened 2021-11-01
l3prop: use \keyval_parse:nnn for ..._from_keyval:Nn

Skillmon:
This PR will change the `_from_keyval:Nn` functions of the `l3prop` module to use `\keyval_parse:nnn` instead of the more limited implementation in the `l3prop` module. This will result in the property list functions supporting active equals and commas. Since usage on stackexchange indicates that the property list functions are also used to parse user input (see e.g. [here](https://tex.stackexchange.com/a/418021/117050) and [here](https://tex.stackexchange.com/a/589159/117050)) and not only as an interface at the programming level this seems like a good idea.

As a bonus, this will result in a speed increase of roughly 30 %.

I've made a minor change to an internal of `l3keys` (the macro which would throw an error on blank key names) to allow those as long as `\l__keyval_allow_blank_keys_bool` is true, which I set inside the `l3prop` functions. One could consider to make this boolean public (and if so, one could also consider turning this into a flag and using `\ifodd` on its height to allow expandably turning it off/on). This change will only slow down this case and not affect normal performance for non-blank key names. If a behaviour change would be ok, I'd argue to disallow blank key names in `l3prop`'s `..._from_keyval:Nn` functions. In that case 542067d should be reverted and the changes to the documentation in 0f6e10f might need some more work.

A side effect is that some of the error messages are formatted differently:

- `\keyval_parse:nnn` will throw an expandable error message on superfluous equals signs
- if `\tex_expanded:D` is available, missing equals sign errors will always come *after* every superfluous equals sign errors
- the missing braces errors will not include the user input verbatim but only after spaces and outer braces are stripped.

I also changed the `\prop_concat:NNN` function to use `\prop_map_inline:Nn` instead of `...map_tokens:Nn`, this is (considerably) slower if the last property list argument contains few entries (<5) but faster starting at around 5 elements. For many elements the difference becomes insignificant again. So that commit could be reverted.

### DIFF
--- a/l3kernel/l3prop.dtx
+++ b/l3kernel/l3prop.dtx
@@ -804,16 +804,7 @@
 % \begin{macro}{\prop_const_from_keyval:Nn, \prop_const_from_keyval:cn}
 % \begin{macro}{\prop_put_from_keyval:Nn, \prop_put_from_keyval:cn}
 % \begin{macro}{\prop_gput_from_keyval:Nn, \prop_gput_from_keyval:cn}
-% \begin{macro}
-%   {
-%     \@@_from_keyval:n,
-%     \@@_from_keyval_loop:w,
-%     \@@_from_keyval_split:Nw,
-%     \@@_from_keyval_key:n,
-%     \@@_from_keyval_key:w,
-%     \@@_from_keyval_value:n,
-%     \@@_from_keyval_value:w
-%   }
+% \begin{macro}{\@@_missing_eq:n}
 %   To avoid tracking throughout the loop the variable name and whether
 %   the assignment is local/global, do everything in a scratch variable
 %   and empty it afterwards to avoid wasting memory.  Loop through items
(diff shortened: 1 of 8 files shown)


l3kernel/l3prop.dtx
```diff
@@ -832,80 +823,48 @@
...
   }
 \cs_generate_variant:Nn \prop_const_from_keyval:Nn { c }
-\cs_new_protected:Npn \prop_put_from_keyval:Nn #1#2
+\cs_new_protected:Npn \prop_put_from_keyval:Nn
   {
-    \prop_set_eq:NN \l_@@_internal_prop #1
-    \@@_from_keyval:n {#2}
-    \prop_set_eq:NN #1 \l_@@_internal_prop
-    \prop_clear:N \l_@@_internal_prop
+    \bool_if:NTF \l__keyval_allow_blank_keys_bool
```
blefloch, line 847: I'm surprised that `l3doc` wouldn't complain here about a foreign internal being used.

PhelypeOleinik: It probably does, but it's a warning.  Maybe that boolean should be `__kernel`?

Skillmon: @PhelypeOleinik seems reasonable. Should I push here or open a new PR?

blefloch:
> @PhelypeOleinik <https://github.com/PhelypeOleinik> seems reasonable. Should I
> push here or open a new PR?

I don't know how easy it is to pick up a commit from an already merged pull
request, so separate PR please if that's not a hassle.

PR #1023, opened 2021-11-17
LuaMetaTeX/LMTX compatibility

zauguin:
Adds support for some of the name changes in LuaMetaTeX. It assues that it runs in a format which provides most of the legacy aliases, but fixes a few remaining issues.

This makes expl3 work with LuaMetaLaTeX and almost with lmtx. The almost refers to file lookup paths, lmtx doesn't find the Unicode data files unless I copy them to the current directory. But since lmtx is basically a separate distribution I'm assuming that that's just an issue of how this is installed.


josephwright (approved):

josephwright:
Looks good to me

PR #1028, opened 2021-11-25
Add support for validity scope for keys

josephwright:
This PR covers the `expl3` part of the plans for option handling. All internal l3keys work is covered here at present.

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -687,16 +687,16 @@
 % way, \pkg{l3keys} allows this information to be specified using the
 % \texttt{.validity:n} property.
 %
-% \begin{function}[added = 2021-11-26]{.usage:n}
+% \begin{function}[added = 2022-01-10]{.validity:n}
 %   \begin{syntax}
-%     \meta{key} .usage:n = \meta{scope}
+%     \meta{key} .validity:n = \meta{scope}
 %   \end{syntax}
 %   Defines the \meta{key} to have usage within the \meta{scope}, which
 %   should be one of \texttt{general}, \texttt{preamble} or \texttt{load}.
 % \end{function}
 %
-% \begin{variable}[added = 2021-11-29]
-%   {\l_keys_usage_load_prop, \l_keys_usage_preamble_prop}
+% \begin{variable}[added = 2022-01-10]
+%   {\l_keys_validity_load_prop, \l_keys_validity_preamble_prop}
 %   \pkg{l3keys} itself does \emph{not} attempt to redefine keys based on the
 %   validity scope. Rather, this information is made available with these
 %   two property lists. These hold an entry for each module (prefix); the
@@ -1650,11 +1650,11 @@
 %    \end{macrocode}
 % \end{variable}
 %
-% \begin{variable}{\l_keys_usage_load_prop, \l_keys_usage_preamble_prop}
+% \begin{variable}{\l_keys_validity_load_prop, \l_keys_validity_preamble_prop}
 %   Global data for document-level information.
 %    \begin{macrocode}
-\prop_new:N \l_keys_usage_load_prop
-\prop_new:N \l_keys_usage_preamble_prop
+\prop_new:N \l_keys_validity_load_prop
+\prop_new:N \l_keys_validity_preamble_prop
 %    \end{macrocode}
 % \end{variable}
 %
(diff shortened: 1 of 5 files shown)


davidcarlisle (approved):
looks like we discussed, although haven't really tested until the variaous parts come together.

l3kernel/l3keys.dtx
```diff
@@ -679,6 +679,29 @@
...
+% until typesetting begins. To allow formats to support this in a structured
+% way, \pkg{l3keys} allows this information to be specified using the
+% \texttt{.validity:n} property.
+%
+% \begin{function}[added = 2021-11-25]{.validity:n}
+%   \begin{syntax}
+%     \meta{key} .validity:n = \meta{scope}
+%   \end{syntax}
+%   Defines the \meta{key} to have validty within the \meta{scope}, which
+%   should be one of \texttt{general}, \texttt{preamble} or \texttt{load-time}.
```
FrankMittelbach, line 695: would suggest just "load" or maybe "loadtime" (but then it isn't preambletime either) be not with a hyphen ... however no strong feelings

l3kernel/l3keys.dtx
```diff
@@ -679,6 +679,29 @@
...
+% \end{function}
+%
+% \begin{variable}{\g_keys_load_time_seq, \g_keys_preamble_seq}
```
FrankMittelbach, line 698: same here (load or loadtime) and the same goes for other places obviously

l3kernel/l3keys.dtx
```diff
@@ -2559,6 +2630,28 @@
 % \end{macro}
 % \end{macro}
 %
+% \subsection{Key properties for \LaTeXe{} options}
+%
```
FrankMittelbach, line 2634: could do with a few words on where they are supposed to be used, but perhaps that is too early if we haven't decided on the final name for `\SetKeys`

FrankMittelbach (approved):
minor comments

car222222:
Not sure if this the right place for a comment on the name "validity"?

Just to say that "usage" is a more mundane (and hence perhaps more widely understood) alternative. 

(I think that the value "loadtime" (or "atload", "load-time") is a better companions to "general", "preamble".  
Is "general" too vague?)

josephwright:
I don't plan to merge immediately here: we probably want to refine some naming. What do others think about @car222222's `.usage:n` idea? On `general`, it could be `global` or `unscoped` or `open-ended` or `free`. I hyphenated `load-time` as that reads better to me: I did wonder about `single` (which would go nicely with `.usage:n`).

josephwright:
I've set as a draft so we can discuss here without a merge being carried out accidentally too early. (I'll squash everything once we have a solid plan.)

l3kernel/l3keys.dtx
josephwright: Yes, I'm not sure if this even lives in `l3keys`: we could have some method for aliasing property names, which would then allow it to be handled in `ltkeys` (sill to be written).

car222222:
"single" is OK if it is sufficiently precise: i.e., is it clear when the "single use" must happen?

u-fischer:
I don't like `single` very much. The point is not that the key can be used only once, but that it can be used only in the package option when the package is loaded and then it is disabled (so even if it hasn't been used at all). So imho some word with "load" is better (and imho no hyphen, even if it reads better: I would always have to look up if it is `load-time` or `load_time`).

FrankMittelbach:
> I don't like `single` very much. The point is not that the key can be used only once, but that it can be used only in the package option when the package is loaded and then it is disabled (so even if it hasn't been used at all). So imho some word with "load" is better (and imho no hyphen, even if it reads better: I would always have to look up if it is `load-time` or `load_time`).

I agree. I'm perfectly happy with "general", "preamble" and "load" (short and snappy and not imprecise).
I'm ok with "validity"  but "usage" is also fine and perhaps better. So my vote would be for "usage=load".

l3kernel/l3keys.dtx
FrankMittelbach: yes, should perhaps moved there (when written)

josephwright:
I have a suggestion for separating out the '`expl3`' handlers from the 'document level' ones. I could set up a simple switching interface that would swap out the unique csname part used to find handlers. That would let the two sets co-exist without being available at the same time: how does that sound?

Also, at present I'm assuming that `l3keys` and `ltkeys` both use `keys` as their prefix: that lets `ltkeys` get at `l3keys` internals - OK?

l3kernel/l3keys.dtx
```diff
@@ -3249,6 +3358,7 @@
       { \prg_return_true: }
       { \prg_return_false: }
   }
+\prg_generate_conditional_variant:Nnn \keys_if_exist:nn { nx } { T , F , TF }
```
Skillmon, line 3361: Should be `ne` as variant instead of `nx`, imho. `\keys_if_exist:nn` is expandable, and even if the kernel does not yet strictly require `\expanded` it is in practice almost always available.

l3kernel/l3keys.dtx
```diff
@@ -679,6 +679,31 @@
...
+%
+% \begin{function}[added = 2021-11-26]{.usage:n}
+%   \begin{syntax}
+%     \meta{key} .usage:n = \meta{scope}
+%   \end{syntax}
+%   Defines the \meta{key} to have usage within the \meta{scope}, which
+%   should be one of \texttt{general}, \texttt{preamble} or \texttt{load}.
+% \end{function}
+%
+% \begin{variable}{\g_keys_usage_load_prop, \g_keys_usage_preamble_seq}
```
Skillmon, line 698: Currently the documented names mismatch the implemented ones (starting in line 1653).

l3kernel/l3keys.dtx
josephwright: Well ... that's complicated as they handle `#` differently, so we can never drop `x`-type even where the underlying code is expandable. I'll think about this one.

l3kernel/l3keys.dtx
josephwright: Fixed. Opinion on whether I should go with two `prop`, one `clist` and one `prop` or the `prop`/`seq` split as now?

Skillmon: For programming a `seq` is better than a `clist` (less overhead in mapping items). And since the load time options are package/module dependent a `prop` seems like a good choice. So my opinion is pretty much stick with the current variables.

l3kernel/l3keys.dtx
Skillmon: But that's a new addition so you can pick the one you want now. And the handling of data which should be turned into a string is more natural using `e`-type expansion, imho, because of the `#` handling.

l3kernel/l3keys.dtx
FrankMittelbach: that very much depends on the viewpoint I would say. For expl3 based packages I agree but for 2e type packages a seq or a prop is not that helpful and we intend to cater for both types of class/packages here. My inclination would be therefore to go with clists (but it is after midnight so who knows what I think tomorrow). Overhead in processing is not really that much of an argument given that we talk about a handful of packages per doc with a handful of keys in the their lists.

josephwright:
Could we get a decision here as it links to the 2e options handling?

l3kernel/l3keys.dtx
```diff
@@ -679,6 +679,31 @@
...
+% of validity. Some will be only available once, others will only be valid
+% until typesetting begins. To allow formats to support this in a structured
+% way, \pkg{l3keys} allows this information to be specified using the
+% \texttt{.validity:n} property.
+%
+% \begin{function}[added = 2021-11-26]{.usage:n}
+%   \begin{syntax}
+%     \meta{key} .usage:n = \meta{scope}
+%   \end{syntax}
+%   Defines the \meta{key} to have usage within the \meta{scope}, which
```
FrankMittelbach, line 694: I think usage is far to general to make make much sense if you look over code (not knowing it); from that perspective I think validity is better but I can live with both. But the documentation should match: it still talks about .validity:n above

l3kernel/l3keys.dtx
```diff
@@ -2547,6 +2636,13 @@
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}{.validity:n}
```
FrankMittelbach, line 2639: same here, or do I read the diff incorrectly

FrankMittelbach:
> I have a suggestion for separating out the '`expl3`' handlers from the 'document level' ones. I could set up a simple switching interface that would swap out the unique csname part used to find handlers. That would let the two sets co-exist without being available at the same time: how does that sound?

when would that switching happen and how often? I'm not sure I see how this could be done reliablly without a lot of tracking, but perhaps I'm mistaken

> Also, at present I'm assuming that `l3keys` and `ltkeys` both use `keys` as their prefix: that lets `ltkeys` get at `l3keys` internals - OK?

I would think so I don't think we need to invent further prefixes here.

l3kernel/l3keys.dtx
FrankMittelbach: if the `\g_keys_usage_preamble_seq` is in any way made available to 2e package (ie not under its expl3 name) then I think a clist is better for it. But perhaps there are no good use case why that should be offered.

l3kernel/l3keys.dtx
josephwright: I've gone for `.validity:n`

l3kernel/l3keys.dtx
josephwright: It's no longer a `clist`: a `prop` for each area makes most sense.

josephwright:
> > I have a suggestion for separating out the '`expl3`' handlers from the 'document level' ones. I could set up a simple switching interface that would swap out the unique csname part used to find handlers. That would let the two sets co-exist without being available at the same time: how does that sound?
> 
> when would that switching happen and how often? I'm not sure I see how this could be done reliablly without a lot of tracking, but perhaps I'm mistaken

I'd have to use two different prefixes for the two types of property. It feels like we might run into issues with nesting, so I think on balance I'd like to avoid it.

PR #1031, opened 2022-01-10
Refactor peek space/filler removal

josephwright:
This PR deprecates `\peek_<test>_ignore_spaces:n`, documents `\peek_remove_spaces:n` and introduces `\peek_remove_filler:n`. It retains `\peek_<test>:N(TF)` and `\peek_<test>_remove:N(TF)` as they need us to insert `\group_align_safe_(begin|end):` in the correct places.

We likely need to provide some documentation for updating code:
currently we do not have an appropriate file.


l3kernel/l3token.dtx
```diff
@@ -974,20 +904,34 @@
...
+%   removes the token and checks the next token. If neither of these cases
+%   apply, expands the next token using \texttt{f}-type expansion, then checks
+%   the resulting leading token in the same way. If after expansion the next
+%   token is neither of the two test cases, the \meta{code} will be inserted
+%   into the input stream. Typically this will contain a \texttt{peek}
+%   operation, but this is not required.
+%   \begin{texnote}
+%     This is essentially a macro-based implementation of how \TeX{} handles
+%     the search for a left brace after for example \tn{everypar}, except that
+%     any non-expandable token ends the \meta{filler}.
```
FrankMittelbach, line 933: ... without causing an error as the \TeX{} operation would  in that case (or something like that).

FrankMittelbach (approved):

l3kernel/l3token.dtx
```diff
@@ -974,20 +904,34 @@
...
-%   Tests if the next non-space \meta{token} in the input stream has the
-%   same meaning as the \meta{test token} (as defined by the test
-%   \cs{token_if_eq_meaning:NNTF}).  Explicit and implicit space tokens
-%   (with character code 32 and category code 10) are ignored and
-%   removed by the test and the \meta{token} is removed from the
-%   input stream if the test is true. The function then places
-%   either the \meta{true code} or \meta{false code} in the input stream
-%   (as appropriate to the result of the test).
+%   Peeks ahead and detect if the following token is a space (category code
+%   $10$ and character code $32). If so, removes the token and checks the
```
FrankMittelbach, line 912: missing $ causing the doc fail

FrankMittelbach (changes requested):

josephwright:
I'll fix the failing build on main: it's an issue with the Lua scripting not the TeX code

blefloch:
Quick comment, not a full review for now: searching for catcode 10 charcode 32 is wrong.  I think it should just be catcode 10.

josephwright:
> Quick comment, not a full review for now: searching for catcode 10 charcode 32 is wrong. I think it should just be catcode 10.

It's the same in `\peek_remove_spaces:n` - 
```latex
\if_meaning:w \l_peek_token \c_space_token
````

**Comment by josephwright** [comment]
Good to go here? I think once this one lands there is a release due.

File: l3kernel/l3token.dtx [hunk 3]
```diff
@@ -974,20 +904,35 @@
...
+%   inserted into the input stream. Typically this will contain a \texttt{peek}
+%   operation, but this is not required.
+% \end{function}
+%
+% \begin{function}[added = 2022-01-10]{\peek_remove_filler:n}
+%   \begin{syntax}
+%     \cs{peek_remove_filler:n} \Arg{code}
+%   \end{syntax}
+%   Peeks ahead and detect if the following token is a space (category code
+%   $10$ and character code $32$) or has meaning equal to \cs{scan_stop:}. If so,
```
blefloch, line 923: Here it should certainly be just a catcode test to match what TeX does (the code below is correct).  On the other hand `\peek_remove_spaces:n` should remain unchanged because the use case is to ignore actual spaces in the file.

l3kernel/l3token.dtx
```diff
@@ -2914,6 +2859,57 @@ end
...
+    \else:
+      \if_meaning:w \l_peek_token \scan_stop:
+        \exp_after:wN \exp_after:wN \exp_after:wN
+          \@@_true_remove:w
+      \else:
+        \exp_after:wN \exp_after:wN \exp_after:wN
+          \@@_remove_filler_expand:w
+      \fi:
+    \fi:
+  }
```
blefloch, line 2900: Maybe add a comment specifying that this nesting of conditionals is safe even if `\l_peek_token` is a TeX conditional (or an outer macro), because `\l_peek_token` is only skipped over if it is a space.

l3kernel/l3token.dtx
```diff
@@ -2914,6 +2859,57 @@ end
...
+\cs_new_protected:Npn \@@_remove_filler_expand:w
+  {
+    \token_if_expandable:NTF \l_peek_token
+      { \@@_remove_filler:w }
+      { \@@_false:w }
+  }
```
blefloch, line 2906: `\token_if_expandable:NTF` (wrongly in my opinion) rejects undefined control sequences as being non-expandable.  This means the `remove_filler` code will stop when seeing an undefined control sequence.  Please document this edge case.

l3kernel/testfiles/m3peek001.lvt
```diff
@@ -205,14 +181,21 @@
...
     \tex_lowercase:D { \peek_remove_spaces:n { \tex_show:D } ~ }
   }
 
+
+\TEST { \peek_remove_filler:n }
+  {
+    \tl_set:Nx \l_tmpa_tl { \use:n { \scan_stop: \c_space_tl } ~ \scan_stop: } 
+    \use:n { \peek_remove_filler:n { \tex_show:D } ~ \c_space_token } ~ \c_space_tl
+      \l_tmpa_tl _
+  }
```
blefloch, line 198: Can you also add a test for `\exp_not:N`, like `\peek_remove_filler:n { \show } \exp_not:N \undefined \relax X`?  I think the code will correctly treat `\exp_not:N \undefined` as `\relax` but I don't remember.

blefloch (changes requested):
Mostly good but see a few comments.

l3kernel/l3token.dtx
FrankMittelbach: agreed

l3kernel/l3token.dtx
FrankMittelbach: @blefloch why do you think that it is wrong to stop scanning if you hit an undefined csname? It isn't a filler part that would be skipped over in the TeX sense and I can't think of a use case where you would want to skip over undefined csnames. Or am I miss something?

josephwright: I've documented but I don't see how one could describe an undefined control sequence as expandable: `\expandafter\foo\undefined` is an error.

PhelypeOleinik: I think the point here is that while scanning for a filler, TeX will expand the undefined cs, throw an (expandable) error, and proceed, so `\expanded \foo {a}` expands to just `a` plus the error.  I'd expect `\peek_remove_filler:n {} \foo a` to do the same.

josephwright: I've adjusted the code so we get the TeX error.

l3kernel/testfiles/m3peek001.lvt
josephwright: I've tested and we _don't_ currently allow `\exp_not:N` here. That happens because we have a two-step process, using `\romannumeral` to deal with expansion and then `\futurelet` to deal with token testing. To allow `\exp_not:N \foo` we'd have to do everything by hand. I think that's not worth it, but I am ready to hear counter-arguments!

blefloch: You are right, I thought that the token remained under the effect of `\noexpand` long enough to be detected (cf how `\exp_after:wN \peek_after:Nw \exp_after:wN ? \exp_not:N \foo` sets `\l_peek_token` to a weird version of `\relax`).  But it seems we cannot do so.

PR #1035, opened 2022-01-26
Add \color_math:nn(n)

josephwright:


josephwright:
I've implemented to try to use as much of the recent peek stuff as possible. I've also tried to abstract the limits and prime code: hopefully this works OK.

josephwright:
Do we want to hard-code the `\limits` stuff by testing for `\tex_[<>|display|no]limits:D`? That would leave only the prime business as a configuration point.

FrankMittelbach:
> Do we want to hard-code the `\limits` stuff by testing for `\tex_[<>|display|no]limits:D`? That would leave only the prime business as a configuration point.

I see no reason why not. The 2e \mathcolor is also restricted to that and if we ever want to make it more flexible we could still do that later in an upward compatible way, but I doubt that there is a need for it.

FrankMittelbach (approved):

PR #1036, opened 2022-01-27
Build l3doc.pdf (fixes #384)

### LINKED ISSUE #384: make l3doc available via texdoc
At the moment `texdoc l3doc` opens `expl3.pdf` but it would be nice to have the manual of the class available via this command. I’m not sure if this can be solved by the L3 team or if its a thing for the `texdoc` team. However as far as I see there is only `l3doc.dtx` but no `*.pdf` on CTAN.

I know that `l3doc` is kind of tacky and of course I can compile the manual from the `*.dtx` myself …

mgkurtz:


FrankMittelbach:
At this point in time we do not really want to actively encourage the use of l3doc outside the kernel sources, because we think a lot of it inside has to change and partly in incompatible ways. This is part of the reason why we do not distribute a manual for it. There is a note to this effect in the docs but perhaps this should be made more prominent.

mgkurtz:
Ok, fine. Thanks for the answer. I was just irritated, that issue #384 at some point diverged and there was no decision whether to actually distribute l3doc.pdf (including the warning) or not at all. So I assume l3doc shall not be further promoted and issue #384 shall be stalled until the [l3doc project](https://github.com/latex3/latex3/projects/2) has made enough progress?

Can you suggest, what to use in the meantime for documenting LaTeX3 code? Besides of course waiting or helping out in the l3doc project ☺

FrankMittelbach:
well, you can of course use it as is, all I'm saying that it has some issues and things we don't like, but the priority for changing it is low (as the effort is high compared to the gains), but eventually it will probably change and anything incompatible might mean you need to follow if we update. It has some nice features compared to a simple doc file so you might be prepared to live with that caveat. 

@josephwright maybe we should just distribute it (perhaps with a slighlty bolder warning) and blaim everything on @wspr  :-)

josephwright (changes requested):
Could we have a ChangeLog entry, and as @FrankMittelbach says something added to the source to say 'this may change'.

mgkurtz:
> Could we have a ChangeLog entry, and as @FrankMittelbach says something added to the source to say 'this may change'.

I have added both now. Though, I am unsure, whether they are fine like that.

Hopefully the ci test should also work now.

FrankMittelbach:
> I have added both now. Though, I am unsure, whether they are fine like that.

it should say class not package in the note I guess.
 
> Hopefully the ci test should also work now.

we'll see, I started it it, it is not enabled automatically for non-team-members it seems.

PR #1048, opened 2022-02-10
f-type expand key names

josephwright:


josephwright:
@Skillmon Fancy taking a look?

josephwright:
Notes

- This is much slower than the old code :( Suggestions on better approaches quite welcome
-  I've not covered the case where `f`-type expansion leads to a macro which starts with a space then an expandable token, as that gets complicated and is likely pathological
- As we use f-type expansion, one might still have e.g. `a\b` as a key name - not sure how we document, etc., this

PhelypeOleinik:
I would argue that a less cautious approach would work in this case:  1) grab the next token (leading spaces are gone without testing), then 2) test if it's expandable (we have to test, otherwise we get stuck in a loop); 3) if it is expandable, use the trick in `\@@_loop:Nw` to expand ahead, then go back to 1.  4) If the grabbed token is not expandable, then grab up to the `=` and make a key name (it could use the same `\cs:w ... \cs_end:` normalisation done in file names), and proceed as normal.

This would be (I think) rather fast, because it only has to grab one (group of) token ahead, and expand it.  It would cope for the second point in Joseph's comment above, because it would only stop at a truly unexpandable token. A case like `a\b` would still not work for some definitions of `\b`, but I think it's okay to tell users to behave sometimes :)

In step 1 of the process above, we would lose braces in a key-val pair like `{key}name = value`.  If keeping braces is wanted here (weird input, but :), step 1 could also do `\tl_if_head_is_group:nTF` (which is quite a bit faster than `\tl_if_head_is_space:nTF`, if I recall correctly) to account for that.

Just a late-night rambling :)  Tomorrow I can write actual code if this approach makes sense

josephwright:
@PhelypeOleinik 

To expand on

>     * As we use f-type expansion, one might still have e.g. `a\b` as a key name - not sure how we document, etc., this

The only way I could think of to do this would be to use a slow 'action' mapping, taking each token and expanding as required up the first `=`, then grabbing without expansion to the `,` and so on. That would be similar to `\text_expand:n` or similar: slow to the point we couldn't really use it. And I'm not sure it's really what @FrankMittelbach was after in the first place :)

josephwright:
@PhelypeOleinik 

> I would argue that a less cautious approach would work in this case: 1) grab the next token (leading spaces are gone without testing), 

That will strip braces, which is something we have to be super-cautious about in a keyval parser. What happens for example with `{a,b} = c`, which is not entirely unreasonable (it might be some sensible phrase for the context).

josephwright:
@PhelypeOleinik 

> 2) test if it's expandable (we have to test, otherwise we get stuck in a loop); 3) if it is expandable, use the trick in `\@@_loop:Nw` to expand ahead, then go back to 1. 4) If the grabbed token is not expandable, then grab up to the `=` and make a key name (it could use the same `\cs:w ... \cs_end:` normalisation done in file names), and proceed as normal.

I had hoped to avoid tests entirely and use an approach similar to `\tl_trim_spaces:n` to remove the leading spaces, but as that code is rEXP (without putting it inside and `\expanded`), and as there are not _that_ many expansions inside `\tl_if_head_is_space:nTF`, I decided to use a private copy of the latter ideas.

I guess one thing that didn't get agreed is a spec for protected macros. Using an `f`-type expansion they get expanded, which is much the same as say `\begin{\macro}` as `c`-type expansion also ignores `\protected` status. Testing for 'expandability' could cut either way here, depending on exactly how one wants to handle macros. But that feels like it could get a little slow.

josephwright:
> Just a late-night rambling :) Tomorrow I can write actual code if this approach makes sense

Go for it :) The really tricky part of reworking is handling the change in the comma loops, rather than the macro expansion itself.

Skillmon:
> Go for it :) The really tricky part of reworking is handling the change in the comma loops, rather than the macro expansion itself.

Sorry for writing such complicated comma loops. As an aside: I'm not entirely sure how to get best performance out of the above suggestions. But note that since `\keyval_parse:nnn` has restricted input (there is at least one token that's forbidden in the input and that is `\s__keyval_tail`) we can use that to write a fast space-token test via argument grabbing logic (that would then be considerably faster than `\tl_if_head_is_group:nTF`; there are already similar macros with `\__keyval_if_empty:w` etc.).

> @Skillmon Fancy taking a look?

Will do, but not sure when I manage to :)

josephwright:
> > Go for it :) The really tricky part of reworking is handling the change in the comma loops, rather than the macro expansion itself.
> 
> Sorry for writing such complicated comma loops. As an aside: I'm not entirely sure how to get best performance out of the above suggestions. But note that since `\keyval_parse:nnn` has restricted input (there is at least one token that's forbidden in the input and that is `\s__keyval_tail`) we can use that to write a fast space-token test via argument grabbing logic (that would then be considerably faster than `\tl_if_head_is_group:nTF`; there are already similar macros with `\__keyval_if_empty:w` etc.).

Like I said to @PhelypeOleinik, I did try that using a version of the existing space-trimming code, but as it's based on `f`f-type expansion, it gets 'interesting', and using a test seemed easier to implement and discuss. (We can after all look at performance later if the feeling is the API is right.)

PhelypeOleinik:
I pushed a commit with a different implementation.  If that doesn't look right, feel free to revert the commit (I thought it would be easier than another branch).  The implementation goes more or less as I described in my earlier comment, except that

> That will strip braces, which is something we have to be super-cautious about in a keyval parser. What happens for example with `{a,b} = c`, which is not entirely unreasonable (it might be some sensible phrase for the context).

I failed to realise half the problem with stripping braces, so my implementation ended up being just an extension to Joseph's.

The code now does, more or less: (1) `f`-expand ahead; (2) look at the first token in the argument and do a three-way branching (as usual: space, group, and otherwise). (3.1) If a space is found, gobble and go back to (2); (3.2) If a `{` is found, do the usual key-val parsing; (3.3) otherwise, check if the following token is expandable; if it is, go to (1), else do the key-val parsing.

Since the code does a lot more, it is of course slower (and as fast as my optimisation abilities go).  I benchmarked the current keyval parser to around 70 ops, @josephwright's implementation to 142 ops, and mine to 156 ops.  It's even slower, but it expands as far as possible, so this has no problem regarding a macro that expands to a space and another expandable token (which I think is worth the extra cycles).

I think the code is slower in part because it doesn't really integrate with the key-val parser, rather I think it makes a two-step process, so likely it could be made faster with proper integration (for example, some space-trimming is no longer needed, since we are guaranteed to have no leading spaces in key names).

P.S.: The only failed test is `m3keyval001`, because now the code expands the two leading active commas in `\keyval_active_spaces`, rather than just one, giving one more Undefined control sequence error. I'm inclined to leave this be, but a test for active chars could be added.

P.P.S.: No idea why Actions failed on the documentation.  Locally it builds fine.

Skillmon:
@PhelypeOleinik regarding your PS: Imho no leading active commas should ever be expanded! A leading active comma means blank item, blank items are skipped. Active commas and commas of category other should be treated identically, because that's what we documented.

josephwright:
On active commas - they could occur but we should look to get them set up like UTF-8 actives (`\ifincsname` + `\protected`): I'd not worry about that rather special case.

josephwright:
(I mean specifically leading active commas)

josephwright:
I'm happy with the code: I think we need to see if @FrankMittelbach is happy with the results :)

lvjr:
If there are some worries about the speed impact, how about using a different function for this new feature? Then package writers can decide which one to use.

Skillmon:
This has also potential to break `l3prop`, as the `_from_keyval` functions there are currently using `\keyval_parse:nnn`, and with this the property names are no longer strings.

josephwright:
> If there are some worries about the speed impact, how about using a different function for this new feature? Then package writers can decide which one to use.

Performance here is more a concern in the sense that any more complex approach is slower than the original (and `l3keys` is already a little slower than say `expvk` let alone `keyval`): something like 'professional pride'. In reality, that will not significantly impact users - what will more is having different behaviours in difference keys. @FrankMittelbach is keen we bring keyvals into line with other arguments one sees in documents, where almost always things can be saved in a macro. That means having two different approaches fails to address the desired outcome.

josephwright:
> This has also potential to break `l3prop`, as the `_from_keyval` functions there are currently using `\keyval_parse:nnn`, and with this the property names are no longer strings.

I will need to think about that

Skillmon:
Reverting https://github.com/latex3/latex3/pull/1012 would be an option (though `l3prop` would then be category code fragile against active commas or equals signs again).

josephwright:
We have reflected here and decided against this change.

PR #1075, opened 2022-03-26
Support space tokens in \char_generate:nn

zauguin:
Pretty much what the title says... Add LuaTeX support by injecting token userdata instead of relying on `cprint` to avoid changing the charcode to 32, for other engines it just removes the check.

### DIFF
--- a/l3kernel/l3token.dtx
+++ b/l3kernel/l3token.dtx
@@ -1529,24 +1529,22 @@
 %    \begin{macrocode}
 \cs_new:Npn \@@_generate_aux:w #1 ; #2 ;
   {
-    \if_int_odd:w 1
-        \if_int_compare:w #2 = 10 \exp_stop_f: \else: 0 \fi:
-        \if_int_compare:w #1 = \c_zero_int     \else: 0 \fi: \exp_stop_f:
-      \msg_expandable_error:nn { char } { null-space }
+    \if_int_odd:w 0
+        \if_int_compare:w #2 < 1  \exp_stop_f: 1 \fi:
+        \if_int_compare:w #2 = 5  \exp_stop_f: 1 \fi:
+        \if_int_compare:w #2 = 9  \exp_stop_f: 1 \fi:
+        \if_int_compare:w #2 > 13 \exp_stop_f: 1 \fi: \exp_stop_f:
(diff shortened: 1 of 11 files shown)


l3kernel/l3token.dtx
```diff
@@ -1527,12 +1529,10 @@
...
-    \if_int_compare:w #2 = 10 \exp_stop_f:
-      \if_int_compare:w #1 =  \c_zero_int
-        \msg_expandable_error:nn { char } { null-space }
-      \else:
-        \msg_expandable_error:nn { char } { space }
-      \fi:
+    \if_int_odd:w 1
+        \if_int_compare:w #2 = 10 \exp_stop_f: \else: 0 \fi:
+        \if_int_compare:w #1 = \c_zero_int     \else: 0 \fi: \exp_stop_f:
+      \msg_expandable_error:nn { char } { null-space }
```
zauguin, line 1535: There's probably a nicer and/or faster way to check this(?)

blefloch: Would `\if_int_compare:w #2#1 = 100 \exp_stop_f:` work?  (I'm assuming that `#1` and `#2` are already made of digits at this point.)  The other way around `#1#2=10` seems more risky because it can also be true for catcode 0 and charcode 1, which may or may not be filtered elsewhere, I don't know from this snippet.

zauguin: > Would `\if_int_compare:w #2#1 = 100 \exp_stop_f:` work?

AFICT it does. Thanks! 

> I'm assuming that `#1` and `#2` are already made of digits at this point.

That's fair, they come from `\int_value:w`.

> The other way around `#1#2=10` seems more risky because it can also be true for catcode 0 and charcode 1, which may or may not be filtered elsewhere, I don't know from this snippet.

This snippet comes directly before the other cases are checked, so this should only trigger for 10/0.

I'll change it to `#2#1 = 100`.

zauguin: Testing shows that `\if_int_compare:w #2#1 = 100 \exp_stop_f:` does not work directly since the input isn't checked enough at this point: `#1` can still be negative. But the order doesn't really matter now so a bit of reordering makes sure that that case is handled earlier.

PR #1087, opened 2022-04-24
Token to meaning

Skillmon:
This PR adds a global name-space function of the internally used `\__char_change_case_catcode:N`. However, this doesn't replace `\__char_change_case_catcode:N` where it was already used, as it is slightly slower (for stability reasons).

Changed aspects compared to `\__char_change_case_catcode:N` are:

- supports more categories (1, 2, and 6 added)
- wrapped inside `\group_align_safe_begin:` and `\group_align_safe_end:`
- alignment-critical steps are expanded by `\int_value:w ... \exp_stop_f:`


josephwright:
I'm happy with the idea here, but I wonder about `to` in the name - I see why you've used it, but I wonder if you `\token_catcode:N` fits better with things like `\cs_replacement_spec:N` (which feels in the same general area to me).

Skillmon:
@josephwright the name was chosen to be symmetrical with `\token_to_meaning:N` (which could've been named `\token_meaning:N`, imho), and to avoid future confusion I'd suggest to stick with it, but as always: The team has the last word in it :)

Skillmon:
> Looking back, I wonder about \token_ here as the code assumes a character token: shouldn't it cover being passed a control sequence or similar? Or it's a char function.

No, it also deals with macros (see https://github.com/latex3/latex3/pull/1087/files#diff-14601246c9f21ea1c31bc7ccdbda0542afe6830834356457b534ed8ecc5e7645R504), so imho, `token` is the correct category, not `char`.

Which is the correct "added" date?

Tests are a good point... :)

josephwright:
Mark date as today: will go into next `expl3` release

Skillmon:
There is something else that needs thought: Is `13` as the return value of an undefined control sequence fine? The current code returns 13 if no other category was matched...

josephwright:
> There is something else that needs thought: Is `13` as the return value of an undefined control sequence fine? The current code returns 13 if no other category was matched...

Control sequences are catcode 16

muzimuzhi:
This PR added `\token_to_catcode:N` but titled "Token to meaning"?

josephwright:
@muzimuzhi  We are quite laid-back :)

Skillmon:
Ohh no, someone found out! The branch name was a brain fart, and it just stayed by that name, and I never checked the PR heading (it's auto-generated from the branch name).

PR #1091, opened 2022-05-08
Add new command `\lua_load_module:n`

gucci-on-fleek:
This PR adds a new command `\lua_load_module:n` to the `l3luatex` module as a replacement for `\lua_now:n { require "module" }`.

### Better error handling

Right now, if you try and add a Lua module that doesn't exist, you get a fairly unpleasant error message.

```tex
\ExplSyntaxOn
    \lua_now:n { require "does-not-exist" }
\ExplSyntaxOff
```

```log
[\directlua]:1: module 'does-not-exist' not found:
	no field package.preload['does-not-exist']
	[kpse lua searcher] file not found: 'does-not-exist'
stack traceback:
	[C]: in function 'require'
	[\directlua]:1: in main chunk.
\lua_now:e #1->\__lua_now:n {#1}

l.13     \lua_now:n { require "does-not-exist" }

The lua interpreter ran into a problem, so the
remainder of this lua chunk will be ignored.

! Emergency stop.
```

The problem here is that the file `does-not-exist.lua` doesn't exist, but this isn't immediately obvious from the error message if you don't know any Lua. With this PR, you get a much nicer error message.

```tex
\ExplSyntaxOn
    \lua_load_module:n { does-not-exist }
\ExplSyntaxOff
```

```log
! LaTeX3 Error: Lua module `does-not-exist' not found.

For immediate help type H <return>.
 ...

l.13     \lua_load_module:n { does-not-exist }


The file `does-not-exist.lua' could not be found in the current directory or
in the TeX search path. Please ensure that the file was properly installed and
that the filename database is current.
```

(This is the main motivation for this new command)

### Very commonly used

A significant portion of `\lua_now:(n|e)` calls are used just to load a Lua module. Across GitHub, [12 files match `/\\lua_now:(n|e)\s*\{.*require[\s\(]*("|')/` ](https://cs.github.com/?scopeName=All+repos&scope=&q=%2F%5C%5Clua_now%3A%28n%7Ce%29%5Cs*%5C%7B.*require%5B%5Cs%5C%28%5D*%28%22%7C%27%29%2F#) while [61 files match `/\\lua_now:(n|e)\s*\{/`](https://cs.github.com/?scopeName=All+repos&scope=&q=%2F%5C%5Clua_now%3A%28n%7Ce%29%5Cs*%5C%7B%2F#). These numbers include forks and other copies of `expl3`; if we exclude those I'd say that roughly a quarter of `\lua_now:(n|e)` calls are used to import a Lua module. I think that this is sufficiently popular to merit a specific interface.

### Encourages correct use of Lua

Using `\lua_now` can lead to some pretty surprising results if you aren't familiar with how LuaTeX processes `\directlua`. Having `\lua_load_module:n` should encourage authors to avoid `\lua_now`. In an ideal world, you should only ever need `\lua_load_module:n`, although of course `\lua_now` is sometimes still useful.

### Hooks

With this interface, it would be possible for someone to add hooks like `ltfilehook` package. This probably isn't useful — but it would at least be possible now.

## Specific Details

These are some specific details where I'm unsure; feel free to comment or suggest changes here (or anywhere else too!).

### “package” vs “module”

I named the function `\lua_load_module:n` instead of `\lua_load_package:n` since I _think_ that “module” is the correct term, but it's a little confusing.

From [the Lua manual](https://www.lua.org/manual/5.4/manual.html#6.3):

> The **`package`** library provides basic facilities for loading **modules** in Lua. It exports one function directly in the global environment: `require`. Everything else is exported in the table `package`.

Let me know if you think I should pick a different command name.

### Testing if the module exists

I'm using `\file_if_exist:nTF { #1 .lua }` to see if the module exists. To be technically correct, we should use a Lua test like:

```lua
function search(name)
	for _, searcher in ipairs(package.searchers) do
		if type(searcher(name)) == "function" then
			return true
		end
	end
	return false
end
```

LuaTeX by default has `package.searchers[1]` check to see if the module is already loaded, then `package.searchers[2]` searches with `kpse`. This is essentially equivalent to `\file_if_exist`, so I doubt there will be any difference in practice, but I can use the technically-correct way if you prefer.

### String escaping

I'm using `\lua_escape` to escape the module name. This is technically necessary I guess, but in practice I doubt that anyone would name a file like `a%\b~c $  d.lua` so it can probably safely be removed.

### Message level

I'm using `\msg_error` if the module can't be found, although `\msg_critical` might be more appropriate since there's not much that most packages can do if they can't load their Lua module.

### Variants

`\lua_load_module:n` only covers `\lua_now:n { require "module" }`; I could trivially extend this to `\lua_load_module:nn -> \lua_now:n { #1 = require "#2" }` if you think that's useful.

### Source documentation

All the `l3luatex` code in `source3.pdf` is italicized; I have no idea what is causing this.

### File location

I added this command to `l3luatex.dtx` since that seemed to be the most logical place to put it, although perhaps I should have put it in `l3candidates.dtx`?

---

Let me know if you have any feedback. Thanks!

### DIFF
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -361,6 +361,9 @@ local cprint     = tex.cprint
 local write      = tex.write
 local write_nl   = texio.write_nl
 local utf8_char  = utf8.char
+local package_loaded    = package.loaded
+local package_searchers = package.searchers
+local table_concat      = table.concat
 
 local scan_int     = token.scan_int or token.scan_integer
 local scan_string  = token.scan_string
@@ -605,6 +608,33 @@ end
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}[int]{try_require}
+%   Loads a Lua module. This function loads the module similarly to the standard
+%   Lua global function |require|, with a few differences. On success,
+%   |try_require| returns |true, module|. If the module cannot be found, it
+%   returns |false, err_msg|. If the module is found, but something goes wrong
+%   when loading it, the function throws an error.
+%    \begin{macrocode}
+local function try_require(name)
+  if package_loaded[name] then
+    return true, package_loaded[name]
+  end
+
+  local failure_details = {}
+  for _, searcher in ipairs(package_searchers) do
+    local loader, data = searcher(name)
+    if type(loader) == 'function' then
+      package_loaded[name] = loader(name, data) or true
+      return true, package_loaded[name]
+    elseif type(loader) == 'string' then
+      failure_details[#failure_details + 1] = loader
+    end
+  end
+
+  return false, table_concat(failure_details, '\n')
+end
+%    \end{macrocode}
+% \end{macro}
 %
 % \begin{macro}{\@@_load_module_p:n}
 %   Check to see if we can load a module using |require|. If we
@@ -617,7 +647,7 @@ local c_false_bool = token_create(0, char_given)
 local c_str_cctab  = token_create('c_str_cctab').mode
 
 luacmd('@@_load_module_p:n', function()
-  local success, result = pcall(require, scan_string())
+  local success, result = try_require(scan_string())
   if success then
     set_macro(c_str_cctab, 'l_@@_err_msg_str', '')
     put_next(c_true_bool)
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -347,23 +347,20 @@ local tonumber = tonumber
 %
 %   Local copies of standard functions.
 %    \begin{macrocode}
-local abs               = math.abs
-local byte              = string.byte
-local floor             = math.floor
-local format            = string.format
-local gsub              = string.gsub
-local lfs_attr          = lfs.attributes
-local open              = io.open
-local os_date           = os.date
-local setcatcode        = tex.setcatcode
-local sprint            = tex.sprint
-local cprint            = tex.cprint
-local write             = tex.write
-local write_nl          = texio.write_nl
-local utf8_char         = utf8.char
-local package_loaded    = package.loaded
-local package_searchers = package.searchers
-local table_concat      = table.concat
+local abs        = math.abs
+local byte       = string.byte
+local floor      = math.floor
+local format     = string.format
+local gsub       = string.gsub
+local lfs_attr   = lfs.attributes
+local open       = io.open
+local os_date    = os.date
+local setcatcode = tex.setcatcode
+local sprint     = tex.sprint
+local cprint     = tex.cprint
+local write      = tex.write
+local write_nl   = texio.write_nl
+local utf8_char  = utf8.char
 
 local scan_int     = token.scan_int or token.scan_integer
 local scan_string  = token.scan_string
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -187,14 +187,6 @@
 %   found, nothing is returned with \emph{no error raised}.
 % \end{function}
 %
-% \begin{function}[added = 2022-05-14]{ltx.utils.require}
-%   \begin{syntax}
-%     |err_msg = ltx.utils.require(|\meta{module name}|)| \\
-%   \end{syntax}
-%   Loads a Lua module, returining |nil| on success and an error message
-%   on failure.
-% \end{function}
-%
 % \end{documentation}
 %
 % \begin{implementation}
@@ -249,16 +241,16 @@
 % \end{macro}
 %
 % \begin{macro}{\lua_load_module:n}
-%   Wrapper around |ltx.utils.require'|\meta{module}|'|.
+%   Wrapper around |require'|\meta{module}|'|.
 %    \begin{macrocode}
-\cs_generate_variant:Nn \bool_if:nF { e }
+\str_new:N \l_@@_err_msg_str
 \cs_generate_variant:Nn \msg_error:nnnn { nnnV }
 \cs_new_protected:Npn \lua_load_module:n #1
   {
-    \bool_if:eF { \@@_load_module_p:n { #1 } }
+    \bool_if:nF { \@@_load_module_p:n { #1 } }
       {
         \msg_error:nnnV
-          { luatex } { module-not-found } { #1 } { \g_@@_err_msg_str }
+          { luatex } { module-not-found } { #1 } { \l_@@_err_msg_str }
       }
   }
 %    \end{macrocode}
@@ -616,42 +608,11 @@ end
 %    \end{macrocode}
 % \end{macro}
 %
-% \begin{macro}{ltx.utils.require}
-%   Loads a Lua module. This function loads the module just like the standard
-%   Lua global function |require|, with two key differences:
-%   \begin{enumerate}
-%     \item This function is guaranteed not to throw an error.
-%     \item This function does \emph{not} return the loaded module; instead, it
-%           returns |nil| on success and an error message on failure.
-%   \end{enumerate}
-%    \begin{macrocode}
-local function require(name)
-  if package_loaded[name] then
-    return
-  end
-
-  local failure_details = {}
-  for _, searcher in ipairs(package_searchers) do
-    local loader, data = searcher(name)
-    if type(loader) == 'function' then
-      package_loaded[name] = loader(name, data) or true
-      return
-    elseif type(loader) == 'string' then
-      failure_details[#failure_details + 1] = loader
-    end
-  end
-
-  return table_concat(failure_details, '\n')
-end
-
-ltx.utils.require = require
-%    \end{macrocode}
-% \end{macro}
 %
 % \begin{macro}{\@@_load_module_p:n}
-%   Check to see if we can load a module using |ltx.utils.require|. If we can
-%   load the module, then we load it immediately. Otherwise, we save the error
-%   message in |\g_@@_err_msg_str|.
+%   Check to see if we can load a module using |require|. If we
+%   can load the module, then we load it immediately. Otherwise, we save the
+%   error message in |\l_@@_err_msg_str|.
 %    \begin{macrocode}
 local char_given   = command_id'char_given'
 local c_true_bool  = token_create(1, char_given)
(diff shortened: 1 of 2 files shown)


zauguin:
I'm generally in favor of this. Some comments:

> The problem here is that the file `does-not-exist.lua` doesn't exist, but this isn't immediately obvious from the error message if you don't know any Lua. With this PR, you get a much nicer error message.

One problem I see with the new message is that it contains much less information about where the file was actually searched. It doesn't matter so much for simple Lua files, but especially with customized searchers I think that the additional information provided by Lua's require is very valuable. Maybe that could be added at the end of the help text such that the user first gets user-friendly messages and then some technical details?

> since I _think_ that “module” is the correct term

I agree.

> I'm using `\file_if_exist:nTF { #1 .lua }` to see if the module exists. To be technically correct, we should use a Lua test like:
> 
> [...]
> 
> LuaTeX by default has `package.searchers[1]` check to see if the module is already loaded, then `package.searchers[2]` searches with `kpse`. This is essentially equivalent to `\file_if_exist`, so I doubt there will be any difference in practice, but I can use the technically-correct way if you prefer.

This should definitely use the more correct test. For once, the `kpse` search is not actually equivalent to `\file_if_exists` since `\file_if_exists` does a TeX file search with kpathsea, while Lua `require`'s do a Lua file search. Therefore they use different search paths (TEXINPUTS vs LUAINPUTS). By default it seems to work since many Lua files in TeX Live are in the `tex` directory, but e.g. files in the `scripts` directory are found with `require` but not with `\file_if_exist`. Beside that it's not unusual for `searchers` to be different in some cases. E.g. when shell-escape is enabled a searcher for binary modules is added and packages like `luapackageloader` mess with the searcher list.

Additionally there is the old issue of file searches being rather slow, especially on a certain well-known operating system.
So maybe it would be best to avoid any repeated lookups by reimplementing require with different error handling for this.

> I'm using `\lua_escape` to escape the module name. This is technically necessary I guess, but in practice I doubt that anyone would name a file like `a%\b~c $ d.lua` so it can probably safely be removed.

While I agree with the general sentiment that people using weird character in filenames don't deserve much support, I also think that we should still handle such names correctly.

That being said I would recommend scanning the name with `token.scan_argument` which avoids the entire issue.

> I'm using `\msg_error` if the module can't be found, although `\msg_critical` might be more appropriate since there's not much that most packages can do if they can't load their Lua module.

That's a good question I don't really have a qualified opinion about. @josephwright what do you think would be better here? 

> ### Variants
> 
> `\lua_load_module:n` only covers `\lua_now:n { require "module" }`; I could trivially extend this to `\lua_load_module:nn -> \lua_now:n { #1 = require "#2" }` if you think that's useful.

I think it's better to leave it with `\lua_load_module:n`. While there are certainly usecases where `\lua_load_module:nn` makes sense, saving stuff in global variables IMO should never be the default and should require a conscious decision by the author.

> ### Source documentation
> 
> All the `l3luatex` code in `source3.pdf` is italicized; I have no idea what is causing this.

That's unrelated to this change. `doc` makes all code italic which is guarded by more than `StandardModuleDepth` docstrip guards. Since most code in source3 is only guarded by `<*package>` this is set to 1, but in `l3luatex` everything is additionally guarded by `<*tex>` or `<*lua>` and therefore triggers the italic text. We should probably change this at some point...

> ### File location
> 
> I added this command to `l3luatex.dtx` since that seemed to be the most logical place to put it, although perhaps I should have put it in `l3candidates.dtx`?

IMHO the location is fine.

gucci-on-fleek:
@zauguin 

> but especially with customized searchers I think that the additional information provided by Lua's require is very valuable. Maybe that could be added at the end of the help text such that the user first gets user-friendly messages and then some technical details?

Added.

> This should definitely use the more correct test. […] Beside that it's not unusual for searchers to be different in some cases. E.g. when shell-escape is enabled a searcher for binary modules is added and packages like luapackageloader mess with the searcher list.

Fixed.

> I would recommend scanning the name with token.scan_argument

Done.

Let me know if you want any more changes.

l3kernel/l3luatex.dtx
```diff
@@ -208,9 +234,7 @@
 % \begin{macro}[EXP]{\lua_now:n, \lua_now:e}
 % \begin{macro}{\lua_shipout_e:n, \lua_shipout:n}
 % \begin{macro}[EXP]{\lua_escape:n, \lua_escape:e}
-%   Wrappers around the primitives. As with engines other than \LuaTeX{}
-%   these have to be macros, we give them the same status in all cases.
-%   When \LuaTeX{} is not in use, simply give an error message/
+%   Wrappers around the primitives.
```
zauguin, line 237: Any particular reason for this change?

l3kernel/l3luatex.dtx
```diff
@@ -219,6 +243,31 @@
...
 \cs_new:Npn \lua_escape:n #1 { \lua_escape:e { \exp_not:n {#1} } }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\lua_load_module:n}
+%   Wrapper around |ltx.utils.require'|\meta{module}|'|.
+%    \begin{macrocode}
+\cs_generate_variant:Nn \bool_if:nF { e }
```
zauguin, line 254: `\bool_if:nTF` already evaluates the argument, so I don't think that `e` type expansion is needed here.

l3kernel/l3luatex.dtx
```diff
@@ -554,6 +616,61 @@ end
...
+    if type(loader) == 'function' then
+      package_loaded[name] = loader(name, data) or true
+      return
+    elseif type(loader) == 'string' then
+      failure_details[#failure_details + 1] = loader
+    end
+  end
+
+  return table_concat(failure_details, '\n')
+end
```
zauguin, line 645: I would prefer an interface which can also be used if the return value is needed and which follows the Lua convention of returning something which evaluates as false followed by the error when an error happens.

l3kernel/l3luatex.dtx
```diff
@@ -554,6 +616,61 @@ end
...
+
+ltx.utils.require = require
```
zauguin, line 647: I'm not sure if we want to make this public at the moment or if we prefer to keep the option to provide a public `require` with better error management later. If having this available is useful, we could maybe use a name which indicates this? Maybe `try_require`?

l3kernel/l3luatex.dtx
```diff
@@ -219,6 +243,31 @@
...
+%   Wrapper around |ltx.utils.require'|\meta{module}|'|.
+%    \begin{macrocode}
+\cs_generate_variant:Nn \bool_if:nF { e }
+\cs_generate_variant:Nn \msg_error:nnnn { nnnV }
+\cs_new_protected:Npn \lua_load_module:n #1
+  {
+    \bool_if:eF { \@@_load_module_p:n { #1 } }
+      {
+        \msg_error:nnnV
+          { luatex } { module-not-found } { #1 } { \g_@@_err_msg_str }
```
zauguin, line 261: `\str_new:N \g_@@_err_msg_str` is missing.

l3kernel/l3luatex.dtx
```diff
@@ -554,6 +616,61 @@ end
...
+%    \begin{macrocode}
+local char_given   = command_id'char_given'
+local c_true_bool  = token_create(1, char_given)
+local c_false_bool = token_create(0, char_given)
+local c_str_cctab  = token_create('c_str_cctab').mode
+
+luacmd('@@_load_module_p:n', function()
+  local result = require(scan_string())
+  if result == nil then
+    set_macro(c_str_cctab, 'g_@@_err_msg_str', '', 'global')
```
zauguin, line 664: Why global?

l3kernel/l3luatex.dtx
```diff
@@ -554,6 +616,61 @@ end
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}{ltx.utils.require}
+%   Loads a Lua module. This function loads the module just like the standard
+%   Lua global function |require|, with two key differences:
+%   \begin{enumerate}
+%     \item This function is guaranteed not to throw an error.
```
zauguin, line 623: It's not though. If the module is found, but loading it triggers an error, then `require` will throw an error. (I think that's correct behavior, but it contradicts this statement)

zauguin (approved):
Thanks. Some smaller comments, but nothing that couldn't be fixed later if necessary.

l3kernel/l3luatex.dtx
gucci-on-fleek: The non-LuaTeX error code needs to come after the definition of `\lua_load_module:n`, but it would be odd to define `\lua_load_module:n` before we wrap the primitives, so I broke the original `macrocode` block into two.

l3kernel/l3luatex.dtx
gucci-on-fleek: Correct. Fixed

l3kernel/l3luatex.dtx
gucci-on-fleek: That interface would make much more sense. And it turns out that with that interface, I can just use `pcall` instead of reimplementing `require`, which simplifies things quite a bit.

l3kernel/l3luatex.dtx
gucci-on-fleek: Added

l3kernel/l3luatex.dtx
gucci-on-fleek: Wasn't sure if there was any hidden grouping somewhere. Changed to local

l3kernel/l3luatex.dtx
zauguin: Oh sorry, I missed that you moved to comment to later. I thought that it just got removed which confused me.

l3kernel/l3luatex.dtx
zauguin: Personally I preferred the older version since it separates error messages better. By moving `require` completely into a `pcall` we don't just catch errors due to the module not being found but also error from failing to load the module. Of course, if the actual error is that a module failed to load, having a "Lua module `...' not found." error message would be very misleading.

gucci-on-fleek: Sure. I've (re)added the `try_require` function. 

One problem now though is that if the module throws an error when it's loaded, it puts TeX into a bad state. If you were to do `\directlua{require "something-that-throws-an-error"}`, the Lua interpreter will error out, but this won't propagate to the TeX end. With this change, `\lua_load_module:n {something-that-throws-an-error}` also puts TeX into a bad state since `put_next(c_{true/false}_bool)` is never called. We get a nice Lua stack trace either way though, so I think that this is fine.

l3kernel/l3luatex.dtx
gucci-on-fleek: I've made the function private with the name `try_require`.

zauguin (approved):

zauguin:
Thank you for the PR.

PR #1100, opened 2022-06-21
add pagesize setting to the backend

u-fischer:
This adds the code to set the mediabox /pagesize currently used in graphics-def into l3backend so that the  mediabox are set always and not only if packages like graphics or geometry are loaded. 

There is no public interface currently (changing the pagesize in mid-document doesn't work for all engines with these functions) and testfiles are still missing.

### DIFF
--- a/l3backend/l3backend-pdf.dtx
+++ b/l3backend/l3backend-pdf.dtx
@@ -768,6 +768,47 @@
 % \end{macro}
 % \end{macro}
 %
+% \subsubsection{PDF page size (mediabox)}
+% This code is more or less a copy of the code in graphics-def.
+% The code tests that the dimensions are larger than
+% zero as setting a zero page size doesn't make sense.
+% The page size is set in \cs{@kernel@before@begindocument}:
+% this is after the backend code has been loaded but before other
+% settings from package like geometry or graphics or memoir.
+% Similar to the setting in graphics nothing is done if
+% mag has been set. For dvi backends we delay the setting
+% to the shipout/firstpage hook.
+% \begin{macro}{\@@_backend_pagesize_set:nn}
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_backend_pagesize_set:nn #1#2
+  {
+    \bool_lazy_and:nnT
+      {\dim_compare_p:nNn {#1}>{0pt}} {\dim_compare_p:nNn {#2}>{0pt}}
+      {
+        \__kernel_backend_literal:e
+           {papersize=\dim_eval:n{#1},\dim_eval:n{#2}}
+      }
+  }
+
+\tl_gput_right:Nn \@kernel@before@begindocument
+ {
+   \int_compare:nNnT{\mag} = {\@m}
+    {
+      \__kernel_backend_first_shipout:n
+       {
+         \cs_if_exist:NTF \stockwidth
+           {
+             \@@_backend_pagesize_set:nn {\stockwidth}{\stockheight}
+           }
+           {
+             \@@_backend_pagesize_set:nn {\paperwidth}{\paperheight}
+           }
+       }%
+    }
+ }
+%    \end{macrocode}
+% \end{macro}
+%
 %    \begin{macrocode}
 %</dvips>
 %    \end{macrocode}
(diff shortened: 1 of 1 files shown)


l3backend/l3backend-pdf.dtx
```diff
@@ -768,6 +768,47 @@
...
+      }
+  }
+
+\tl_gput_right:Nn \@kernel@before@begindocument
+ {
+   \int_compare:nNnT{\mag} = {\@m}
+    {
+      \__kernel_backend_first_shipout:n
+       {
+         \cs_if_exist:NTF \stockwidth
```
josephwright, line 799: Do we really need this?

l3backend/l3backend-pdf.dtx
```diff
@@ -768,6 +768,47 @@
...
+  {
+    \bool_lazy_and:nnT
+      {\dim_compare_p:nNn {#1}>{0pt}} {\dim_compare_p:nNn {#2}>{0pt}}
+      {
+        \__kernel_backend_literal:e
+           {papersize=\dim_eval:n{#1},\dim_eval:n{#2}}
+      }
+  }
+
+\tl_gput_right:Nn \@kernel@before@begindocument
```
josephwright, line 793: We can't be sure that this exists: remember that the backend has to work with non-LaTeX format.s

l3backend/l3backend-pdf.dtx
```diff
@@ -1227,6 +1268,43 @@
...
+      {\dim_compare_p:nNn {#1}>{0pt}} {\dim_compare_p:nNn {#2}>{0pt}}
+      {
+       \dim_set:Nn  \tex_pagewidth:D {#1}
+       \dim_set:Nn  \tex_pageheight:D {#2}
+      }
+  }
+
+\tl_gput_right:Nn \@kernel@before@begindocument
+ {
+   \int_compare:nNnT{\mag} = {\@m}
```
josephwright, line 1294: Not keen on `\@m` in `expl3` code, nor `\mag` if we mean `\tex_mag:D`.

josephwright (changes requested):
We have a lot of common code here, which doesn't live in `l3backend` at all. We need to work out where this is going at a different level - for the present could be `l3pdf` I think.

davidcarlisle (commented):
The  danger of enabling this by default rather than say just for new documents specifying `\DocumentMetadata` is changes to existing layout

If this document is processed on a site which defaults pdf size to A4, you get an arrow pointing to 1, but setting the pge size here is like uncommenting graphics and the page number is cut off as the short page is enforced

```
\documentclass{article}
\addtolength\textheight{9\baselineskip}

%\usepackage{graphics}

\begin{document}

aa\\bb

\rule{20pt}{620pt}

page number here ...........................$\downarrow$

\end{document}
```

the existing document is "wong" as it is setting a letter paper page block on a A4 page, but the manual setting of `\textheight` is correcting for that so auomatically forcing the "correct" US Letter media size ends up losing the page foot. 

there are probably a lot of documents doing essentially this.

That said, if we could document this and make the change it would be a good thing, having a default of setting a US Letter text block on an A4 page is hard to justify.

car222222:
Just a brief comment for now:

More generally, the whole of the backend, together with its related files such as l3pdf etc., will need a major cleanup/reworking before moving anything out of the experimental 'pdf-management/latex-lab' world.

davidcarlisle:
> I'm really not sure the \mag or 0pt tests are needed 

This was a historic compromise at the time to avoid breaking existing platex setup and in paticular avoiding timing issues over package loading order. With expl3 being in the format and so simplifying possible load orders, we can probably look at that again, with the platex maintainers.

> I think we don't need a second hook use.

Origially some dvi drivers had a "first page setting wins" and some had  "last page setting wins" and some had "last page setting wins, but must be on first page". That meant finding a place to insert the code was interesting.  But dvips changed its default and other dvi drivers have dropped out of use and I believe all supported back ends work in essentially the same way now with regard to timing, and as you say just spilt on needing a `\special` or setting internal registers

josephwright:
I'm working on ideas in https://github.com/latex3/latex3/compare/setpdfpagesize-alt?expand=1: backend part is done, but I want a common interface in `l3pdf` (to follow later).

josephwright:
Handled in a slightly different way after discussion.

PR #1120, opened 2022-08-05
Add \text_map_... functions

josephwright:
Currently missing at the implementation level are regional indicator and extended pictogram rules: to do later as the requirement becomes clearer.

I plan to add more tests based on https://www.unicode.org/Public/UCD/latest/ucd/auxiliary/GraphemeBreakTest.txt, but that will come in a separate commit as it needs a bit of separate work to automate fully.


josephwright:
Note that at present this code recurses into brace groups. The thinking is that `\emph{Graphemes}` might be reasonable to handle, so we need recursion. Does leave open the question of what `\text_map_break:` should do there. Also I had to decided how to handle control sequences, and at the moment there is no special treatment for math more, awkward control sequence arguments, etc.

josephwright:
(Really comes down to the fact that mapping to input like this really only makes sense if it's 'pure text' - once it's not, all hell breaks loose.)

josephwright:
One might argue that like for case changing, rather than a mapping we could provide a 'transformation' approach, which therefore would avoid some of those questions. But it might be hard to define what it _does_ mean.

PhelypeOleinik (approved):

josephwright:
@PhelypeOleinik  Thanks

josephwright:
Other than the regional business, I think the test failures are now pretty low-priority: most seem to relate to text that a user would never give as they mix up chars from different alphabets.

l3kernel/testfiles/m3text006.lvt
```diff
@@ -0,0 +1,141 @@
...
+
+\START
+\AUTHOR{Joseph Wright}
+
+\ExplSyntaxOn
+
+\TEST { Grapheme~mapping }
+  {
+    \OMIT
+      \ior_open:Nn \g_tmpa_ior { GraphemeBreakTest.txt }
```
FrankMittelbach, line 21: That test file is a bit heavy without any documentation what it does or supposed to be tested. Could there be a few words up at the top  saying that?

l3kernel/testfiles/m3text006.tlg
```diff
@@ -0,0 +1,540 @@
...
+TEST 1: Grapheme mapping
+============================================================
+TRUE
+TRUE
+TRUE
+TRUE
+TRUE
+TRUE
+TRUE
+TRUE
```
FrankMittelbach, line 14: if that ever says something else you have probably a hard time to figures out which TRUE is which :-)

FrankMittelbach (commented):
only looked over the documentation parts

l3kernel/testfiles/m3text006.tlg
josephwright: True, and but when it does that it's quite easy to add a `\TYPEOUT` and check

josephwright: Ah, I actually left that in the test, so a failure will print the text which is wrong.

FrankMittelbach: ok then :-)

PR #1141, opened 2022-10-14
Load Unicode data for all engines

josephwright:
This PR moves case data to a two-part table structure, allowing full Unicode case changing for all engines.* 

There are various questions I would like people to consider before merging

- Does the order of output from `\codepoint_to_bytes:n` make sense? It's the same as the deprecated `\char_to_utfviii_bytes:m`. Also, do we still need this as we now have `\codepoint_generate:n`
- Is it OK that 8-bit case changing does *not* preserve catcode for the ASCII range? (This was the easiest path: ASCII range we need active chars anyway)
- Do we see any issues with NFD in Greek? I wonder if I need to NFC the chars in 8-bit engines as accents are otherwise after letters
- Related: should it be `\codepoint_to_nfd:n` or `\codepoint_to_NFD:n`?
- `\str_lowercase:n` doesn't currently implement the final-sigma rule: it could, but should it?
- I've called the 'over-ride for one codepoint' function `\text_declare_case_mapping:nnn`, but perhaps a better name would be good?


josephwright:
The (*) is about pTeX. Any form of handling of characters outside of the ASCII range is hard there, and I'm not totally confident that what's implemented is 'right' or even that there is a 'right' approach.

l3kernel/l3str.dtx
```diff
@@ -843,6 +843,16 @@
...
+% \section{Deprecated functions}
+%
+% \begin{function}[EXP, added = 2019-11-26]
+%   {\str_foldcase:n, \str_foldcase:V}
+%   \begin{syntax}
+%     \cs{str_foldcase:n} \Arg{tokens}
+%   \end{syntax}
+%   A previous name for the functionally-identical \cs{str_casefold:n}.
+% \end{function}
+%
```
PhelypeOleinik, line 855: Weren't we completely dropping documentation for deprecated functions?

josephwright: That was the case when we were removing deprecated functions over time. We've now said we won't, so I've been leaving the docs around as they will never disappear. I'm happy to remove if that feels like a better plan.

PhelypeOleinik: Deprecated functions are for code that still uses them, not for newer code, so they don't really need documentation (as a way to discourage their usage).  People will keep using deprecated functions in new code if they are documented.

What could be useful instead would be a “migration manual” so that people looking for the documentation of deprecated functions will find a way to replace them with the correct versions

josephwright: Perfectly reasonable point: do we want that in this PR or separately?

PhelypeOleinik: I think a separate PR to keep it on topic.  I can draft that.

Does a separate section at the end of interface3 sound good?

josephwright: It can go in after candidates I guess: we have 'Additions and Removals' for that part anyway.

josephwright:
@PhelypeOleinik Are you happy with the PR other than the question of how to handle deprecated command documentation?

PhelypeOleinik (approved):

josephwright:
@PhelypeOleinik Thanks: now I need to get some feedback from everyone else on the questions I put in the PR text :)

PR #1192, opened 2023-04-11
Add case changing customisation

josephwright:
This will be the basis for fixing latex3/latex2e#1033. The key question is does the interface here make sense: codepoing by number, inserted tokens 'as is'.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -9,6 +9,10 @@ this project uses date-based 'snapshot' version identifiers.
 
 ### Added
 - `\box_set_clipped:N` and variants (renamed from `\box_clip:N`)
+- `\text_declare_lowercase_mapping:nn(n)`,
+   `\text_declare_titlecase_mapping:nn(n)`
+   and `\text_declare_uppercase_mapping:nn(n)` to allow customization of
+   case changing outcomes
 
 ### Changed
 - `\box_set_trim:Nnnnn` and `\box_set_viewport:Nnnnn` moved to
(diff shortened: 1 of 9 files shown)


josephwright:
The other question is where to put 'standard' overrides: do they go in `expl3` or in 2e or ...? I suspect in `expl3` so they are 'always live', and I also suspect we will add as-and-when they come up (so just the `\v{J}` one to start with).

l3kernel/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+- `\text_declare_<thing>case_mapping:nn(n)` to allow customisation
+  of case changing outcome
```
FrankMittelbach, line 12: as mentioned elsewhere, I suggest to always spell out the command names in full even if this means mentioning 3 or more names rather than some generic one ... people search for names and using grep etc and won't find them in this way.

l3kernel/l3text.dtx
```diff
@@ -145,7 +145,7 @@
 %   \texttt{SpecialCasing.txt}. For \pTeX{}, only the ASCII range is
 %   covered as the engine treats input outside of this range as east Asian.
 %
-%   Language-sensitive conversions are enabled using the \meta{language}
+%   Locale-sensitive conversions are enabled using the \meta{BCP-47}
 %   argument, and follow Unicode Consortium guidelines. Currently, the
 %   languages recognised for special handling are as follows.
```
FrankMittelbach, line 150: wouldn't that be "locale strings recognized..." or something like that instead of languages?
or  "the BCP47 tags ...."

FrankMittelbach (approved):
2 queries

davidcarlisle (approved):

PR #1197, opened 2023-04-14
Move l3box functions to stable

josephwright:
As discussed at the team meeting, including the rename (old names unused in TL)


l3kernel/l3box.dtx
```diff
@@ -874,6 +874,68 @@
...
+%   unless \cs{box_clip:N} is subsequently applied.
+%   The updated \meta{box} is an
+%   hbox, irrespective of the nature of the \meta{box} before the trim
+%   operation is applied.
+%   The behavior of the operation where the trims requested is
+%   greater than the size of the box is undefined.
+% \end{function}
+%
+% \begin{function}[added = 2019-01-23]
+%   {
```
FrankMittelbach, line 921: shouldn't that give the current day? well, I guess there are arguments for both so not sure

l3kernel/l3box.dtx
```diff
@@ -874,6 +874,68 @@
...
+%     \cs{box_set_viewport:Nnnnn} \meta{box} \Arg{llx} \Arg{lly} \Arg{urx} \Arg{ury}
+%   \end{syntax}
+%   Adjusts the bounding box of the \meta{box} such that it has lower-left
+%   co-ordinates (\meta{llx}, \meta{lly}) and upper-right co-ordinates
+%   (\meta{urx}, \meta{ury}). All four co-ordinate positions are
+%   \meta{dim exprs}. Material outside of the bounding box is
+%   still displayed in the output unless \cs{box_clip:N} is
+%   subsequently applied.
+%   The updated \meta{box} is an
+%   hbox, irrespective of the nature of the \meta{box} before the viewport
```
FrankMittelbach, line 935: is the nature of the box changed or (more likely I guess) is another boxing level added? If so mit might be worth pointing out that you generate new boxing levels each time you apply such an operation.

FrankMittelbach (approved):
minor comment/query

l3kernel/l3box.dtx
josephwright: The code in the current form was added then to `l3candidates`: the move to stable doesn't affect that. So I thought this made the most sense: will be different for a move of `l3graphics`, for example, as that's not currently in `l3kernel`.

FrankMittelbach: that's why I said there are arguments both ways

l3kernel/CHANGELOG.md
```diff
@@ -7,9 +7,19 @@ this project uses date-based 'snapshot' version identifiers.
...
+### Changed
+- `\box_set_trim:Nnnnn` and `\box_set_viewport:Nnnnn` moved to
+  stable kernel
+
 ### Fixed
 - Incorrect internal name in `\pdf_pagesize_gset:nn`
 
+### Removed
+- Experimental `\box_clip:N` and variants
+
```
muzimuzhi, line 22: To better reflect it's a renaming, I personally perfer
```suggestion
### Changed
- Experimental `\box_clip:N` and variants renamed to `\box_set_clipped:N` and
  variants
- `\box_set_clipped:N`, `\box_set_trim:Nnnnn`, and `\box_set_viewport:Nnnnn`
  moved to stable kernel

### Fixed
- Incorrect internal name in `\pdf_pagesize_gset:nn`

```

josephwright: I see what you mean, but the idea in  a formal change log is that *anything* that 'appears' goes down in the `Added` and *anything* that 'disappears' goes down in `Removed`: the fact it's a rename is a detail from that point of view.

josephwright: I've clarified the renaming in the changelog

PR #1201, opened 2023-05-10
Cache file seen status

josephwright:


l3kernel/l3file.dtx
```diff
@@ -2849,9 +2849,40 @@
...
+%   a slightly odd gobble here.
+%    \begin{macrocode}
+\cs_new:Npn \@@_full_name_auxi:nn #1#2
+  {
+    \if:w \scan_stop: #1 \scan_stop:
+    \else:
+      \exp_after:wN \use_none:n
+        \cs:w @@_seen_ \tl_to_str:n {#2} : \cs_end:
+    \fi:
+    #1
```
FrankMittelbach, line 2883: not sure I understand that one. If `#1` aka `\@@_size:n{...}` returns empty then why execute it a second time? I.e., shouldn't the `#1` move into the `\else:`?

FrankMittelbach (approved):
with one query that might lead to a tiny improvement (unless I'm mistaken)

l3kernel/l3file.dtx
josephwright: I've already `e`-type expanded `#1`, so it is the result of `\pdffilesize`; thus it's either the value we need to return *or* it's entirely empty. Either way, it doesn't need to be inside the conditional.

FrankMittelbach: ah yes, it is already gone, so no further file access, but still a tiny bit slower, because TeX has to look up what is at `#1` (nothing) instead of just bypassing the else clause at high speed --- most likely never measurable though

davidcarlisle (approved):

PR #1209, opened 2023-05-18
Add \file_input_raw:n

josephwright:


l3kernel/l3file.dtx
```diff
@@ -917,6 +917,21 @@
...
+%   \end{syntax}
+%   Searches for \meta{file name} in the path as detailed for
+%   \cs{file_if_exist:nTF}, and if found reads in the file as
+%   additional \TeX{} source. No data concerning the file is
+%   tracked. If the file is not found, no action is taken.
+%   \begin{texnote}
+%     This function is intended only for contexts where files must
+%     be read purely by expansion, for example at the start of a
+%     table cell in a \tn{halign}.
+%   \end{texnote}
```
FrankMittelbach, line 932: In particular, it omits the execution of all file hooks.

or something like that. Also: shouldn't it be  "an \tn{halign}" ?

l3kernel/l3file.dtx
```diff
@@ -3355,6 +3370,24 @@
 % \end{macro}
 % \end{macro}
 %
+% \begin{macro}{\file_input_raw:n, \@@_input:n}
```
FrankMittelbach, line 3373: `\@@_input_raw:n` ?

l3kernel/l3file.dtx
```diff
@@ -3355,6 +3370,24 @@
...
+%   No error checking, no tracking.
+%    \begin{macrocode}
+\cs_new:Npn \file_input_raw:n #1
+  { \exp_args:Ne \@@_input_raw:n { \file_full_name:n {#1} } }
+\cs_new:Npx \@@_input_raw:n #1
+  {
+    \exp_not:N \tl_if_blank:nF {#1}
+      {
+        \exp_not:N \tex_input:D
+          \sys_if_engine_luatex:TF
```
FrankMittelbach, line 3383: you can run only in one engine for a given run, so why make this a runtime test?

FrankMittelbach (changes requested):
some minor suggestions for improvement

l3kernel/l3file.dtx
josephwright: This is all inside an `x`-type definition precisely for that reason :)

l3kernel/l3file.dtx
muzimuzhi: File hooks are LaTeX2e things so I think for expl3 functions there're always no file hooks involved?

josephwright: I thin the h' in `halign` is voiced, so 'a' not 'an' (so like 'hotel' not 'hour')

car222222: Not around my manor it ain't!  Look it up??:-)

It is in fact neither voiced nor unvoiced as, in my part of TeXland, it is not even pronounced!--
halign is voiced as aitch-align
hence "an halign", please!

car222222: Not around my manor it ain't!  Look it up??:-)

It is in fact neither voiced nor unvoiced as, in my part of TeXland, it is not even pronounced!--
halign is voiced as aitch-align
hence "an halign", please!

car222222: Joseph, given his profession, may be confusing it "haline"??

josephwright:
I take it everyone is happy with the suggested name?

l3kernel/l3file.dtx
```diff
@@ -3355,6 +3370,24 @@
...
+            { {#1} }
+            { \exp_not:N \__kernel_file_name_quote:n {#1} \scan_stop: }
+        }
```
davidcarlisle, line 3386: Can we not (yet) assume a   primitive `\input` takes  braces? (so lose the luatex test?) (needs texlive 2020 or later)

l3kernel/l3file.dtx
```diff
@@ -917,6 +917,21 @@
...
 %
+% \begin{function}[added = 2023-05-18, EXP]{\file_input_raw:n}
+%   \begin{syntax}
+%     \cs{file_input_raw:n} \Arg{file name}
+%   \end{syntax}
+%   Searches for \meta{file name} in the path as detailed for
+%   \cs{file_if_exist:nTF}, and if found reads in the file as
+%   additional \TeX{} source. No data concerning the file is
+%   tracked. If the file is not found, no action is taken.
+%   \begin{texnote}
```
davidcarlisle, line 928: > no action is taken.

sounds like it is skipped (like `\@input`) but I think you mean "no action is taken _by latex_" so  you get an engine level file not found error.

l3kernel/l3file.dtx
josephwright: I wondered about that: I think we can't yet at like you say it's TL'20, and even moving to TL'19 as a requirement is quite 'up to date'!

l3kernel/l3file.dtx
josephwright: Er, not: we have at this point applied `\file_full_name:n` and got an empty result, so we branch to using `\tl_if_blank:nF` and so a missing file will be entirely skipped.

davidcarlisle: oops, sorry I should have noticed `full_name_` did that.

This seems wrong, normalising the name  is  needed for assigning the correct hook if you have hooks, but here to be honest  I think I'd just let it to the primitive, or at most define it as `\def\file_input_raw:n#1{\@@input#1\relax}` so braces work on old releases.

The main use case here is to underlie a document level input command that can be used in tables, but if this one skips missing file errors, I think any document level command would have to not use this, and be defined directly from the primitive.

josephwright: I've adjusted to issue an expandable error - so something we still control

Issue #1210, opened 2023-05-19
Add `\str_mdfive_hash:n` and `\str_get_mdfive_hash:nN`?


muzimuzhi:
`\tex_mdfivesum:D`, which corresponds to `\pdfmdfivesum` in pdfTeX and `\mdfivesum` in XeTeX, has two syntaxes:
- `\tex_mdfivesum:D file <general text>` returns md5 sum of file `<general text>`
- `\tex_mdfivesum:D <general text>` returns md5 sum of `<general text>` taken as a string.

`\file_mdfive_hash:n`, `\file_get_mdfive_hash:nN`, and `\file_get_mdfive_hash:nNTF` wraps the first syntax and this issue suggests to add expl3 functions for the second syntax.

Possible implementation:
```tex
\documentclass{article}

\ExplSyntaxOn
% \cs{str_mdfive_hash:n} \Arg{token list}
\cs_new:Npn \str_mdfive_hash:n
  { \tex_mdfivesum:D }

% \cs{str_mdfive_hash:nN} \Arg{token list} \meta{tl var}
\cs_new_protected:Npn \str_get_mdfive_hash:nN #1#2
  { \__kernel_tl_set:Nx #2 { \tex_mdfivesum:D {#1} } }
\ExplSyntaxOff

\begin{document}
\ExplSyntaxOn
\str_mdfive_hash:n {abc} \par
\str_get_mdfive_hash:nN {abc} \l_mdfive_str
\l_mdfive_str
\ExplSyntaxOff
\end{document}
```

josephwright:
My initial question is 'what's the use case?'. For files, I see the utility (and keep meaning to implement something to be useful for debugging), but other than 'because we can', I don't see where it's useful for strings.

muzimuzhi:
I thought of this when trying to define an `enum` data type as #692 popped up when it got closed.  With syntax like
```tex
\clist_const:Nn \c_enum_colors_clist { RED, GREEN, BLUE }
\enum_new:NN \l_color_enum \c_enum_colors_clist
\enum_set:Nn \l_color_enum { RED }
\enum_case:Nn \l_color_enum
  {
    {RED} {...}
    {GREEN} {...}
    {BLUE} {...}
  }
```
to make enum vars belonging to different enum types/lists never equal, every enum var should hold the whole list or the md5 sum of the list it belongs to.

I admit it's not a solid use case.

josephwright:
[...]

PR #1213, opened 2023-05-19
Add \str_mdfive_hash:n

josephwright:
As title :) I've not done a `get` version as the `\file_get_..` ones are largely historical: before we switch to an expandable file-name search.


l3kernel/l3str.dtx
```diff
@@ -787,6 +787,13 @@
 %   not to \texttt{\i}).
 % \end{function}
 %
+% \begin{function}[added = 2023-05-19, EXP]{\str_mdfive_hash:n}
+%   \begin{syntax}
+%     \cs{str_mdfive_hash:n} \Arg{str}
+%   \end{syntax}
+%   Expands to the MD5 sum generated from the \meta{str}.
```
muzimuzhi, line 794: Similar to `\str_if_eq:nnTF` and `\str_if_eq:eeTF` ([doc](https://github.com/latex3/latex3/blob/139c10f8930315b7c27f8000fea839d7e9ef72fb/l3kernel/l3str.dtx#L247-L266), [code](https://github.com/latex3/latex3/blob/139c10f8930315b7c27f8000fea839d7e9ef72fb/l3kernel/l3str.dtx#L1169-L1180)), I think the argument of `\str_mdfive_hash:n` need not already be string, but can be a general token list.
```tex
\cs{str_if_eq:nnTF} \Arg{tl_1} \Arg{tl_2} \Arg{true code} \Arg{false code}
```

But then
```tex
\cs_new:Npn \str_mdfive_hash:n #1
  { \exp_not:Nw \tex_mdfivesum:D \exp_not:Nw { \tl_to_str:n {#1 } } }
\cs_new:Npn \str_mdfive_hash:e #1 { \tex_mdfivesum:D {#1} }
```

muzimuzhi: I was aware of the `\Arg{tl_1}` part in `\str_if_eq:nnTF` doc when opening #1210 (hence `\cs{str_mdfive_hash:n} \Arg{token list}` was used in the attached example there), but didn't get a full understanding about how to understand the differences between  `\str_` module name and `tl_1`/`tl_2` arg spec until just then, after reading its implementation.
- `\str_xxx:n` accepts a token list but converts it to string before calling the primitive,
- `\str_xxx:e` calls the primitive directly and leaves the tl2str conversion to primitive itself.

l3kernel/testfiles/m3str002.lvt
```diff
@@ -56,5 +56,26 @@
     \tl_if_eq:NNTF \l_tmpa_tl \l_tmpb_tl \TRUE \FALSE
   }
 
+\OMIT
+\tl_set:N \l_tmpa_tl { a }
+\tl_set:N \l_tmpb_tl { a }
```
muzimuzhi, line 61:
```suggestion
\tl_set:Nn \l_tmpa_tl { a }
\tl_set:Nn \l_tmpb_tl { a }
```

muzimuzhi: Hmm the two `\tl_set:N ...` didn't fail the checks because they are in between `\OMIT ... \TIMO`, hence the corresponding error lines were filtered out by `l3build`.  Less robust than I imagined.

l3kernel/testfiles/m3str002.lvt
```diff
@@ -56,5 +56,26 @@
...
+    \NEWLINE
+    \str_mdfive_hash:n { & \par ! ## }
+    \NEWLINE
+    \str_mdfive_hash:n { \l_tmpa_tl }
+    \NEWLINE
+    \str_mdfive_hash:n { \l_tmpb_tl }
+    \NEWLINE
+    \str_mdfive_hash:e { \l_tmpa_tl }
+    \NEWLINE
+    \str_mdfive_hash:e { \l_tmpb_tl }
```
muzimuzhi, line 76: Just self-talk, not a review.  This is certainly not the first time I learnt that some primitives fully expand their args before taking them as string.  I had to use sth like
```tex
% md5 for string "a"
\str_if_eq:eeTF
  { \str_mdfive_hash:e { \l_tmpa_tl } }
  { \str_mdfive_hash:e { a } }
  { EQUAL } { INEQUAL }
\NEWLINE
% md5 for string "\l_tmpa_tl "
\str_if_eq:eeTF
  { \str_mdfive_hash:n { \l_tmpa_tl } }
  { \str_mdfive_hash:e { \c_backslash_str l_tmpa_tl ~ } }
  { EQUAL } { INEQUAL }
```
to convince myself what `\tex_mdfivesum:D` does.

muzimuzhi (approved):

PR #1235, opened 2023-06-19
Drop notes about slow \expanded emulation

muzimuzhi:
Do we want to leave one copy of "more precisely in non-LuaTeX engines older than 2019" and even add a note about "`\expanded` is required since `l3kernel` 2023-05-15 release" in the doc?

Marked as draft because I hope to check over it tomorrow to avoid any missing.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,10 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+- Drop notes about slow `\expanded` emulation. `\expanded` is required since
+  release 2023-05-15
+
 ## [2023-06-16]
 
 ### Changed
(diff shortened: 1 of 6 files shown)


muzimuzhi:
> Do we want to leave one copy of "more precisely in non-LuaTeX engines older than 2019" and even add a note about "`\expanded` is required since `l3kernel` 2023-05-15 release" in the doc?

I decided to not include such note because the existence of `\expanded` is already guarded in `l3bootstrap.dtx`, see 076d79c5ad4c9e6c438f23233ec048adb6e1141a.

muzimuzhi:
Hmm missing one in `l3kernel/README.md`
https://github.com/latex3/latex3/blob/78f4885b20f22f30c60ecb4123a787f9ce037a60/l3kernel/README.md?plain=1#L62-L64

josephwright:
I've rebased and merged manually.

muzimuzhi:
See merged commits in https://github.com/latex3/latex3/compare/8e75a5746390598bfe5d320e482ba880ec062253...681877c02ef58bea3cd8faa2bec779421788f79e

PR #1240, opened 2023-06-27
Refine titlecasing and introduce formal sentence casing

josephwright:


josephwright:
@moewew Direction of travel is hopefully clear here: I will add a mechanism to skip words shortly (I have plans)

josephwright:
@moewew Would welcome your thoughts here: I've now got a full draft set up

l3kernel/testfiles/m3text002.ptex.tlg
```diff
@@ -390,27 +424,27 @@ l. ...  }
...
 lower
 UPPER
 Title
 Title
 WORDS lower
 words UPPER
-Words UPPER
-Words \text_case_switch:nnnn {normal}{lower}{UPPER}{Title}
+Words Title
+Words Title
```
moewew, line 436: Is there going to be a test case for `\text_sentencecase` here as well?

l3kernel/l3text.dtx
```diff
@@ -192,9 +196,9 @@
 %
 %  For titlecasing, note that there are two functions available. The
 %  function \cs{text_titlecase:n} applies (broadly) uppercasing to the first
-%  letter of the input, then lowercasing to the remainder. In contrast,
+%  letter of each \meta{word}, then lowercasing to the remainder. In contrast,
 %  \cs{text_titlecase_first:n} \emph{only} carries out the uppercasing operation,
-%  and leaves the balance of the input unchanged. Determining whether
+%  and leaves the balance of the each \meta{word} unchanged. Determining whether
```
moewew, line 201: Typo: "of the each <word>". (I also had a bit of trouble understanding "balance" here at first, OED says this usage is 'commerce slang'.)

l3kernel/testfiles/m3text002.ptex.tlg
josephwright: That wasn't the plan. Lower-, upper- and titlecasing have specific Unicode characters involved, so are 'fundamentally distinct'. Sentence casing isn't the same: you are applying titlecasing in a selective fashion. The switching mech. is there to handle the situation where something non-expandable needs to change depending on lower/upper/title status, so really shouldn't need to be extended as far as I can see (counter-examples of course could show I'm wrong). Any extension would need a new function, and that would be tricky. at the 2e level.

l3kernel/l3text.dtx
josephwright: I've adjusted

l3kernel/testfiles/m3text002.ptex.tlg
moewew: But `\text_sentencecase` *does* interface with `\text_case_switch:nnnn`, right? At the start we get the "Title" output and afterwards "lower", or is it different? (I guess I just wanted to make sure this is the case and since I didn't manage to run the code I just wanted to see a test here that confirmed this.)

josephwright: Yes, perhaps I should make that clearer: sentence casing means applying titlecasing only to the first word of the input. (I may eventually look at dealing with multiple sentences, which would mean of course it would apply titlecasing to the start of each sentence: looks quite tricky/slow.)

josephwright:
> Is there going to be a `\text_sentencecase_first:n`? Currently we use `\text_titlecase_first:n` in `\MakeCapital` to capitalise the first letter of a string (which may in rare instances consist of several words). From a backwards compatibility (and performance) point it might be nicer to have a function that changes only the first word and leaves the rest alone.

I hadn't added that largely as I didn't remember the original use case for `\text_titlecase_first:n`. At the same time, I've always been a bit wary of that function as it could lead to some odd output (something like input `IJSSELMEER` in Dutch would give `IjSSELMEER`, for example). However, adding `\text_sentencecase_first:n` is trivial, so I'll include it in the PR.

josephwright:
> I'm assuming `\NoCaseChange` also works in `\text_sentencecase:n`? And `\text_case_switch:nnnn` selects either Title or lowercase in `\text_sentencecase` depending on the position?

Both correct: I will make that clearer in the documentation.

car222222:
I approve of this change:  balance ==>  remainder  :-).

moewew:
> I hadn't added that largely as I didn't remember the original use case for `\text_titlecase_first:n`. At the same time, I've always been a bit wary of that function as it could lead to some odd output (something like input `IJSSELMEER` in Dutch would give `IjSSELMEER`, for example). However, adding `\text_sentencecase_first:n` is trivial, so I'll include it in the PR.

Hmm, the point about IJSSELMEER is a good one. With `titlecase` and `sentencecase` now applying to all words, maybe the more interesting function would be one that only applies to the first *word* and then leaves all the rest alone...

But to be fair I'm not sure how useful all of these variants are in general. My thoughts were mainly based on being "as backwards compatible as possible in `biblatex`". I'm not sure if the behaviour in `biblatex` was based on anything other than it being easier to implement.

josephwright:
> Hmm, the point about IJSSELMEER is a good one. With `titlecase` and `sentencecase` now applying to all words, maybe the more interesting function would be one that only applies to the first _word_ and then leaves all the rest alone...

That had occurred to me :) We could provide a general 'first word-remaining words' interface as that's what I've done internally, although I think we'd then need some better descriptors for some of the outcomes.

> But to be fair I'm not sure how useful all of these variants are in general. My thoughts were mainly based on being "as backwards compatible as possible in `biblatex`". I'm not sure if the behaviour in `biblatex` was based on anything other than it being easier to implement.

Stability is important, so I will look at this. There's of course the possibility that stability means we can't change `\text_ttitlecase:n` bur rather have to use a new name: I hope I can make the case that a change is reasonable, but have to see what @FrankMittelbach, @davidcarlisle, @car222222 think in particular.

car222222:
Do we have any information about the current usage of `\text_ttitlecase:n`?  Or about the history of how it got 
misnamed?

josephwright:
@car222222  The original idea was that `\text_titlecase:n` (and indeed the other functions) would be building blocks providing the Unicode mapping for characters but likely within a word only, with a putative word-boundary mapping function dealing with more complex division. However, that then leaves open the question of what happens when you have a space (or other word boundary) inside the argument to `\text_titlecase:n`. At the same time, I think it's clear that for performance reasons, multiple nested mappings are likely to be problematic. So that leads to the conclusion that `\text_...case:n` need to deal with multi-word entry.

moewew:
It might be interesting to compare the results and implementation here to the existing packages https://ctan.org/pkg/mfirstuc and https://www.ctan.org/pkg/titlecaps.

josephwright:
@jbezos It was suggested at a team meeting that your input would be good here. Does 'titlecasing' make sense in general, or should it only be provided at the `babel` level? The open question was is it even meaningful in languages other than English - I hope you can comment. (@zauguin  that a fair summary?)

josephwright:
@moewew  I think I've covered everything they can do - I still need to do `sentencecase-first` (which I guess again might be suggested as 'only sensible in English').

car222222:
Almost the only language I know at all well (amongst those that even have capitals in the script :-) is French --

Majuscules dans les titrep

Unlike most areas of French grammar, the capitalization of French titles of books, movies, etc. does not follow a clearly defined set of rules. According to Le bon usage, French title capitalization is inconsistent, with competing systems used by writers, publishers, and other authoritative sources.

dbitouze:
> Majuscules dans les titres
> 
> Unlike most areas of French grammar, the capitalization of French titles of books, movies, etc. does not follow a clearly defined set of rules. According to Le bon usage, French title capitalization is inconsistent, with competing systems used by writers, publishers, and other authoritative sources.

Indeed, I guess it is hopeless to implement the rules for French (see [Principes de capitalisation des titres](https://fr.wikipedia.org/wiki/Wikip%C3%A9dia:Conventions_typographiques#Titres_d'%C5%93uvres_(livres,_films,_etc.))).

josephwright:
One possible position would of course be just to 'fix' the per-word behaviour but with no customisation: that would mean that the kernel does 'Unicode titlecasing' in the sense it can be _used_ to build a language-dependent approach. Would mean any language-specific versions would be slower, and would mean even English would need `babel`/`polyglossia`/... support. (Overall at present I favour allowing for customisation, as I think it's not too bad, but I am open to counter-arguments.)

josephwright:
Code-wise I think this shows what is doable, so I'm largely after reviews at the 'concept' level. @gmilde I would welcome your input here too.

moewew:
I prepared https://github.com/plk/biblatex/pull/1293 for `biblatex`. If everything here is approved, we can merge the `biblatex` PR and start the release procedure (because it will involve building Biber binaries, it might take a while).

moewew:
Forgot to say: I haven't yet tried to test the `biblatex` changes with the the changes here (mainly because I have not tried to use dev versions of `expl3` yet), so if anyone wants who can do that wants to test the `biblatex` changes (fortunately that is as easy as taking [`blx-case-expl3.sty`](https://github.com/plk/biblatex/pull/1293/files#diff-3d6d73f3690bcdcd5e20b22ec854f3ac7795662e3fa34c4f49466e1fa6dd6f40) from the PR) that would be appreciated.

gmilde:
> Stability is important, so I will look at this. There's of course the
> possibility that stability means we can't change \text_ttitlecase:n
> but rather have to use a new name: I hope I can make the case that a
> change is reasonable, 

I feel uneasy about changing the behaviour of a LaTeX macro. I would
rather keep the current behaviour with a clear description,
an example of its actual behaviour, and a reference to the new macro(s)
that can be used to get titlecase for all words of the argument.

Abandoning the term "sentence case", leaves the possibility for, e.g.,

`\text_titlecase:n` with the current behaviour maybe with alias `\text_titlecase_first_word:n`
`\text_titlecase_words:n` maybe with alias `\text_capitalise:n`.

In use cases relevant for LaTeX (enforcing of certain styles for titles,
section headings, ... in a style file), the input may contain abbreviations,
proper names, German nouns, and WikiWords, typically already with correct
use of capital letters:

  Schneewittchen und die sieben Zwerge.
  works by Shakespeare typeset using LEGO bricks
  die UNO Vollversammlung -- the general assembly of the UNO
  MitarbeiterInnen fordern mehr Lohn
  
Therefore, I see a greater demand for a "titlecasing function" that leaves
everything except the first letter as-is. 
(We may need to look at the first two characters in locales where
two-characters are considered a letter and must both be
upcased/titlecased.)

The "user facing" names could be a bit less technical than `titlecase_first_initial`,
maybe `\Capitalize` and `\CapitalizeWords`. However this would require a good documentation of the difference to `\MakeUppercase`, so a more verbose but self-explaining name might be better. (Look at the existing packages for inspiration or ask their authors or users.)

The case of UPPERCASE words that should be converted to titlecase or
lowercase is rather rare and could be catered for by nesting: 
`\MakeLowercase{\CapitalizeWords{WHATEVER IT TAKES}}`.
So, maybe we don't need a new function `\MakeTitlecaseWords`.

gmilde:
> Stability is important, so I will look at this. There's of course the
> possibility that stability means we can't change \text_ttitlecase:n
> but rather have to use a new name: I hope I can make the case that a
> change is reasonable, 

I feel uneasy about changing the behaviour of a LaTeX macro. I would
rather keep the current behaviour with a clear description,
an example of its actual behaviour, and a reference to the new macro(s)
that can be used to get titlecase for all words of the argument.

Abandoning the term "sentence case", leaves the possibility for, e.g.,

`\text_titlecase:n` with the current behaviour maybe with alias `\text_titlecase_first_word:n`
`\text_titlecase_words:n` maybe with alias `\text_capitalise:n`.

In use cases relevant for LaTeX (enforcing of certain styles for titles,
section headings, ... in a style file), the input may contain abbreviations,
proper names, German nouns, and WikiWords, typically already with correct
use of capital letters:

  Schneewittchen und die sieben Zwerge.
  works by Shakespeare typeset using LEGO bricks
  die UNO Vollversammlung -- the general assembly of the UNO
  MitarbeiterInnen fordern mehr Lohn
  
Therefore, I see a greater demand for a "titlecasing function" that leaves
everything except the first letter as-is. 
(We may need to look at the first two characters in locales where
two-characters are considered a letter and must both be
upcased/titlecased.)

The "user facing" names could be a bit less technical than `titlecase_first_initial`,
maybe `\Capitalize` and `\CapitalizeWords`. However this would require a good documentation of the difference to `\MakeUppercase`, so a more verbose but self-explaining name might be better. (Look at the existing packages for inspiration or ask their authors or users.)

The case of UPPERCASE words that should be converted to titlecase or
lowercase is rather rare and could be catered for by nesting: 
`\MakeLowercase{\CapitalizeWords{WHATEVER IT TAKES}}`.
So, maybe we don't need a new function `\MakeTitlecaseWords`.

josephwright:
> > Stability is important, so I will look at this. There's of course the
> > possibility that stability means we can't change \text_ttitlecase:n
> > but rather have to use a new name: I hope I can make the case that a
> > change is reasonable,
> 
> I feel uneasy about changing the behaviour of a LaTeX macro.

In general that's the position of the team, too, of course. However, I feel we need to avoid being completely boxed in if there are situations where there is clearly an issue with current behaviours *if* the balance of making a change is reasonable. It's a value-judgement, admittedly, but to me with a relatively new feature where we have a clear justification for a change, that is a reasonable position.

> Abandoning the term "sentence case", leaves the possibility for, e.g.,
> 
> `\text_titlecase:n` with the current behaviour and `\text_titlecase_words:n`.

I do like the idea of avoiding 'sentence': I feel it's likely to be a potential future issue.

> In use cases relevant for LaTeX (enforcing of certain styles for titles, section headings, ... in a style file), the input may contain abbreviations, proper names, German nouns, and WikiWords, typically already with correct use of capital letters:
> 
> Schneewittchen und die sieben Zwerge. works by Shakespeare typeset using LEGO bricks die UNO Vollversammlung -- the general assembly of the UNO MitarbeiterInnen fordern mehr Lohn UK and EU agree to disagree

I guess this comes down to where you imagine things being used. I was mainly thinking of bibliographies, where data might come from databases that store only uppercase chars (Web of Knowledge for example used to do this for many years).

> Therefore, I see a greater demand for a "titlecasing function" that leaves everything except the first letter as-is. (We may need to look at the first two characters in locales where two-characters are considered a letter and must both be 
upcased/titlecased.)

Don't worry about number of 'characters': that's a core part of the implementation anyway.

> The "user facing" names could be a bit less technical than `titlecase_first_initial`, maybe `\Capitalize` and `\CapitalizeWords`. However this would require a good documentation of the difference to `\MakeUppercase`, so a more verbose but self-explaining name might be better. (Look at the existing packages for inspiration or ask their authors or users.)
> 
> The case of UPPERCASE words that should be converted to titlecase or lowercase is rather rare and could be catered for by nesting: `\MakeLowercase{\CapitalizeWords{WHATEVER IT TAKES}}`. So, maybe we don't need a new function `\MakeTitlecaseWords`.

I think you mean `\CapitalizeWords{\MakeLowercase{WHATEVER IT TAKES}}`: a fair point and one that can be implemented.

----


I think reading this over I'll make a separate branch that implements a slightly different solution and we can compare and contrast.

josephwright:
I've given this a bit more thought: I'm going to close here and start a new PR with a slightly different approach inspired bby @gmilde

jbezos:
@josephwright I'm catching up on the messages I didn't have time to read, and I missed this whole thread. I'll try to read it in the next few days.

jbezos:
A terminological point. What do we really mean by ‘title case’? At least, three options:

1. The casing to be applied to a title, literally. This depends on the language, the style and even the kind of title (for example, in Spanish capitalization for books and for periodicals are different). It’s based on semantics.

2. A method to capitalize words (not based on semantics). Here ‘title case’ is just a convenient way to identify a casing method, based on what we can (more or less) expect in typical (English) titles. But it’s not restricted to titles, because it applies to proper nouns, too.

3. In Unicode is a character property, identifying letters with a special behavior when title casing (in sense 2) is applied. That’s all. There are some explanations and definitions, indeed, but I think the only purpose is to provide some context to this claim:
   
> The need for a normative Titlecase_Mapping property in the Unicode Standard derives from the fact that the standard contains certain digraph characters for compatibility.

Interestingly, CSS does **not** provide a property for `titlecase`, just `lowercase`, `uppercase` and `capitalize` (which puts the first letter of each word, **if lowercase**, in titlecase; other letters are left unchanged). There are many sensible comments here: https://www.w3.org/TR/css-text-3/#propdef-text-transform
   
So, for me the key point is defining what ‘title case’ means for LaTeX. If the meaning is 1, then we must consider variants and, for consistency, ‘lower case’ should be also language dependent (nouns in German), imho. This is not a short term task (or even middle term), and requires contextual and semantic analysis (NLP? AI? 😉).

In sense 2, I think ’title case’ (more orthographic) is not at the same level as ‘sentence case’ (more grammatical). I agree with the comment “Sentence casing isn't the same: you are applying titlecasing in a selective fashion”. So, I’d avoid it, too.

josephwright:
@jbezos  The aim is 'titlecase` as decribed in the Unicode FAQ, i.e. a combination of (2) and (3). In particular, although I agree that  the property `Titlecase_Mapping` is largely about compatibilty, you can't just say 'uppercase the first char' because of e.g. Dutch `ij` -> `IJ` and Greek accent handling (retained when titlecasing, dropped when uppercasing).

car222222:
It is therefore definitely sensible, even at this late stage, 
to drop altogether the term “titlecase”; except, of course, as 
a reference to the Unicode meaning: i.e., referring only to the character transformation applied to an initial character (or sequence of characters such as ‘ij’). 

Thus I am agreeing with (maybe my interpretation) of the essence of what @jbezos just wrote.

car222222:
I reopened this since both the maes and the documentation still need discussion and further action.

josephwright:
@car222222 I closed here as I've decided against this particular code suggestion: please see https://github.com/latex3/latex3/pull/1247 for my updated suggestion

josephwright:
> It is therefore definitely sensible, even at this late stage, to drop altogether the term “titlecase”; except, of course, as a reference to the Unicode meaning: i.e., referring only to the character transformation applied to an initial character (or sequence of characters such as ‘ij’).
> 
> Thus I am agreeing with (maybe my interpretation) of the essence of what @jbezos just wrote.

As I've pointed out before, the Unicode FAQ do use 'titlecase' for the idea of 'capitalising': https://unicode.org/faq/casemap_charprop.html#4.

car222222:
Ah yes, I got the wrong PR, Sorry! All very confusing . . . 

Maybe both Javier's and my comments should be transferred to #1247.

car222222:
From that FAQ: 'titlecase' *takes its name* from the case format used when forming a title  (emphasis added!) 

And: The titlecase mapping in the Unicode Standard is the mapping applied to the initial character in a word.

I definitely interpret this as saying that Unicode uses the term only in this phrase "titlecase mapping" which is (roughly) a character mapping; but not as a "case format used . . . ", which is only a reference to something other than the use of the term "titlecase" within Unicode.

PR #1244, opened 2023-07-05
Support fp vars and functions

josephwright:


josephwright:
I've copied all of the trial setup here, but I wonder if there should be no `new`/`clear` and only `set`. That would be like colours and avoid the idea this is going to do computer algebra. Thoughts welcome (@davidcarlisle , @FrankMittelbach , @car222222 most likely to have opinions.)

(Aside: I keep meaning to ask the same question about messages - one for the team list!)

l3kernel/l3fp-functions.dtx
```diff
@@ -1,6 +1,6 @@
 % \iffalse
 %
-%% File l3fp-functions.dtx (C) Copyright 2012-2018,2020,2021 The LaTeX Project
+%% File l3fp-functions.dtx (C) Copyright 2012-2018,2020,2021,2023 The LaTeX Project
```
FrankMittelbach, line 3: not important but I wonder if you shouldn't simplify these year lists to just 2012-2023. It isn't that significant that 2022 there was no change :-)

FrankMittelbach (approved):

FrankMittelbach:
one more comment I forgot. Shouldn't ltnews gets a few words about this, without going into details, but more like an FYI? (given that we don't any longer provide new l3news)

josephwright:
> one more comment I forgot. Shouldn't ltnews gets a few words about this, without going into details, but more like an FYI? (given that we don't any longer provide new l3news)

I'll write something about '`expl3` updates' for `ltnews`.

l3kernel/l3fp-functions.dtx
josephwright: IANAL :)

josephwright:
> > one more comment I forgot. Shouldn't ltnews gets a few words about this, without going into details, but more like an FYI? (given that we don't any longer provide new l3news)
> 
> I'll write something about '`expl3` updates' for `ltnews`.

Once of course we have this agreed and merged!

l3kernel/l3fp-functions.dtx
FrankMittelbach: really? :-) had to look IANAL up first ... not in my vocabulary. But anyway, I think we have and had the (C) all the time even if we didn't explicitly mentioned it in some years because there was no change. But up to you.

josephwright: @FrankMittelbach  The 2e sources use a simple range, all of the L3 ones are current 'explicit change years only' - so a change here should be handled separately from the PR.

FrankMittelbach: sure fine. I'm not worried I just think it looks more tidy

l3kernel/l3fp-symbolic.dtx
```diff
@@ -445,23 +402,23 @@
...
 \cs_new:Npn \@@_symbolic_binary_to_tl:Nww #1, #2; #3; @
   {
-    \use:ff
+    \use:ee
       { \exp_last_unbraced:Nf ( { \@@_to_tl_dispatch:w #2; } ) }
       {
-        \use:ff
+        \use:ee
           { \@@_types_cs_to_op:N #1 }
           { \exp_last_unbraced:Nf ( { \@@_to_tl_dispatch:w #3; } ) }
```
Skillmon, line 416:
```suggestion
        \use:e
          {
            \@@_types_cs_to_op:N #1
            ( \@@_to_tl_dispatch:w #3; )
          }
```

Skillmon: Also, nesting `\use:e(e)` inside `\use:e(e)` seems unnecessary here.

l3kernel/l3fp-symbolic.dtx
```diff
@@ -445,23 +402,23 @@
...
       }
   }
 \cs_new:Npn \@@_symbolic_function_to_tl:Nw #1, #2@
   {
-    \use:ff { \@@_types_cs_to_op:N #1 }
+    \use:ee { \@@_types_cs_to_op:N #1 }
```
Skillmon, line 421: Again, unnecessary `\use:ee`, also, I haven't checked whether `\@@_symbolic_function_to_tl:Nw` is used outside an `e`-expansion context anywhere, but if not, the `\use:e` can be dropped completely here.

l3kernel/l3fp.dtx
```diff
@@ -658,6 +658,137 @@
 %   the \meta{code} should make use of the \meta{tl~var}.
 % \end{function}
 %
+% \subsection{Symbolic expressions}
+%
+% Floating point expressions support variables: these are local-only constructs.
```
Skillmon, line 663: Does that mean that `\fp_new_variable:n` will only initialise the variable on a local scope? If so, this breaks with `expl3`-conventions of other modules! If not, the wording should be adjusted.

l3kernel/l3fp-symbolic.dtx
josephwright: Good point - I've sorted that on main as it also applies there

PhelypeOleinik (approved):
I haven't reviewed the code thoroughly.  The documentation looks good.  Maybe this is worth an ltnews entry?

josephwright:
> I haven't reviewed the code thoroughly. The documentation looks good. Maybe this is worth an ltnews entry?

@FrankMittelbach  already asked that :) I just wanted to be sure we are happy with the merge at a concepts level - I suspect we are!

PhelypeOleinik:
> @FrankMittelbach already asked that :)

Oh, sorry, I missed that :)

> I just wanted to be sure we are happy with the merge at a concepts level - I suspect we are!

Aye!

PR #1245, opened 2023-07-07
add 4 missing xetex primitives

davidcarlisle:
added four primitives missing from l3names. XeTeXprotrudechars is a renamed pdftex primitive, the other three are xetex-specific.


davidcarlisle:
If I marked tex_protrudechars as xetex it moaned it was not equal to the primitive (as I dropped the XeTeX prefix) so I marked that one as `\spec` in the test which means it just prints the `\meaning` which then meant it showed a diff between ones that define and do not define this

davidcarlisle:
alternative markup welcome, I wasn't expecting to have to add ptex and uptex tlg here

josephwright:
@davidcarlisle Sorted :)

josephwright (approved):

PR #1247, opened 2023-07-08
Refine titlecasing

josephwright:
Follows from discussion in https://github.com/latex3/latex3/pull/1240. Here, a clearer split occurs between titlecasing one an multiple words without any additional naming needed. It also avoids any change to formal behaviour of the existing functions: note though that the new functions do _not_ lowercase the remainder of their input.

### DIFF
--- a/l3kernel/l3text.dtx
+++ b/l3kernel/l3text.dtx
@@ -94,12 +94,12 @@
 %
 % \section{Case changing}
 %
-% \begin{function}[EXP, added = 2019-11-20, updated = 2022-10-13]
+% \begin{function}[EXP, added = 2019-11-20, updated = 2023-07-08]
 %   {
-%     \text_lowercase:n,  \text_uppercase:n,  \text_titlecase:n,
-%       \text_titlecase_first:n,
-%     \text_lowercase:nn, \text_uppercase:nn, \text_titlecase:nn,
-%       \text_titlecase_first:nn
+%     \text_lowercase:n,  \text_uppercase:n,  \text_titlecase_all:n,
+%       \text_titlecase_once:n,
+%     \text_lowercase:nn, \text_uppercase:nn, \text_titlecase_all:nn,
+%       \text_titlecase_once:nn
 %   }
 %   \begin{syntax}
 %     \cs{text_uppercase:n}  \Arg{tokens}
@@ -115,11 +115,13 @@
 %
 %   Upper- and lowercase have the obvious meanings. Titlecasing may be regarded
 %   informally as converting the first character of the \meta{tokens} to
-%   uppercase and the rest to lowercase. However, the process is more complex
+%   uppercase. However, the process is more complex
 %   than this as there are some situations where a single lowercase character
 %   maps to a special form, for example \texttt{ij} in Dutch which becomes
-%   \texttt{IJ}. The \texttt{titlecase_first} variant does not attempt
-%   any case changing at all after the first letter has been processed.
+%   \texttt{IJ}. There are two functions available for titlecasing: one which
+%   applies the change to each \enquote{word} and a second which only applies
+%   at the start of the input. (Here, \enquote{word} boundaries are spaces:
+%   at present full Unicode word breaking is not attempted.)
 %
 %   Importantly, notice that these functions are intended for working with
 %   user \emph{text for typesetting}. For case changing programmatic data see
@@ -190,11 +192,7 @@
 %       available using \pdfTeX{}.
 %   \end{itemize}
 %
-%  For titlecasing, note that there are two functions available. The
-%  function \cs{text_titlecase:n} applies (broadly) uppercasing to the first
-%  letter of the input, then lowercasing to the remainder. In contrast,
-%  \cs{text_titlecase_first:n} \emph{only} carries out the uppercasing operation,
-%  and leaves the balance of the input unchanged. Determining whether
+%  Determining whether
 %  non-letter characters at the start of text should switch from upper- to
 %  lowercasing is controllable. When \cs{l_text_titlecase_check_letter_bool} is
 %  \texttt{true}, characters which are not letters (category code~$11$) are
--- a/l3kernel/doc/l3obsolete.txt
+++ b/l3kernel/doc/l3obsolete.txt
@@ -53,6 +53,10 @@ Function                            Date deprecated
 \seq_indexed_map_inline:Nn               2020-06-18
 \seq_indexed_map_function:NN             2020-06-18
 \sys_load_deprecation:                   2021-01-11
+\text_titlecase:n                        2023-07-08
+\text_titlecase:nn                       2023-07-08
+\text_titlecase_first:n                  2023-07-08
+\text_titlecase_first:nn                 2023-07-08
 \tl_case:cn                              2023-05-23
 \tl_case:cnF                             2023-05-23
 \tl_case:cnT                             2023-05-23
--- a/l3kernel/l3deprecation.dtx
+++ b/l3kernel/l3deprecation.dtx
@@ -549,6 +549,27 @@
 %    \end{macrocode}
 % \end{macro}
 %
+% \subsection{Deprecated \pkg{l3text} functions}
+%
+% \begin{macro}[EXP]{\text_titlecase:n, \text_titlecase_first:n}
+% \begin{macro}[EXP]{\text_titlecase:nn, \text_titlecase_first:nn}
+%    \begin{macrocode}
+\__kernel_patch_deprecation:nnNNpn { 2023-07-08 } { \text_titlecase_once:n }
+\cs_gset:Npn \text_titlecase:n #1
+  { \text_titlecase_once:n { \text_lowercase:n {#1} } }
+\__kernel_patch_deprecation:nnNNpn { 2023-07-08 } { \text_titlecase_once:n }
+\cs_gset:Npn \text_titlecase:n #1
+  { \text_titlecase_once:n {#1} }
+\__kernel_patch_deprecation:nnNNpn { 2023-07-08 } { \text_titlecase_once:nn }
+\cs_gset:Npn \text_titlecase:nn #1#2
+  { \text_titlecase_once:nn {#1} { \text_lowercase:n {#2} } }
+\__kernel_patch_deprecation:nnNNpn { 2023-07-08 } { \text_titlecase_once:nn }
+\cs_gset:Npn \text_titlecase_first:nn #1#2
+  { \text_titlecase_once:nn {#1} {#2} }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
 % \subsection{Deprecated \pkg{l3tl} functions}
 %
 %    \begin{macrocode}
--- a/l3kernel/l3text-case.dtx
+++ b/l3kernel/l3text-case.dtx
@@ -97,7 +97,7 @@
 \cs_new:Npn \text_titlecase:n #1
   { \@@_change_case:nnn { title } { } {#1} }
 \cs_new:Npn \text_titlecase_first:n #1
-  { \@@_change_case:nnn { titleonly } { } {#1} }
+  { \@@_change_case:nnn { title } { } {#1} }
 \cs_new:Npn \text_lowercase:nn #1#2
   { \@@_change_case:nnn { lower } {#1} {#2} }
 \cs_new:Npn \text_uppercase:nn #1#2
@@ -105,7 +105,7 @@
 \cs_new:Npn \text_titlecase:nn #1#2
   { \@@_change_case:nnn { title } {#1} {#2} }
 \cs_new:Npn \text_titlecase_first:nn #1#2
-  { \@@_change_case:nnn { titleonly } {#1} {#2} }
+  { \@@_change_case:nnn { title } {#1} {#2} }
 %    \end{macrocode}
 % \end{macro}
 % \end{macro}
@@ -159,40 +159,32 @@
 %   {
 %     \@@_change_case_switch_lower:nnNnnnn ,
 %     \@@_change_case_switch_upper:nnNnnnn ,
-%     \@@_change_case_switch_title:nnNnnnn ,
-%     \@@_change_case_switch_titleonly:nnNnnnn
+%     \@@_change_case_switch_title:nnNnnnn
 %   }
 % \begin{macro}[EXP]
 %   {
 %     \@@_change_case_letterlike_lower:nnN ,
 %     \@@_change_case_letterlike_upper:nnN ,
-%     \@@_change_case_letterlike_title:nnN ,
-%     \@@_change_case_letterlike_titleonly:nnN
+%     \@@_change_case_letterlike_title:nnN
 %   }
 % \begin{macro}[EXP]{\@@_change_case_letterlike:nnnnN}
 % \begin{macro}[EXP]
 %   {
 %     \@@_change_case_custom_lower:nnn ,
 %     \@@_change_case_custom_title:nnn ,
-%     \@@_change_case_custom_upper:nnn ,
-%     \@@_change_case_custom_titleonly:nnn
+%     \@@_change_case_custom_upper:nnn
 %   }
 % \begin{macro}[EXP]{\@@_change_case_custom:nnn}
 % \begin{macro}[EXP]
 %   {
 %     \@@_change_case_codepoint_lower:nnn ,
 %     \@@_change_case_codepoint_upper:nnn ,
-%     \@@_change_case_codepoint_title:nnn ,
-%     \@@_change_case_codepoint_titleonly:nnn
+%     \@@_change_case_codepoint_title:nnn
 %   }
 % \begin{macro}[EXP]{\@@_change_case_lower_sigma:nnnn}
 % \begin{macro}[EXP]{\@@_change_case_lower_sigma:nnnw}
 % \begin{macro}[EXP]{\@@_change_case_lower_sigma:nnnN}
-% \begin{macro}[EXP]
-%   {
-%     \@@_change_case_codepoint_title:nn     ,
-%     \@@_change_case_codepoint_titleonly:nn
-%   }
+% \begin{macro}[EXP]{\@@_change_case_codepoint_title:nn}
 % \begin{macro}[EXP]{\@@_change_case_codepoint_title:nnnn}
 % \begin{macro}[EXP]
 %   {\@@_change_case_codepoint:nnnn, \@@_change_case_codepoint_aux:nnnn}
@@ -205,10 +197,9 @@
 %   }
 % \begin{macro}[EXP]
 %   {
-%     \@@_change_case_next_lower:nn     ,
-%     \@@_change_case_next_upper:nn     ,
-%     \@@_change_case_next_title:nn     ,
-%     \@@_change_case_next_titleonly:nn ,
+%     \@@_change_case_next_lower:nn ,
+%     \@@_change_case_next_upper:nn ,
+%     \@@_change_case_next_title:nn ,
 %     \@@_change_case_next_end:nn
 %   }
 %   As for the expansion code, the business end of case changing is the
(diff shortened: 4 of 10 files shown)


josephwright:
@moewew  Sorry I've changed the plan a bit. This one I think would not require any immediate `biblatex` action, though you might want to shift over time.

josephwright:
@gmilde Hoepfully more to your taste: no immediate change to existing functions.

josephwright:
Note that I've not at present added a 'word exception' mechanism, though that is trivial if we stick with the 'word = between two spaces' approximation. Thoughts on this aspect welcome.

moewew:
So just to be sure: This implements `\text_titlecase_once`, which capitalises the first word and leaves the rest alone, and `\text_titlecase_all`, which capitalises all words? 

Does "capitalise" here means roughly (barring special cases like Dutch "IJ") uppercase first letter and leave the rest alone or does it mean uppercase first letter and lowercase the rest?

`biblatex` needs a function that capitalises the first word and lowercases all others, as that's our "sentence case". That would no longer be available with this change, right? (Currently that's what `\text_titlecase` does for us. We use `\text_titlecase_first` to capitalise the first word in a string and leave the rest alone.)

car222222:
Is it necessary to introduce "titlecase_once".? 
"once" seems too imprecise.
Is there a problem with using "titlecase_first"?  
Even more precise would be "titlecase_firstword".

josephwright:
@car222222  I was thinking of the existing `tl` functions, where we have `replace_once` and `replace_all`. I'm happy with `first` remaining if it is clear enough, and if that then makes sense with `all`. (I thought of including `word` in the naming, but it got a bit long.)

FrankMittelbach:
> @car222222 I was thinking of the existing `tl` functions, where we have `replace_once` and `replace_all`. I'm happy with `first` remaining if it is clear enough, and if that then makes sense with `all`. (I thought of including `word` in the naming, but it got a bit long.)

I think the argument of using consistent names is compelling even if "firstword" is slightly clearer if just seen without context or without knowledge of expl3. We use "once" always in the sense of "first occurance" and thus using the standard anme is on the whole better in my opinion.

zauguin (approved):

car222222:
@FrankMittelbach I am happy that "once" always means the first occurrence, but here there is nothing to indicate clearly what is occurring! Using "wordonce" (better "oneword") and "allwords" would be clearer. 

I fact replace_first would have been much better:-).

car222222:
In the case if replace, it is very likely that the one to be replaced (once) is the last occurrence rather than the first.

josephwright:
I'm happy if we want to go with `first`, but I would then look to adjust other occurrences.

car222222:
@josephwright Do you mean other occurrences of "once", 
or of "first"?

Note also that in this case we still need to indicate the object affected, i.e., "word", since the term "titlecase" does not obviously have much relationship with words.

josephwright:
> So just to be sure: This implements `\text_titlecase_once`, which capitalises the first word and leaves the rest alone, and `\text_titlecase_all`, which capitalises all words?

Correct.

> Does "capitalise" here means roughly (barring special cases like Dutch "IJ") uppercase first letter and leave the rest alone or does it mean uppercase first letter and lowercase the rest?

Broadly uppercase first, leave rest alone. @gmilde suggested that on balance when you look across different languages, leaving non-initial characters alone is likely best. (Also, Unicode only describe titlecase in terms of the action at the start of a word, and do not mention the effect on the remain of the input.)

> `biblatex` needs a function that capitalises the first word and lowercases all others, as that's our "sentence case". That would no longer be available with this change, right? (Currently that's what `\text_titlecase` does for us. We use `\text_titlecase_first` to capitalise the first word in a string and leave the rest alone.)

Current suggestion would retain `\text_titlecase:n` with unchanged behaviour by applying `\text_lowercase:n` inside` \text_titlecase_once:n`. However, that's relatively slow as it'stwo full passes. If it is overall decided that we need a function (or set of functions) which lowercase everything that's not titlecased, then I think a clearer name would be good. 'capitalise' has alredy been suggested,  doesn't get into 'sentence' questions, etc. So perhaps `\text_captialise_once:n` (or `first`) and `\text_capitalise_all:n`? (Likely implementation would be a faster version of the 'just apply lowercase first' plan.)

car222222:
But "capitalise" still does not convey that it is applied to words!
Can we have something that involves "word", please?

josephwright:
@car222222 On `titlecasing', I think the [Unicode FAQ](https://unicode.org/faq/casemap_charprop.html#4) are clear it's a word-based property:

> [Titlecase](https://www.unicode.org/glossary/#titlecase) takes its name from the case format used when forming a title, in which the initial letter in a word is capitalized and the rest are not. Titlecase is also used in forming a sentence by capitalizing the first word, and for forming proper names. The titlecase mapping in the [Unicode Standard](https://www.unicode.org/glossary/#unicode_standard) is the mapping applied to the initial character in a word.

I'm not sure how one could describe 'capitalisation' in a way that didn't link to words.

car222222:
capitalize: write or print (a word or letter) in capital letters

But also:  to write a letter of the alphabet as a capital, or to write the first letter of a word as a capital

And:  To capitalize a word is to make its first letter a capital letter

So again, it seems that getting "word" in there will remove all ambiguities.

car222222:
Something completely different, try:

_initialcap_allwords

_initialcap_firstword

josephwright:
@moewew  I wonder if it';s best for me to set up to internally-optimise the `\text_...case:n { \text_...case:n { ...} }` structure so we _don't_ need another name: I think `\text_titlecase_all:n { \text_lowercase:n { <input> } }` looks quite clear and usable.

moewew:
Hmmm, with my `biblatex` hat on I mainly care about having a tool to do the job, I don't really care whether I can use a predefined function or have to cook up my own combination of `\text_titlecase_first:n` and `\text_lowercase:n`. Of course it would be good if the implementation could be efficient (or at least not actively wasteful).

My personal opinion is that it would be cool to have the option to stack `\text_...` functions without significant performance hit, but my fear would be that that might be tricky to pull off if one wants to keep the code clean. I don't know what your approach is in LaTeX3 when it comes to balancing these kind of things.

car222222:
How can it get Signed Off now?  
Are you not still working on the code? 

When the code and names are decided, then it will need something to fill this void:

`% \begin{documentation}
%
% \end{documentation}`

josephwright:
> How can it get Signed Off now? Are you not still working on the code?

Code is finished at this point: I am not planning to include the 'skip words' idea at this stage.

> When the code and names are decided, then it will need something to fill this void:
> 
> ```
> % \begin{documentation}
> %
> % \end{documentation}
> ```

The documentation for all of the `l3text-...` files is in `l3text.dtx`.

car222222:
Do you should remove this empty environment from this file? 

Or put therein the reference to l3text.dtx (where I shall now check).

car222222:
Here is some ﻿further information on use of the term 'titlecase' in and around Unicode (and a little more generally).

@josephwright wrote (elsewhere):

the Unicode FAQ does use 'titlecase' for the idea of 'capitalising': https://unicode.org/faq/casemap_charprop.html#4
﻿
However, according to this FAQ (and my research) the Unicode 
documentation in reality uses the term only within these two 
specific terms:

1.  the name of a character mapping, “titlecase mapping”, which 
is the character mapping that is used by “Capitalisation”;

2.  the name of a Unicode “general category” of characters: 
Lt (letter, titlecase). 

The above referenced FAQ does indeed also allude (without a 
reference) to the possible use elsewhere of the term  “titlecase”, 
but it seems to clearly eschew (==not support) such usage within
Unicode documentation beyond the cases of the two terms listed above.

Elsewhere, the term is used, occasionally and inconsistently, for the somewhat ill-defined idea of transforming “a title” by uppercasing (or, in sophisticated cases, titlecasing) the start of some more or less well-defined selection of the words. 
Also occasionally, for a transformation that changes only the initial character(s) of a word or string (and is never applied to normal multi-word text).   

I could not find any examples of its use in relation to the capitalisation of all words in multi-word text.

car222222:
Note that in both of the Unicode use-cases, 'titlecase' is an adjective rather than a verb; and there is no meaning attached even to "making something titlecase".

josephwright:
@car222222  The only obvious alternative is `capitalise`, but then one looses the fact that it's not uppercasing the first(ish) character, and that it's explicitly a case-change function.

car222222:
@josephwright 

Yes, “capitalise”, (or “Capitalise”:-) seems OK,
but it still needs clear explanations:
-- of its exact meaning;
-- that it gets applied to words and;
-- that (in this implementation) it uses, as it should, the titlecase mapping.   

Looses?  Do you mean that it does not contain the subsequence “case”?

If it must be a verb containing the string “case”, then try:
 “initialcapcase”,   or   “capitalcase” —
but it will still need the same extensive explanation of what it is, and that it 
works on words.

jbezos:
The relevant point is:

> -- of its exact meaning;
> -- that it gets applied to words and;
> -- that (in this implementation) it uses, as it should, the titlecase mapping.

The name is not so important, provided the exact behavior is explained.

car222222:
@jbezos +1 Thanks.

josephwright:
Are we good-to-go here?

moewew:
To be honest, I got a bit lost here. Since we closed https://github.com/plk/biblatex/pull/1293 because plans changed, I haven't looked into the `biblatex` implications. We definitely need some time to adapt. Question is what the final plans in the kernel are.

l3kernel/l3text.dtx
```diff
@@ -115,11 +115,13 @@
...
-%   uppercase and the rest to lowercase. However, the process is more complex
+%   uppercase. However, the process is more complex
 %   than this as there are some situations where a single lowercase character
 %   maps to a special form, for example \texttt{ij} in Dutch which becomes
-%   \texttt{IJ}. The \texttt{titlecase_first} variant does not attempt
-%   any case changing at all after the first letter has been processed.
+%   \texttt{IJ}. There are two functions available for titlecasing: one which
+%   applies the change to each \enquote{word} and a second which only applies
+%   at the start of the input. (Here, \enquote{word} boundaries are spaces:
+%   at present full Unicode word breaking is not attempted.)
```
car222222, line 124: at present, full . . .

l3kernel/l3text.dtx
```diff
@@ -190,11 +192,7 @@
...
 %   \end{itemize}
 %
-%  For titlecasing, note that there are two functions available. The
-%  function \cs{text_titlecase:n} applies (broadly) uppercasing to the first
-%  letter of the input, then lowercasing to the remainder. In contrast,
-%  \cs{text_titlecase_first:n} \emph{only} carries out the uppercasing operation,
-%  and leaves the balance of the input unchanged. Determining whether
+%  Determining whether
 %  non-letter characters at the start of text should switch from upper- to
 %  lowercasing is controllable. When \cs{l_text_titlecase_check_letter_bool} is
```
car222222, line 197: I am not sure that the phrase 
"switch from . . . to lower casing"
is meaningful in this context!  
What "switching" is going on with "titlecasing"?

car222222 (approved):
Not sure that I follow all this documentation, but the stuff on "titlecase" specifically looks good.

l3kernel/l3text.dtx
josephwright: I've reworded: this is an area I'd like to go back to anyway as using catcode is not really right, and we now do have Unicode class available - but that's a different PR to this one :)

l3kernel/l3text.dtx
josephwright: Fixed

moewew:
There are a couple of references to `titlecase_once` (which is not defined in the current version of `expl3` on my machine), but then it is not documented (or implemented?) anywhere. Is that intended?

I don't think I'll manage to get the current GitHub `dev` version of `expl3` running in time, so I'll have to do this blind.

Can you please tell me what
```latex
\documentclass{article}

\ExplSyntaxOn
\def\test#1{\text_titlecase_first:n{\text_lowercase:n{#1}}}
\ExplSyntaxOff

\begin{document}
\test{Lorem ipsum Dolor sit Amet A}

\test{lorem ipsum Dolor sit Amet A}
\end{document}
```
produces with this PR merged?

josephwright:
> There are a couple of references to `titlecase_once` (which is not defined in the current version of `expl3` on my machine), but then it is not documented (or implemented?) anywhere. Is that intended?

No - I'll tidy those up - we had a bit of back-and-forth about `first` vs `once`.

> I don't think I'll manage to get the current GitHub `dev` version of `expl3` running in time, so I'll have to do this blind.

No pressure - we just did a release, I can sit on this for a while.

> \test{Lorem ipsum Dolor sit Amet A}
> \test{lorem ipsum Dolor sit Amet A}

Both will give `Lorem ipsum dolor sit amet a`. The overall feeling seemed to be we don't need to lowercase then titlecase in most applications, and that just titlecasing is enough. If that turns out to be problematic, I'll sort something a bit more integrated but where it's clear that `first`/`all` is about the 'words' and that the lowercasing is different.

moewew:
Can you tell me what
```latex
\documentclass{article}
\usepackage{csquotes}

\ExplSyntaxOn
\bool_set_false:N  \l_text_titlecase_check_letter_bool
\def\test#1{%
  \text_titlecase_first:n{\text_lowercase:n{#1}}\par
  \text_titlecase:n{#1}}
\ExplSyntaxOff

\begin{document}
\test{\enquote{lorem ipsum}}

\test{\enquote*{lorem ipsum}}
\end{document}
```
gives with the PR merged?

I get

> “Lorem ipsum”
>“Lorem ipsum”
>‘l’orem ipsum
>‘lorem ipsum’

picture of output reproduced above


where the "‘l’orem ipsum" is not what I want. That means that at least in the current version of the kernel the nested `\text_` calls are not a proper replacement for what we need. (So I cannot just switch `biblatex` to using that for good. We actually need some case distinction on how old the kernel is... presuming it works on dev.)

josephwright:
@moewew `'lorem ipsum'` for both. As the deprecation code already covers `\text_titlecase:n`, I'm not sure why you'd want `\text_titlecase_first:n { \text_lowercase:n { .. } }` at all.

moewew:
Hmmm.

To be honest, I'm a bit lost here. I just thought `\text_titlecase:n` was removed and we had to look for an alternative. Can we just keep on using it even though it is marked as deprecated?

josephwright:
@moewew We never remove any functions, although here there are some edge-case changes in behaviour as deprecation goes with shifting to emulation. If you need a function that lowercases first and want that to work before and after the update, something like
```latex
\cs_if_exist:NTF \text_titlecase_all:n
  {
    \cs_new:Npn \__mypkg_tilelcase:n #1
      { \text_titlecase_first:n { \text_lowercase:n {#1} } }
  }
  { \cs_new_eq:NN \__mypkg_tilelcase:n \text_titlecase_first:n }
```
would work.

However, one of the questions I ws trying to sort early on is to what extent that is actually required. The suggestion was that lowercasing 'the remainder' is likely not that useful, as the typical pattern is only to worry about the first character (`\text_fiflecase_first:n`) or the start of each word (`\text_titlecas_all:n`), leaving the other characters unchanged.

moewew:
> However, one of the questions I was trying to sort early on is to what extent that is actually required.

We need it to turn titles into sentence case. Entries in the `.bib` file are supposed to be given in title case and for styles that need it `biblatex` provides a function to convert the title into sentence case (as does BibTeX). I can't say anything about how useful such a function is in general, but `biblatex` definitely needs it.

FrankMittelbach:
@moewew what is the biblatex approach currently to handle uppercase letters that supposed to stay unchanged, say, something like IBM in the title?

moewew:
@FrankMittelbach The default is to use curly braces thanks to some very clever code by @josephwright and @blefloch that essentially turns `{...}` into `\NoCaseChange{...}` (see the long discussion in https://github.com/plk/biblatex/issues/960 and in particular https://github.com/moewew/biblatex/commit/c485f1b8079d56497b17343d7883b8fcd42f961a) - same as in classical BibTeX (it even gets the BibTeX exceptions broadly right). But users can also choose to not have curly braces imply case protection and use `\NoCaseChange` instead (I prefer this, because the curly braces thing gets messy what with them also being used for arguments).

@josephwright  Performance is a secondary concern, though I will admit that I have wondered how much the new two-pass scheme will effect performance. My primary concern is backwards compatibility. We need to retain the documented behaviour of `\MakeSentencecase` and friends and should ideally also preserve its current behaviour as closely as possible (though I could probably live with small deviations in edge cases - this whole thing is just to complicated).

I don't really care how this is implemented either on the `expl3` side or the `biblatex` side as long as the `biblatex` code follows best practices so that we can be as sure as one can be that what we do is supported for the foreseeable future, benefits from improvements in `expl3`, and won't break because of what we did. I thought one was not supposed to use deprecated functions, so I looked for a replacement and thought it would be `\text_titlecase_first:n { \text_lowercase:n { ... } }`. I noticed that those commands were available even in old kernel versions, which would make things easier for us, because we wouldn't have to branch on `expl3` versions. So I tried just replacing `\text_titlecase:n` with that. Then I found the unfortunate example with `csquotes`' `\enquote*` shown above that does not produce a good result with my `expl3` kernel, which suggested we can't just switch to `\text_titlecase_first:n { \text_lowercase:n { ... } }` for older kernels, because that would break backwards compatibility, as `\text_titlecase:n` does the right thing there. (All tested with `L3 programming layer <2023-08-29>` without this PR and not with the `dev` version.)

If using the deprecated function will not change behaviour significantly, will work in the foreseeable future and will not generate warnings that users could complain about I'm perfectly fine doing that. If using deprecated functions is generally frowned upon or may cause warnings (e.g. with stricter `expl3` code checking settings), I'd rather not go down that route, because sooner or later someone will complain.

josephwright:
@moewew  I think I can solve the `\enquote*` issue. There's an oversight in the current code if we 'break' with exactly one brace group left. Luckily, that can be patched such that it will be repaired for both the old and new approaches, and in a way you could simply add to `biblatex` as a 'hotfix' for a transition period:
```latex
\cs_gset:Npn \__text_change_case_break:w #1 \q__text_recursion_stop
  {
    \__text_change_case_break_aux:w ? #1
  }
\cs_gset:Npn \__text_change_case_break_aux:w #1 \q__text_recursion_tail
  {
    \__text_change_case_store:o { \use_none:n #1 }
    \__text_change_case_end:w
  }
```
With that, you should be able to use `\text_titlecase_first:n { \text_lowercas:n { ... } }` in an update. Would that be acceptable?

The issue I've been having is that 'titlecase' really means just changing the first char, and depending on the exact usage it may or may not be expected to lowercase the remainder of the input. I really would rather split the two concepts.

If there is a performance issue, I can see a way to do a lookahead and shortcut some of the code that would be repetitive.

Skillmon:
@josephwright random optimisation note: Use `\prg_do_nothing:` instead of `?` to protect against accidental brace-stripping, and just `\__text_change_case_store:o {#1}` without the `\use_none:n`.

josephwright:
@moewew  OK, I will hold off from a release until at least the start of next week - probably will do one soon-ish as this is a non-trivial change.

moewew:
If a new `biblatex` release is required to avoid things breaking, we will probably need more time, because we also need to release Biber (and that means waiting for the binaries to be built by other contributors), but maybe we don't actually need one, because `\text_titlecase:n` is not removed... (I guess I'll see if that's the case on Sunday.)

josephwright:
@moewew Ah, right, yes: I'll still wait to make sure there's nothing unexpected

moewew:
Prepared https://github.com/plk/biblatex/pull/1310 for `biblatex`. As far as I understood, we do not need to merge and release this immediately after the `expl3` update because you have backwards compatibility code in place to keep `\text_titlecase:n` around. Is that correct? Then I would wait for the `expl3` update to arrive on my machine, run a few more tests and see if anything changed.

josephwright:
@moewew  Sounds good: I will do a release today

PR #1250, opened 2023-07-16
Enable link breaking for dvips

josephwright:


l3backend/l3backend-pdf.dtx
```diff
@@ -571,38 +571,35 @@
...
-%   at some stage. At present this code is disabled as there is an open
-%   question about the name of the hook: to be resolved at the \LaTeXe{}
-%   end.
+% \begin{macro}{\@makecol}
+%   Install the necessary code to break links.
 %    \begin{macrocode}
-\use_none:n
+\bool_lazy_and:nnT
+  { \cs_if_exist_p:N \hook_gput_code:nnn }
+  { \cs_if_exist_p:N \@makecol }
```
FrankMittelbach, line 579: I think, for now hooking into the generic hook is fine. Eventually this should be moved to a kernel hook once \@makecol has one (not just in latex-lab)

l3backend/l3backend-pdf.dtx
```diff
@@ -571,38 +571,35 @@
...
   {
-    \cs_if_exist:NT \@makecol@hook
+    \hook_gput_code:nnn { cmd / @makecol / before }
+      { backend }
       {
-        \tl_put_right:Nn \@makecol@hook
+        \box_if_empty:NF \l_shipout_box
           {
```
FrankMittelbach, line 585: but I don't understand why \l_shipout_box should be non-empty at this point. I presume it is correct, but could there be perhaps a little explanation?

l3kernel/testfiles/m3expl001.ptex.tlg
```diff
@@ -170,6 +170,12 @@ Defining \__pdf_backend_link_minima: on line ...
 Defining \__pdf_backend_link_outerbox:n on line ...
 Defining \__pdf_backend_link_sf_save: on line ...
 Defining \__pdf_backend_link_sf_restore: on line ...
+Defining \__hook cmd/@makecol/before on line ...
+Defining \g__hook_cmd/@makecol/before_code_prop on line ...
+Defining \__hook_toplevel cmd/@makecol/before on line ...
+Defining \__hook_next cmd/@makecol/before on line ...
+Defining \g__hook_cmd/@makecol/before_labels_clist on line ...
+Defining \g__hook_cmd/@makecol/before_reversed_tl on line ...
```
FrankMittelbach, line 178: is there a way to actually test the functionality? Looks like all the tests do is showing that the hook got initialized

FrankMittelbach (commented):
just a couple of comment (basically saying I don't understand it :-)

l3backend/l3backend-pdf.dtx
josephwright: This may need re-working: I suspect it's got to be hard-coded box 255 at present.

l3kernel/testfiles/m3expl001.ptex.tlg
josephwright: A test would need to be at the PDF level: are we happy with that?

FrankMittelbach: dunno, but obviously the code is not working and the tests aren't in any way testing that it works. If we can't test it then the current tests could goo too I guess. If we can test "something" then we should do it :-)

car222222: @FrankMittelbach How do you know that "the code is not working"?
Did you test it somehow:-)?

FrankMittelbach: just from your description of what it should do ... and very clearly doesn't

l3backend/l3backend-pdf.dtx
u-fischer: I tried the following and it works fine, but imho without the pdfmanagement it is not of much use: hyperref then can't/doesn't use the backend commands and the l3pdfannot commands are not available. Wouldn't it then make more sense to tie it directly to the `\@makecol` redefinition in new-or-1 or -2?

~~~~
\DocumentMetadata{}
\documentclass{book}
\usepackage{hyperref}
\textwidth2cm
\begin{document}
xx \href{url}{long long long long long long long}

\ExplSyntaxOn
\pdfannot_link:nnn { URI }
{/A<</Type/Action/S/URI/URI(https://www.latex-project.org)>>}
{ a~very~long~link~with~lots~of~text }
\ExplSyntaxOff

\end{document}
~~~~

josephwright: > I tried the following and it works fine, but imho without the pdfmanagement it is not of much use: hyperref then can't/doesn't use the backend commands and the l3pdfannot commands are not available. Wouldn't it then make more sense to tie it directly to the `\@makecol` redefinition in new-or-1 or -2?
> 
> ```
> \DocumentMetadata{}
> \documentclass{book}
> \usepackage{hyperref}
> \textwidth2cm
> \begin{document}
> xx \href{url}{long long long long long long long}
> 
> \ExplSyntaxOn
> \pdfannot_link:nnn { URI }
> {/A<</Type/Action/S/URI/URI(https://www.latex-project.org)>>}
> { a~very~long~link~with~lots~of~text }
> \ExplSyntaxOff
> 
> \end{document}
> ```

The idea was that nothing should happen unless the user opts-in using the testphase: did I get that wrong?

l3kernel/testfiles/m3expl001.ptex.tlg
car222222: Or it does? ... at least when
 @u-fischer works good magic on it!

l3backend/l3backend-pdf.dtx
u-fischer: @josephwright sorry, I'm not quite sure what the question is here. But as soon as people use \DocumentMetadata they get the pdf management and hyperref uses it (and transparent and pdflscape). If you want to tie the breaking links to an opt-in, then the change to \@makecol should be made in new-or, then they get it only with a testphase key.

josephwright: @u-fischer I mean that the plan was that for `expl3`-based code, the change works 'out of the box' (as `expl3` hyperlinks are clearly 'new' material), but for `hyperref` it would only be people explicitly using the testphase, as otherwise `hyperref` doesn't use the `expl3` driver. That not correct?

l3kernel/testfiles/m3expl001.ptex.tlg
FrankMittelbach: @car222222 no. Ulrike looked at it *after* Joseph altered the right box :-)

josephwright: @FrankMittelbach  Nope, it works with the wrong box provided the link doesn't break between columns :)

FrankMittelbach: LOL

l3backend/l3backend-pdf.dtx
u-fischer: @josephwright no. People get the new hyperref driver as soon as they use the pdfmanagement. It is not possible otherwise. Packages that do PDF stuff *must* use it, or they wouldn't be compatible. This is the case since now around two years and nobody complained (some people explicitly use it only to get nicer hyperref colors ;-))

josephwright: @u-fischer  But that's _is_ testphase: until that management is built into the kernel, it remains opt-in, so most uses will not be impacted.

l3kernel/testfiles/m3expl001.ptex.tlg
car222222: I see!  Or rather I do not as the timing seems all wrong: maybe @josephwright used his time machine to change boxes?  
Or the box is itself a time-machine -- so lease call it \__Tardis_box ??

FrankMittelbach: if you look above to my other comment on the code, it says "outdated" and that is due to Joseph changing from `\l_shipout_box` to 255 by now (we don't get email notifications on l3kernel changes) so no time machine involved. Changing `\l_shipout_box`at this point didn't do anything good, because it isn't set up yet by the OR. This is what made me say the code can't work the way it is.

josephwright:
Following discussion, I'm closing here: we need the 'right' hooks for tagging, and once they are in place, they can also be used for link breaking.

PR #1271, opened 2023-09-26
Pivot toward e-type expansion

josephwright:
This PR 'pivots' toward `e`-type expansion, leaving `x`-type variants for third-party use. The aim is to cut down `x`-type to only those cases required, with a medium-term plan to revise these as they are (hopefully) removed 'in the wild'. Not covered here but a related idea is to minimise `f`-type variants in the same way. This would allow a more-or-less consistent set (`N`/`c` and `n`/`V`/`v`/`e`) for most functions, plus `o` where performance is important or for low-level use, and `f` only for places we need to 'do as TeX does' (e.g. lookahead in some places).

In terms of performance, I think we have minimal impact here. The core data structures use `\__kernel_tl_(g)set:N...` internally, and it remains `\let` to the primitive `\edef`. `\cs_set_protected:Npe`, etc., are also equivalent to current `\cs_set_protected:Npx`: again no performance hit. The `x`-to-`e` change here reflects the fact that `#` tokens do _not_ need to be doubled the arguments to these functions: a long-standing inconsistency when you look back. That plus the fact that `\use:e`, etc., makes life a lot clearer means that the consistency gain feels worth it for me.


l3experimental/l3graphics/l3graphics.dtx
```diff
@@ -236,7 +236,7 @@
     pagebox .choices:nn =
       { art , bleed , crop , media , trim }
       {
-        \tl_set:Nx \l_@@_pagebox_tl
+        \tl_set:Ne \l_@@_pagebox_tl
            { \tl_use:N \l_keys_choice_tl box }
```
Skillmon, line 240:
```suggestion
           { \l_keys_choice_tl box }
```

No checks whether `\l_keys_choice_tl` is defined necessary.

l3kernel/expl3.dtx
```diff
@@ -771,7 +774,7 @@
 % \end{verbatim}
```
Skillmon, line 774: line above uses the wrong variant string, should be `cne` to reflect the changes in the text above.

l3kernel/l3candidates.dtx
```diff
@@ -129,10 +129,11 @@
...
 % \begin{function}[added = 2018-04-01]
 %   {
-%     \tl_build_put_left:Nn, \tl_build_put_left:Nx,
-%     \tl_build_gput_left:Nn, \tl_build_gput_left:Nx,
-%     \tl_build_put_right:Nn, \tl_build_put_right:Nx,
-%     \tl_build_gput_right:Nn, \tl_build_gput_right:Nx,
+%     \tl_build_put_left:Nn, \tl_build_put_left:Ne,
+%     \tl_build_gput_left:Nn, \tl_build_gput_left:Ne,
+%     \tl_build_put_right:Nn, \tl_build_put_right:Ne,
+%     \tl_build_put_right:Nx,
```
Skillmon, line 135: Why is the `\tl_build_put_right:Nx` variant documented, but not `\tl_build_gput_right:Nx`? Imho, either document them both, or drop them both (even though there might be single packages in the wild using a then undocumented and deprecated function). Consistency might be key here. Even if both end up undocumented, `\cs_generate_variant:Nn` will still be able to produce them both, if someone so chooses.

l3kernel/l3cctab.dtx
```diff
@@ -543,7 +543,7 @@
       {
         \seq_gpop:NNF \g_@@_unused_seq \l_@@_internal_a_tl
           { \@@_begin_aux: }
-        \exp_args:Nx \@@_chk_group_begin:n
+        \@@_chk_group_begin:e
```
Skillmon, line 546: Why define a one-off `e` variant if the one-off `x`-variant wasn't necessary? The `\@@_chk_group_end:n` still uses `\exp_args:Ne`.

l3kernel/l3debug.dtx
```diff
@@ -838,22 +838,23 @@
...
       \tl_put_left:Nx
       \tl_put_right:Nn
       \tl_put_right:NV
       \tl_put_right:Nv
+      \tl_put_right:Ne
       \tl_put_right:No
       \tl_put_right:Nx
       \tl_build_begin:N
       \tl_build_put_right:Nn
-      \tl_build_put_right:Nx
```
Skillmon, line 856: `\tl_build_put_right:Nx` is still defined...

l3kernel/l3expan.dtx
```diff
@@ -665,7 +666,7 @@
 %     expansion.  At the beginning of an |f|-type argument, a space
 %     \meta{token} is removed even if it appears as \cs{exp_not:N}
 %     \cs{c_space_token}.  In an |x|-expanding definition
-%     (\cs{cs_new:Npx}), a macro parameter introduces an argument even
+%     (\cs{cs_new:Npe}), a macro parameter introduces an argument even
```
Skillmon, line 669: Line above is about `x` type...

l3kernel/l3expan.dtx
```diff
@@ -1345,6 +1346,13 @@
     \exp_after:wN #3
     \exp_after:wN { \exp:w \@@_eval_register:c {#4} }
   }
+\cs_new:Npn \exp_args:NNNe #1#2#3#4
```
Skillmon, line 1349: I wonder, whether 
```suggestion
\cs_new:Npn \exp_args:NNNe #1#2#3#4 { \tex_expanded:D { \exp_not:n {#1#2#3} {#4} } }`
```

might be faster (but didn't benchmark)

l3kernel/l3int.dtx
```diff
@@ -1386,7 +1387,7 @@
       }
     #2 = \@@_eval:w #1 \@@_eval_end:
```
Skillmon, line 1388: Why the `\@@_eval:w #1 \@@_eval_end:` here? `#1` is already fully evaluated by `\int_eval:n` at this point, a `#1 \exp_stop_f:` should suffice.

l3kernel/l3int.dtx
```diff
@@ -1386,7 +1387,7 @@
...
   }
-\cs_generate_variant:Nn \int_const:Nn { c }
```
Skillmon, line 1389: I think removing this is wrong here.

l3kernel/l3int.dtx
```diff
@@ -2050,7 +2051,7 @@
   {
     \int_compare:nNnTF {#1} > {#2}
       {
-        \exp_args:NNo \exp_args:No \@@_to_symbols:nnnn
```
Skillmon, line 2053: This change changes behaviour if `#3` contains material that might expand further (which I didn't check).

l3kernel/l3keys.dtx
```diff
@@ -216,7 +217,7 @@
 % \end{function}
 %
 % \begin{function}[added = 2011-08-21, updated = 2013-07-10]
-%   {.choices:nn, .choices:Vn, .choices:on, .choices:xn}
+%   {.choices:nn, .choices:Vn}
```
Skillmon, line 220: Not that I know of a use case, but why removing `.choices:xn` without providing `.choices:en`?

l3kernel/l3keys.dtx
```diff
@@ -424,7 +425,7 @@
 % \end{function}
 %
 % \begin{function}[added = 2011-08-21, updated = 2013-07-10]
-%   {.multichoices:nn, .multichoices:Vn, .multichoices:on, .multichoices:xn}
+%   {.multichoices:nn, .multichoices:Vn}
```
Skillmon, line 428: Again `x` removed without providing `e`.

l3kernel/l3keys.dtx
```diff
@@ -2518,17 +2517,19 @@
...
   { \@@_initialise:n {#1} }
 \cs_new_protected:cpn { \c_@@_props_root_str .initial:V } #1
   { \exp_args:NV \@@_initialise:n #1 }
+\cs_new_protected:cpn { \c_@@_props_root_str .initial:e } #1
+  { \exp_args:Ne \@@_initialise:n {#1} }
 \cs_new_protected:cpn { \c_@@_props_root_str .initial:o } #1
   { \exp_args:No \@@_initialise:n {#1} }
 \cs_new_protected:cpn { \c_@@_props_root_str .initial:x } #1
-  { \exp_args:Nx \@@_initialise:n {#1} }
+  { \exp_args:Ne \@@_initialise:n {#1} }
```
Skillmon, line 2532: This seems wrong, `.initial:x` should really use `x`-expansion internally.

l3kernel/l3cctab.dtx
josephwright: Looking back over the code as a whole, I suspect we'd be better defining variants and only use 'raw' `\exp-args:N...` where we _have_ to, e.g. `\exp_arg:NV #1` where `#1` is something passed from higher up where we cannot be sure of the arg spec at all. I was planning to look at that systematically later: I did. a few of them, but they were not all `x`-type related and I decided part-way through to come back to this.

l3kernel/l3candidates.dtx
josephwright: This really comes down to whether this step is a formal deprecation or simply a change of emphasise. We don't document formally deprecated functions but should cover everything else, even where we have non-consistent sets of functions. So it depends how others see the 'plan'.

l3kernel/l3debug.dtx
josephwright: Corrected: these were tricky as the list has to match the hand-tuned definitions.

l3kernel/l3expan.dtx
josephwright: Corrected

l3kernel/expl3.dtx
josephwright: Corrected

l3kernel/l3candidates.dtx
josephwright: Note that the aim here is to reduce the `x`-type variants as e do need to be mindful of hash table usage: pruning over time is part of the 'plan'.

l3kernel/l3expan.dtx
josephwright: I thought about that myself - I was aiming for a def. that was similar to the other hand-tuned cases. (I do wonder about whether we should reduce those down: I suspect it would be better overall to avoid the potentially-riskign hand-defining even if things end up a tiny bit slower.)

l3kernel/l3int.dtx
josephwright: Checking back. that line goes a long way back. We used to directly define `\in_const:Nn`, so this was required, but split things out so we force evaluation of the input once here (to deal with random values, etc.). So we could simply here: but that's independent of the PR so I'll address separately.

l3kernel/l3int.dtx
josephwright: I just moved where it is a bit: now on line 1369.

l3kernel/l3int.dtx
josephwright: I think `#3` here is supposed to comprise character tokens: lets see what others think.

l3kernel/l3keys.dtx
josephwright: I went here for reducing variants if possible, since we don't have all variants for all key properties. No one has used `.choices:xn`, so it seems it's not currently required to have a expanded variant. But it's a low barrier to having one if that is the overall feeling.

l3kernel/l3keys.dtx
josephwright: Same thing as previous comment,

l3kernel/l3keys.dtx
josephwright: True: must have missed that one cleaning up the search-and-replace changes - reverted.

l3kernel/l3candidates.dtx
muzimuzhi: Release 2022-01-12 claimed
https://github.com/latex3/latex3/blob/fedd9a9de989bf4a4156e1dae5e3f0f677520fa7/l3kernel/CHANGELOG.md?plain=1#L523-L525

Looks like another claim for function variants will appear soon.

josephwright: @muzimuzhi This one is tricky. On one hand, as long as the base function exists you can always use `\cs_generate_variant:Nn` to have the `x`-type versions. On the other, there is a risk of `Undefined control sequence`. I am hoping to see what others feel about this balance.

If the feeling is that we have to keep the `x`-type variants for ever (a reasonable position), then I can certainly go back over and re-add them all. There is then a question about hash table usage if one also wants a full set of `e`-type variants: personally I think we could live with that, but I know that this may not be the view of everyone.

josephwright: (We said we won't remove stuff, but didn't specify if we meant base functions or variants too - one way or the other we should tighten up)

josephwright: @Skillmon  I do see the argument for removing the docs for `x`-type variants, although `\exp_args:Nx` at least might have to stay (as it's used to create `x`-type variants themselves so is used by `l3kernel`). I got part-way through a change but think I need to see the overall feeling before I can move that forward: first lets see what the general  view is on the desirability of a change at all.

car222222: Roughly what is the current "hash table usage" of expl3?

josephwright: @car222222 Building the format for pdfLaTeX in the `e-type` branch gives:
```
22938 strings of total length 430829
1916723 memory locations dumped; current usage is 204&347958
21242 multiletter control sequences
```
which is as close as I can come to an estimate.

car222222: Thus, at an order of magnitude of 21242 (or even 10^5, the hash size does not seem to be a problem.  

At least not when compared with the per document requirements of a large document processed by latex-lab!

josephwright: OK, I'll adjust to keep the `x`-types but not document, combining suggestions :)

car222222: Extra info: a recent (2023) setup of pdftex+miktex seems to provide the following size of hash-table: 15000+200000.

So running with expl3 plus a few "heavyweight packages" will typically require over 30% of this total before adding anything from the document itself (such as 10 or even 100 per page for tagging, etc.)

But note that most of this will now be in the "non-hashed" part of the table so access will be slower.

josephwright: Done: only costs around 90 csnames

car222222: A mere drop in the ocean!

josephwright: @muzimuzhi Hopefully you prefer the updated version :)

muzimuzhi: Personally I can live with one-time dropping followed by a claim that no variants will be removed (only deprecated) from now on. Letting breakage plays the role of `e`-type promoter? But yes an instant no-variant-removal policy gains most backwards compatibility for class and package authors.

josephwright: @muzimuzhi I think the plan of adding `\exp_args:N..x` to the deprecation module will do that for the 'interested' programmer -they'll get the warning, end users won't. I'll look at that later today/tomorrow.

josephwright:
Note that the easiest way to trap the deprecation of `x-types here is to apply to `\use:x`, as other than `\cs_...:N(p)x` (which are defined directly), all of the `\exp_args:N...x` functions use `\use:x` internally. However, at present parts of the 2e kernel use `x`-type variants, so we'd have a broken setup if we changed at this stage. So this is a 'phase II' change.

Skillmon:
Another view on performance: This might net to actually being faster, since most `x`-type functions weren't (and still aren't) really using the low level `\edef`-behaviour to the fullest extend, but are often `\cs_generate_variant:Nn ... { ... x }`, and that is slower than an `e`-variant.

josephwright:
@FrankMittelbach I'm wondering on your take here ...

davidcarlisle (approved):

PR #1280, opened 2023-10-15
Detect letters for titlecasing based on Unicode general category

### LINKED ISSUE #984: \text_titlecase:n and string variables
The `\text_titlecase:n` command does not seem to work correctly with string variables. The following document gives me the output “ABC”, but it should be “Abc”.
```
\documentclass{article}
\begin{document}
\ExplSyntaxOn
\str_new:N \l_test_str
\str_set:Nn \l_test_str {abc}
\text_titlecase:n {\l_test_str}
\ExplSyntaxOff
\end{document}
```
Version numbers:
```
This is XeTeX, Version 3.141592653-2.6-0.999993 (TeX Live 2021) (preloaded format=xelatex)
LaTeX2e <2021-06-01> patch level 1
L3 programming layer <2021-08-27>
```

josephwright:
Fixes #984


josephwright:
@moewew  You might have a view here

moewew:
Not sure I understand all implications of this, but from what I understand it appears sensible to me...

josephwright:
I get the feeling this might be a 'suck it an see' PR - go for it?

l3kernel/l3text-case.dtx
```diff
@@ -692,40 +696,39 @@
     \@@_change_case_loop:nnnw {#2} {#3} {#4} #5
   }
 %    \end{macrocode}
-%   For titlecasing, we need to fully expand the new character to see if it
-%   is a letter (or active).
+%   For titlecasing, we need to fully expand the new codepoint to see if it
```
car222222, line 699: Is it clear what "a codepoint" means here?  And how/when such a codepoint can be expanded? (Only functions and active characters are expandable.)

l3kernel/l3text.dtx
```diff
@@ -188,13 +188,12 @@
 %   \end{itemize}
 %
 %  Determining whether non-letter characters at the start of text should count
-%  as the uppercase element is controllable. When \cs{l_text_titlecase_check_letter_bool} is
-%  \texttt{true}, characters which are not letters (category code~$11$) are
+%  as the uppercase element is controllable. When
+%  \cs{l_text_titlecase_check_letter_bool} is \texttt{true}, codepoints which are
```
car222222, line 192: "codepoints" again!

l3kernel/l3text.dtx
```diff
@@ -188,13 +188,12 @@
...
+%  not letters (Unicode general category \texttt{L}) are
 %  \enquote{skipped}: the first \emph{letter} is uppercased.
```
car222222, line 194: not changed and only the first \emph{letter} found is uppercased.

car222222 (approved):
A few suggestions on the wording.

I assume that all the outcomes of all the provided tests have been already checked: full testing (even of each Unicode slot) would be very exhausting! :-) 

So "suck it and see" sounds good.

l3kernel/l3text-case.dtx
josephwright: Changed

l3kernel/l3text.dtx
josephwright: Changed

l3kernel/l3text.dtx
josephwright: Yes, but this is true here: I'm not sure what else i could say

car222222: Yes sure.  I was just pointing it out.

The basic idea is to  have somewhere clear definitions of how you are using the terms "Unicode character" and "Unicode codepoint" and emphasising the distinction, if there is any!  Also a note that the "Unicode" is normally implied from the context.

PR #1287, opened 2023-10-19
Move tl-build functionality to stable

josephwright:


l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+%   results.  The |begin| and |gbegin| functions must be used for local
+%   and global \meta{tl~var} respectively.
+% \end{function}
+%
+% \begin{function}[added = 2018-04-01]
+%   {
+%     \tl_build_put_left:Nn, \tl_build_put_left:Ne,
+%     \tl_build_gput_left:Nn, \tl_build_gput_left:Ne,
+%     \tl_build_put_right:Nn, \tl_build_put_right:Ne,
+%     \tl_build_gput_right:Nn, \tl_build_gput_right:Ne
```
muzimuzhi, line 79: Are more variants needed? Maybe not as many as `\tl_put_left:Nn` variants `V|v|e|o`.
https://github.com/latex3/latex3/blob/2eebe69a42db66364efe7008185cdfe32da22c30/l3kernel/l3tl.dtx#L182-L192

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+      }
+  }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}
+%   {
+%     \tl_build_put_right:Nn, \tl_build_put_right:Ne, \tl_build_put_right:Nx,
+%     \tl_build_gput_right:Nn, \tl_build_gput_right:Ne, \tl_build_gput_right:Nx
```
muzimuzhi, line 175: These `x`-variants `\tl_build_(g)put_right:Nx` are not documented. Or maybe they can be dropped?

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+%
+% \begin{macro}
+%   {
+%     \tl_build_put_left:Nn, \tl_build_put_left:Ne, \tl_build_put_left:Nx,
+%     \tl_build_gput_left:Nn, \tl_build_gput_left:Ne, \tl_build_gput_left:Nx
+%   }
+% \begin{macro}{\@@_build_put_left:NNn}
+%   See \cs{tl_build_put_right:Nn} for all the machinery.  We could
+%   easily provide \cs[no-index]{tl_build_put_left_right:Nnn}, by just
+%   add the \meta{right} material after the \Arg{left} in the
```
muzimuzhi, line 244: add -> adding

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+  {
+    \if_false: { { \fi:
+          \exp_end: \exp_end: \exp_end: \exp_end: \exp_end:
+          \@@_build_last:NNn #1 #2 { }
+        }
+      }
+    \if_meaning:w \c_empty_tl #2
+      \@@_build_begin:NN #1 #2
+    \fi:
+    #1 #2
```
muzimuzhi, line 220: If `#2` is empty, it will be redefined twice, once in `\__tl_build_begin:NN` on line 218, then on line 220. But introducing a helper function seems to expose the internal representation (appending prime in csname).

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+%
+% When creating a \meta{tl~var} by accumulation of many tokens, the performance
+% available using a combination of \cs{tl_set:Nn} and \cs{tl_put_right:Nn} or
+% similar begins to become an issue. To address this, a set of functions are
+% available to \enquote{build} a \meta{tl~var}. The performance of this approach
+% is much more efficient than the standard \cs{tl_put_right:Nn}, but the
+% constructed token list cannot be accessed during construction other than
+% by methods provided in this section.
+%
+% \begin{function}[added = 2018-04-01]{\tl_build_begin:N, \tl_build_gbegin:N}
```
muzimuzhi, line 62: Do the `\tl_build_(g)begin:N` and `\tl_build_(g)end:N` need `c` variants? If so they can be hand-tuned to internally use quicker `\tl_to_str:n` on `c`-arg instead of `\cs_to_str:n` on `N`-arg (didn't check the `end` part).

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+% \begin{macro}{\@@_build_end_loop:NN}
+%   Get the data then clear the \meta{next~tl} recursively until finding
+%   an empty one.  It is perhaps wasteful to repeatedly use
+%   \cs{cs_to_str:N}.  The local/global scope is checked by
+%   \cs{tl_set:Ne} or \cs{tl_gset:Ne}.
+%    \begin{macrocode}
+\cs_new_protected:Npn \tl_build_end:N #1
+  {
+    \@@_build_get:NNN \__kernel_tl_set:Ne #1 #1
+    \exp_args:Nc \@@_build_end_loop:NN { \cs_to_str:N #1 ' } \tl_clear:N
```
muzimuzhi, line 278: Strictly speaking `<next tl>`s are not named like standard latex3 tl variables, but yes there're no better choice than `\tl_(g)clear:N`...

muzimuzhi (changes requested):
Mostly about variants.

BTW, will users expect a `\tl_build_show:N`?

FrankMittelbach:
> Mostly about variants.
> 
> BTW, will users expect a `\tl_build_show:N`?

I think it would make some sense for debugging purposes.

l3kernel/l3tl-build.dtx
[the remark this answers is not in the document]
josephwright: This is by-design: we've moved from `x`- to `e`-type, deprecating `x`-type. That means they stay available but no documented: we've said there will be no more removals from `l3kernel`. The fact these were 'candidates' might give us some wiggle room, but as they have been in use the safe position is not to remove them.

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+%   and global \meta{tl~var} respectively.
+% \end{function}
+%
+% \begin{function}[added = 2018-04-01]
+%   {
+%     \tl_build_put_left:Nn, \tl_build_put_left:Ne,
+%     \tl_build_gput_left:Nn, \tl_build_gput_left:Ne,
+%     \tl_build_put_right:Nn, \tl_build_put_right:Ne,
+%     \tl_build_gput_right:Nn, \tl_build_gput_right:Ne
+%   }
```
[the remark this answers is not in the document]
josephwright: Possibly yes: there's a tension between adding all variants and the impact on the hash table. So here we need to see what variants are actually needed.

josephwright:
> > Mostly about variants.
> > BTW, will users expect a `\tl_build_show:N`?
> 
> I think it would make some sense for debugging purposes.

But wouldn't that just be `get` + the std `show`, or whatever we decide for `get` (I will make a second PR for that).

l3kernel/l3tl-build.dtx
```diff
@@ -0,0 +1,329 @@
...
+%   \cs[no-index]{cs_(g)set_nopar:Npe} rather than
+%   \cs[no-index]{tl_(g)set:Ne} partly for the same reason and partly
+%   because the assignments are interrupted by brace tricks, which
+%   implies that the assignment does not simply set the token list to an
+%   |e|-expansion of the second argument.
+%    \begin{macrocode}
+\cs_new_protected:Npn \tl_build_put_right:Nn #1#2
+  {
+    \cs_set_nopar:Npe #1
+      { \exp_after:wN \exp_not:n \exp_after:wN { \exp:w #1 #2 } }
```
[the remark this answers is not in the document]
josephwright: Yes, true: I will adjust. We do want the maximum efficiency here.

l3kernel/l3tl-build.dtx
josephwright: At the moment, I think this again falls into the 'see use in practice' category - it feels unlikely we need `c`-type variants, so probably should wait and see.

l3kernel/l3tl-build.dtx
josephwright: That's true (and documented) but `\tl_clear:N` only requires the argument is `N`-type, not that it follows the naming convention. For a start, we need a way to handle document-level stuff ...

l3kernel/l3tl-build.dtx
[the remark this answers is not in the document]
car222222: I thought we had concluded that affects such as these on the hash table size ate not a concern, since they are trivial compared with the current size  of its components: both that created by the format itself, and the potentially enormous extra consequences of a complex document!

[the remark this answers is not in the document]
josephwright: Yes, in general but we've not (yet) done a totally comprehensive addition of all variants for `l3kernel` (perhaps limited to functions of three arguments or fewer). I'm happy if we want to do that, but that should be separate from this PR :)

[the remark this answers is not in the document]
car222222: Yes, no such considerations should have been introduced into this PR:-)!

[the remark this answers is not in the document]
josephwright: @car222222  I agree it's probably best if we move toward providing 'the lot' out of the box - it's mainly mechanical work and I can sort - we did have an issue but that was more focussed, perhaps open a new one?

muzimuzhi:
> But wouldn't that just be `get` + the std `show`, or whatever we decide for `get` (I will make a second PR for that).

Oh, I missed the `\tl_build_get:NN` which is still in `l3candidates.dtx`. Then the question becomes, why it's the only tl-build function staying non-stable?

FrankMittelbach:
> > > Mostly about variants.
> > > BTW, will users expect a `\tl_build_show:N`?
> > 
> > 
> > I think it would make some sense for debugging purposes.
> 
> But wouldn't that just be `get` + the std `show`, or whatever we decide for `get` (I will make a second PR for that).

sure, but that is true for most if not all `_show` functions they aren't usually complicated but by convention something is there to show (and log)

josephwright:
> > But wouldn't that just be `get` + the std `show`, or whatever we decide for `get` (I will make a second PR for that).
> 
> Oh, I missed the `\tl_build_get:NN` which is still in `l3candidates.dtx`. Then the question becomes, why it's the only tl-build function staying non-stable?

Three is a question about how to handle this: I will have a second PR sorted later today with one possible approach.

josephwright:
@muzimuzhi https://github.com/latex3/latex3/pull/1302 shows what I think might work better than `get`

josephwright:
I get the feeling no one objects to the basic idea here

PR #1288, opened 2023-10-20
Fix doc typos mainly for fp var and funcs

muzimuzhi:
About `\begin{quote}\let\obeyedline=\newline\obeylines^^A`
  - Compared to inserting `\\` per line, would this way be too tricky?
  - `verbatim` env is another option, but auto indexing is disabled.
  - By default `\obeyedline` uses `\par`, which will make larger visual "line spacing". Hence I redefined it (locally) to use `\newline`.

About `\subsection` -> `\section`, I'm also considering changing their titles or even restructuring them. Maybe in a later PR.
  - It's a bit hard to infer from only the title "Symbolic expressions" to know that expl3 functions about fp expr variables are documented under it.
  - And since "function" is also used to represent an expl3 function, the second title "User-defined functions" is more or less ambiguous.

### DIFF
--- a/l3experimental/l3str/l3str-format.dtx
+++ b/l3experimental/l3str/l3str-format.dtx
@@ -123,7 +123,7 @@
 %
 % \begin{function}[EXP]{\fp_format:nn}
 %   \begin{syntax}
-%     \cs{fp_format:nn} \Arg{fpexpr} \Arg{format specification}
+%     \cs{fp_format:nn} \Arg{fp expr} \Arg{format specification}
 %   \end{syntax}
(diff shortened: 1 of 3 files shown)


FrankMittelbach:
> * It's a bit hard to infer from only the title "Symbolic expressions" to know that expl3 functions about fp expr variables are documented under it.
> 
>     * And since "function" is also used to represent an expl3 function, the second title "User-defined functions" is more or less ambiguous.

well, one could perhaps explicitly adding "fp" or "floating point" in the headings, eg "User-defined fp-functions"

PR #1291, opened 2023-10-23
reserve tbl prefix

u-fischer:

PR #1295, opened 2023-10-23
Split CI into smaller jobs

muzimuzhi:
Currently CI workflow ([`main.yaml`][main.yaml]) ran two jobs (1 check + 1 doc) in parallel and took typically 20--25 minutes, see https://github.com/latex3/latex3/actions/workflows/main.yaml?query=is%3Asuccess.
- By running `l3build check` per bundle, 5 jobs (4 check + 1 doc) will run in parallel.
  For example in https://github.com/muzimuzhi/latex3/actions/runs/6616980533, each `l3build check` job finished in 10 min and the `l3build doc` job took ~16 min, hence the whole CI finished in ~16 min.
- By further running `l3build doc` per bundle, 8 jobs (4 check + 4 doc, but on uneven bundle list) will run in parallel.
  For example in https://github.com/muzimuzhi/latex3/actions/runs/6617727855, the whole CI now finished in ~10 min.

Short-come: If the head and base branches of a PR are in the same repository (e.g. #1291), 16 jobs (8 triggered by `push`, 8 by `pull_request`) will run in parallel for each push, making a long check list. (latex3/latex2e now uses 31 `l3build check` jobs, see for example [this run][latex2e-ci-run]) Fortunately, the number of maximum concurrent jobs supported by GitHub Free plan, on Standard GitHub-hosted runners is 20, larger than 16.

[main.yaml]: https://github.com/latex3/latex3/blob/62c5afee35e44d3a9db4c9432103349251a92a1c/.github/workflows/main.yaml
[latex2e-ci-run]: https://github.com/latex3/latex2e/actions/runs/6613792207


josephwright:
Not keen - but let's see what others think

muzimuzhi:
My motivation: When doing experimental changes (in non-default branches in forks), quicker feedbacks received from CI is better.

FrankMittelbach:
The docs are our next bottleneck, so I'm not adverse to seeing it split eventually. The problem is, however, that the list sizes are rather uneven and that will probably get worse not better.

muzimuzhi:
> The problem is, however, that the list sizes are rather uneven and that will probably get worse not better.

Yes. `l3build check` can use `--first` and `--last` to further split a bundle, but currently I see no way (?) to do similar things to `l3build doc`. For example to build `source3.pdf` in its own job, `l3build doc` need to learn how to exclude a single doc name.

I almost opened an issue in l3build, requesting negative glob pattern in variables and in `l3build doc "!<name>"` when preparing this PR... .

PR #1296, opened 2023-10-23
Docu `\ExplLoaderFileDate` and `\c__kernel_expl_date_tl`

muzimuzhi:
As `\ExplLoaderFileDate` is already used to define `\IfExplAtLeastTF` in LaTeX2e (see https://github.com/latex3/latex2e/pull/1146), latex3 may want it and its expl3 alias `\c__kernel_expl_date_tl` documented somewhere.

Since they're both defined in loader file (`expl3.ltx`, `expl3.sty`, and `expl3-generic.tex`), I think `expl3.pdf` is more appropriate a place than `interface3.pdf` for their documentations. `interface3.pdf` never even mentions "loader(s)".

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+- Documentation for `\ExplLoaderFileDate` in `expl3.pdf`
+
 ## [2023-10-23]
 
 ### Added
--- a/l3kernel/expl3.dtx
+++ b/l3kernel/expl3.dtx
@@ -958,6 +958,16 @@
 % loadable: package loading is dependent on the \LaTeXe{} package-management
 % mechanism.
 %
+% \section{Getting the version of \pkg{expl3}}
+%
+% \begin{function}{\ExplLoaderFileDate}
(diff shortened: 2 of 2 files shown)


josephwright:
Internal 'kernel-wide' stuff is documented in `l3kernel-functions` and definitely not in `expl3.dtx`.

muzimuzhi:
Yes I find the implementation doc for `\c__kernel_expl_date_tl` in `l3kernel-function.dtx`. It seems there's no need to document `\c__kernel_expl_date_tl` elsewhere, and definitely I didn't notice it's an internal variable.
https://github.com/latex3/latex3/blob/62c5afee35e44d3a9db4c9432103349251a92a1c/l3kernel/l3kernel-functions.dtx#L168-L171

Is `\ExplLoaderFileDate` considered internal too?

muzimuzhi:
> Is `\ExplLoaderFileDate` considered internal too?

If so then maybe what's only needed is
```diff
- % \begin{variable}{\c__kernel_expl_date_tl}
+ % \begin{variable}{\ExplLoaderFileDate, \c__kernel_expl_date_tl}
```

josephwright:
I'd say that `\Expl...` can't be regarded as internal: this is a package-author (Camel Case) name

PR #1302, opened 2023-10-25
Add suspend/resume mechanism for tl-build

josephwright:


josephwright:
(For ref., the issue with `get` is that it's not currently named to be clear it gets the intermediate/partial token list: I think a suspend/resume approach is more obvious in this sense.)

FrankMittelbach:
I would have thought the suggestion of something like "get_intermediate" would be clear and then you don't need any resuming as the datastructure would still be there to accept more build commands. Just a thought.

josephwright:
@FrankMittelbach  I'll make a parallel PR that does that, and we'll see which overall feels better :)

josephwright:
Decided against this approach.

PR #1304, opened 2023-10-25
Move \tl_build_get_intermediate:NN to stable

josephwright:
Parallel to https://github.com/latex3/latex3/pull/1302


muzimuzhi:
The `added` date and position of changelog entries need to be updated/adjusted.

PR #1319, opened 2023-11-04
Document `\c_nan_fp`

muzimuzhi:
It was added in commit fb3f38ff (First stage of installing new FPU: copy code and test files into l3kernel, 2012-05-09), along with other fp constants.

Reading l3fp implementation slowly, I found `\c_nan_fp` the only fp constant (in a group of five) that's not documented.

If user wants to process the expansion result of an `<fp expr>`, `\c_nan_fp` is needed in one case of `\str_case_e:nn(TF) {<fp expr>} {<string and code cases>}`.
image

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+- Documentation for `\c_nan_fp`
+
 ### Fixed
 - Normalisation of `.inherit:n` key data (issue \#1314)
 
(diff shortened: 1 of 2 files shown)


muzimuzhi:
CI failed due to a checksum error, which I've encountered [once 11 hours ago](https://github.com/muzimuzhi/latex3/actions/runs/6750919947/job/18354149419), when downloading from the same CTAN mirror.

> /home/runner/texlive/bin/x86_64-linux/tlmgr: checksum error when downloading /tmp/JLIRlLiIlB/Olc52lTwMq from https://us.mirrors.cicku.me/ctan/systems/texlive/tlnet/tlpkg/texlive.tlpdb: digest disagree
> Error: Process completed with exit code 1.

Force pushing to rerun workflows.

PR #1321, opened 2023-11-05
Improve l3color and (x)color coexistance in LuaLaTeX

zauguin:
Currently the `l3color` code interacting with luaotfload assumes that all colors are defined with l3color.
While this is a reasonable long-term goal, it is more practical to assume that some user might also use (x)color to define color names.

Therefore this changes the luaotfload hook to use the (x)color color definition if no color with the same name is defined in l3color.

### DIFF
--- a/l3backend/l3backend-color.dtx
+++ b/l3backend/l3backend-color.dtx
@@ -1308,6 +1308,16 @@ if luaotfload and luaotfload.set_transparent_colorstack then
     local html = htmlcolor:match(value)
     if html then return html end
 
+% If no l3color named color with this name is known, check for defined xcolor colors
+    local l3color_prop = token.get_macro(string.format('l__color_named_%s_prop', value))
+    if l3color_prop == nil or l3color_prop == '' then
+      local legacy_color_macro = token.create(string.format('\\color@%s', value))
+      if legacy_color_macro.cmdname ~= 'undefined_cs' then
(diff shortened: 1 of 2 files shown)


l3backend/l3backend-color.dtx
```diff
@@ -1308,6 +1308,16 @@ if luaotfload and luaotfload.set_transparent_colorstack then
     local html = htmlcolor:match(value)
     if html then return html end
 
+% If no l3color named color with this name is known, check for defined xcolor colors
+    local l3color_prop = token.get_macro(string.format('l__color_named_%s_prop', value))
```
muzimuzhi, line 1312: Similar to commit 096d4ec6 (Use @@ convention in Lua opacity backend code, 2023-11-05). Current file has set
https://github.com/latex3/latex3/blob/69a354502548f0a9ff3f4df8b906d971f86b4fab/l3backend/l3backend-color.dtx#L60

```suggestion
    local l3color_prop = token.get_macro(string.format('l_@@_named_%s_prop', value))
```

muzimuzhi (commented):
Need a test (in `./l3kernel/testfiles-backend` directory)? But it's luatex only...

josephwright (approved):
In many ways I'd rather handle this from the macro end, but I see that right now this works, so will approve but we should remember to revisit once other changes are made.

PR #1335, opened 2023-11-14
Templates

josephwright:
This PR will allow loading of `template` code in the kernel, with only a very thin wrapper left in `l3packages`.


l3packages/xtemplate/xtemplate-2023-10-10.sty
```diff
@@ -0,0 +1,1373 @@
+%%
+%% This is file `xtemplate.sty',
```
FrankMittelbach, line 2: I think such files should identify themselves as a frozen version, e.g., for array-2016-... there is

%%
%% This is an old release of `array.sty',
%%

at the top

josephwright:
I assume everyone is mainly focussing on https://github.com/latex3/latex2e/pull/1132, so provided it fits together are happy with the PR here?

car222222 (approved):

josephwright:
I'll wait for a decision on the 2e PR before merging/releasing: the key is to have the right names set up, which depends what we actually decide to do in 2e!

l3packages/xtemplate/xtemplate.dtx
```diff
@@ -707,2225 +690,47 @@
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-%<@@=xtemplate>
+%<@@=template>
```
muzimuzhi, line 693: A latex3 prefix renaming. Corresponding entries in [`l3prefixes.csv`](https://github.com/latex3/latex3/blob/main/l3kernel/doc/l3prefixes.csv) need an update?

FrankMittelbach: We should not free xtemplate in my opinion but reserve both.

josephwright: We never release prefixes once taken, as older code will still use them. As the change is for compatibility, I don't think we need an update: someone finding `xtemplate` in an internal macro is still best pointed to the L3 repo, and someone finding `template` to the 2e one.

PR #1338, opened 2023-11-15
Move l3graphics to l3kernel (stable)

josephwright:
Functionality here is equivalent to `graphicx` and would allow a move of the latter to use the updated driver framework. 

Note at present no `added = ...` dates but I will do those if the PR is approved.


josephwright:
Updated for @muzimuzhi :)

l3kernel/testfiles-plain/m3plain002.tlg
```diff
@@ -5,7 +5,7 @@ Don't change this file in any respect.
 {retaining \__opacity_backend_fill_stroke:ee=\protected\long macro:->\exp_args:Nee \__opacity_backend_fill_stroke:nn }
 {retaining \__pdf_backend_object_now:ne=\protected\long macro:->\exp_args:Nne \__pdf_backend_object_now:nn }
 {retaining \__pdf_backend_object_write:nne=\protected\long macro:->\exp_args:Nnne \__pdf_backend_object_write:nnn }
-{restoring \l_graphics_search_ext_seq=undefined}
+{restoring \l_graphics_search_ext_seq=macro:->\s__seq }
```
muzimuzhi, line 8: Evidence that the timing issue in #1469 is going to be resolves with this PR.

josephwright:
@muzimuzhi If you want to help move this forward, some tests of `l3graphics` showing it can do the same as `graphicx` would be handy :)

l3kernel/testfiles-backend/m3graphics001.lvt
```diff
@@ -5,13 +5,11 @@
 \documentclass{minimal}
 \input{regression-test}
 
-\RequirePackage{expl3}
+\RequirePackage{expl3} % As we need to pick up options
 \ExplSyntaxOn
```
muzimuzhi, line 9: What really matters here is to load backend driver before using `l3graphics` functions, so using `\sys_load_backend:n {}` or `\sys_ensure_backend:` works too.
```suggestion
\ExplSyntaxOn
\sys_load_backend:n {}
```

josephwright: No, that doesn't work - the way the test is set up uses `\PassOptionsToPackage` here, so currently we need the package loading

josephwright: Sorted by updating the general setup on main then rebasing again with a change here :)

josephwright:
@muzimuzhi If you want to help move this forward, some tests of `l3graphics` showing it can do the same as `graphicx` would be handy :)

muzimuzhi:
> @muzimuzhi If you want to help move this forward, some tests of `l3graphics` showing it can do the same as `graphicx` would be handy :)

To contribute tests in a PR, what's the suggested base branch, the `main` or `l3graphics` (which is the head branch of current PR).

l3kernel/CHANGELOG.md
```diff
@@ -7,18 +7,57 @@ this project uses date-based 'snapshot' version identifiers.
...
+    `\keys_set:nn` breaking some nested usages in edge cases
+
+## [2025-01-14]
+
+### Added
+- `\int_step_tokens:nn`, `\int_step_tokens:nnn`, and `\int_step_tokens:nnnn`
+
+### Fixed
+- Check conditionals are defined when creating variants (see \#1189)
+
```
davidcarlisle, line 48: As shown here (probably as a result of being open too long) this PR appears to have picked up many unrelated changes It could probably benefit from being rebased. that said I have looked at the files locally with the l3graphics files moved to l3kernel (and backend) and it all looks good so I think I'll approve as is.

davidcarlisle (approved):
See inline comment, but approving

l3kernel/CHANGELOG.md
josephwright: Thanks - I'll tidy up before a merge

PR #1341, opened 2023-11-15
Move l3bitset to l3kernel (stable)

josephwright:


l3kernel/doc/source3body.tex
```diff
@@ -588,6 +588,7 @@ \part{Data types}
     l3fp-functions.dtx
   }
 \ExplSyntaxOff
+\DocInput{l3bitset.dtx}
```
muzimuzhi, line 591: Any better position than just between `fp` submodules and `fparray`?

l3kernel/l3bitset.dtx
```diff
@@ -119,40 +121,39 @@
 %         \meta{name2} |=| \meta{index2} |,| \ldots{}
 %      \}
```
muzimuzhi, line 122: From `texdoc l3bitset`, indents are missing in this and several more `syntax` envs.

muzimuzhi (commented):
Not a complete review.

New `bitset` variable type needs to be added to the full type list in `source3body.tex` (sec. 1.1 "Naming functions and variables" in `interface3.pdf`).

l3kernel/l3bitset.dtx
josephwright: That we can deal with post-PR I think

l3kernel/doc/source3body.tex
josephwright: Moved

josephwright:
> Not a complete review.
> 
> New `bitset` variable type needs to be added to the full type list in `source3body.tex` (sec. 1.1 "Naming functions and variables" in `interface3.pdf`).

Added

car222222:
Not quite what I suggested, but it does remove the error.

josephwright:
> Not quite what I suggested, but it does remove the error.

We already had string so I thought a bit of variety was good ...

car222222:
Sure, but my version any need for another name.
It also adds an extra detail at the end.

u-fischer:
l3bitset is as far as I can see only used in the pdfmanagement, so I can perpare for the move and add a test to load it only if not already in the kernel. But generally: should we add some empty wrapper for a transition time if we move experimental modules?

car222222:
Thanks:-)

car222222:
What @u-fischer wrote makes perfect sense to me.
There may be other ways to achieve this?

muzimuzhi:
> If you upload l3kernel with this PR to ctan I will get an error in the pdfmanagement as it uses \RequirePackage{l3bitset}. This could be avoided if there would be (temporary) a small dummy l3bitset.sty which simply issues a warning "do not used this package anymore, code has been moved into l3kernel".

Other packages to be moved from `l3experimental` to `l3kernel` would need similar treatments.

josephwright:
@u-fischer  Oh, that. In the past we did have a stub setup, but that was when we were consolidating widely-used modules. Here, I was imagining you'd go with
```latex
\cs_if_exist:NF \bitset_new:Nn
  { \RequirePackage { l3bitset } }
```
or similar.

u-fischer:
> wouldn't it be better to have `\ProvidesPackage` somewhere?

That wouldn't help. I mean during the "test-PR-phase" the package is still there (in the texmf-dist tree). 

>  Either way, how long would you want this to 'hang around'?

I would remove the package in the next update of the pdf management whenever that it is (currently there is nothing pending). And then l3kernel could remove a stub whenever you like.

josephwright:
@u-fischer  Latest change work for you? If so, it's likely a model for other moves from `l3experimental` if/when we make them (which I would of course like ...)

josephwright:
@FrankMittelbach  I'm trying to avoid having to have sources to DocStrip, etc., which 'come and go' ..

FrankMittelbach:
not the worst argument :-)

PR #1343, opened 2023-11-19
fix deprecated command entries

mbertucci47:
The entries for a few commands in l3deprecations.dtx have the wrong second argument to `\__kernel_patch_deprecation:nnNNpn`, repeating the deprecated command. This causes them to produce the wrong error when debugging, e.g.
```tex
! LaTeX Error: Use \bool_case_true:nT not \bool_case_true:nT
```
Full example:
```tex
\documentclass{article}

\ExplSyntaxOn
\debug_on:n { all }
\ExplSyntaxOff

\begin{document}

\ExplSyntaxOn
\bool_case_true:nT {}{}
\ExplSyntaxOff

\end{document}
```

### DIFF
--- a/l3kernel/l3deprecation.dtx
+++ b/l3kernel/l3deprecation.dtx
@@ -455,13 +455,13 @@
 %
 % \begin{macro}[EXP, noTF]{\bool_case_true:n}
 %    \begin{macrocode}
-\__kernel_patch_deprecation:nnNNpn { 2023-05-03 } { \bool_case_true:n }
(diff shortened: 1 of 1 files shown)


muzimuzhi:
May need an entry in https://github.com/latex3/latex3/blob/main/l3kernel/CHANGELOG.md.

@josephwright Want to test each deprecated function/variable/property (since the no removal policy)?

mbertucci47:
If this needs a changelog entry just let me know and I'll add it.

PR #1370, opened 2023-12-09
Standarise titles of modules and packages

muzimuzhi:
Started as spotting stray space in title of `l3bitset` in toc, 
image

I ended with standarising all titles of `l3backend` and `l3kernel` modules and `l3experimental`, `l3trial`, and `l3packages` packages.

Files in `l3leftovers` (after all they're leftovers) and `xpackages` (some files loaded non-`l3doc` class) are intact.

### DIFF
--- a/l3backend/l3backend-basics.dtx
+++ b/l3backend/l3backend-basics.dtx
@@ -31,7 +31,7 @@
 % \fi
 %
 % \title{^^A
-%   The \textsf{l3backend-basics} package\\ Backend basics^^A
+%   The \pkg{l3backend-basics} package\\ Backend basics^^A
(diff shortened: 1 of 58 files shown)


muzimuzhi:
Did CI step "Install TeX Live" fail so often before?

FrankMittelbach:
> Did CI step "Install TeX Live" fail so often before?

sometimes that happens and then usually the only thing to do is to wait. If the mirror that is picked up in inconsistent (because it is doing updates) you end up with this I guess. My feeling is that it is indeed more often, but right now it looks as if something generally is  wrong (eg bookmarks having an old version is odd).

PR #1349, opened 2023-11-25
Speed up string assignment functions like `\str_set:Nn`

muzimuzhi:
Full list is `\str_(g)set:Nn`, `\str_const:Nn`, `\str_(g)put_left:Nn`, and `\str_(g)put_right:Nn`.

Their current implementations use a structure `\tl_<func>:Ne <str var> { \tl_to_str:n {<token list>}}`, which is slow. Since (the current implementation of) `\tl_to_str:n` needs only one expansion step, the `e`-type variants `\tl_<func>:Ne` can be replaced with `o`-type  variants. In addition since string won't expand further, the `\__kernel_exp_not:w` used in corresponding base functions `\tl_<func>:Nn` is superfluous.

This PR hand-tune these base functions by using structures similar to their `\tl_<func>:Nn` siblings.

If needed, their variants can be hand-tuned as well, just like corresponding `tl` variants. `\str_set:Ne` is one of commonly used variants I think.

I'll add tests on local/global string assignments in a follow-up PR, including extending coverage for local/global token list assignments.

**Benchmark results** on setting `abc \c_empty_tl` (first lines for `\str_set:Nn` and second lines for `\str_set:Ne`)
```
[0] use \tl_set:Ne 
1.15e-6 seconds (4.88 ops)
1.2e-6 seconds (5.09 ops)
[1] use \tl_set:No 
7.76e-7 seconds (3.34 ops)
8.95e-7 seconds (3.81 ops)
[2] base hand-tuned
4.32e-7 seconds (1.83 ops)
6.55e-7 seconds (2.78 ops)
[3] e variant hand-tuned
4.43e-7 seconds (1.8 ops)
5.78e-7 seconds (2.37 ops)
```

<details><summary>Benchmark example</summary>
<p>

```tex
\documentclass{article}
\usepackage{l3benchmark}

\ExplSyntaxOn
\cs_new_protected:Npn \my_str_benchmarks:n #1
  {
    \iow_term:n { #1 }
    \benchmark:n { \str_set:Nn \l_tmpa_str { abc \c_empty_tl } }
    \benchmark:n { \str_set:Ne \l_tmpa_str { abc \c_empty_tl } }
  }

% trigger log line "\g__benchmark_1_int=\count<num>"
\benchmark_once:n { }
\iow_term:n { }

% copy the old implementation
\cs_gset_protected:Npn \str_set:Nn #1#2
  { \tl_set:Ne #1 { \tl_to_str:n { #2 } } }
[...]

PR #1381, opened 2023-12-12
Avoid trimming spaces twice on (raw) key names

muzimuzhi:
`\__keys_trim_spaces:n` is now applied only once, after the last `.` is found.

This is a refactor (found in my almost lost `git stash`) that simplifies `\__keys_property_find:n` a little bit.

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -1822,14 +1822,14 @@
 %   }
 %   Searching for a property means finding the last |.| in the input,
 %   and storing the text before and after it. Everything is turned into
-%   strings, so there is no problem using an \texttt{e}-type expansion. Since
-%   \cs{@@_trim_spaces:n} will turn its argument into a string anyway, this
-%   function uses \cs{cs_set_nopar:Npe} instead of \cs{tl_set:Ne} to gain some
-%   speed.
+%   strings at first, so there is no problem using an \texttt{e}-type expansion
+%   in aux functions. \cs{cs_set_nopar:Npe} instead of \cs{tl_set:Ne} or
+%   \cs{str_set:Ne} is used to gain some speed.
+%   Space trimming on key path is done only once, after the last |.| is found.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_property_find:n #1
   {
-    \cs_set_nopar:Npe \l_@@_property_str { \@@_trim_spaces:n { #1 } }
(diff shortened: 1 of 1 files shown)


l3kernel/l3keys.dtx
```diff
@@ -1822,14 +1822,14 @@
...
-%   speed.
+%   strings at first, so there is no problem using an \texttt{e}-type expansion
+%   in aux functions. \cs{cs_set_nopar:Npe} instead of \cs{tl_set:Ne} or
+%   \cs{str_set:Ne} is used to gain some speed.
+%   Space trimming on key path is done only once, after the last |.| is found.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_property_find:n #1
   {
-    \cs_set_nopar:Npe \l_@@_property_str { \@@_trim_spaces:n { #1 } }
+    \str_set:Nn \l_@@_property_str { #1 }
```
josephwright, line 1832: You've dropped the `\cs_set_nopar:Npe` here, so the text comment doesn't make sense - and do we not still want the extra speed?

josephwright (changes requested):

l3kernel/l3keys.dtx
muzimuzhi: The comment on `\cs_set_nopar:Npe` is actually for aux functions. I can make this clearer if needed.

After #1349, [`\str_set:Nn`][new str_set:Nn] is defined by
```
\cs_new_protected:Npn \str_set:Nn #1#2
  { \__kernel_tl_set:Ne #1 { \__kernel_tl_to_str:w {#2} } }
```
hence
```
\str_set:Nn \l_@@_property_str { #1 }
```
is only one expansion step slower than
```
\cs_set_nopar:Npe \l_@@_property_str { \tl_to_str:n { #1 } }
```

[new str_set:Nn]: https://github.com/latex3/latex3/blob/0ec3512fc5a80ba97752d1d852d6782fe140dbb3/l3kernel/l3str.dtx#L1001-L1002

josephwright (approved):

car222222:
I was unsure about this phrase:
_Everything is turned into strings at first_ 

Do you mean:
_Everything is first turned into strings_

Or something else?

muzimuzhi:
> Do you mean:
> _Everything is first turned into strings_

Yes, and this latter phrase is better.

car222222:
Good, I thought so:-).

PR #1387, opened 2023-12-15
l3keys: Speed up property_find

Skillmon:
This commit speeds up the `\@@_property_find:n` function of `l3keys` by roughly 20% - 30%.

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -1835,7 +1835,7 @@
       \s_@@_nil \@@_property_find_auxii:w
       . \s_@@_nil \@@_property_find_err:w
   }
-\cs_new_protected:Npn \@@_property_find_auxi:w #1 . #2 \s_@@_nil #3
+\cs_new:Npn \@@_property_find_auxi:w #1 . #2 \s_@@_nil #3
   {
     #3 #1 \s_@@_mark #2 \s_@@_nil #3
   }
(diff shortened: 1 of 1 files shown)


l3kernel/l3keys.dtx
```diff
@@ -1843,22 +1844,24 @@
...
+\cs_new:Npn \@@_property_find_auxiv:w
     #1 \s_@@_nil \@@_property_find_auxiii:w
     \s_@@_mark \s_@@_nil \@@_property_find_auxiv:w
   {
+    \if_false: {{{ \fi: }}}
     \cs_set_nopar:Npe \l_@@_property_str { . #1 }
-    \cs_set_nopar:Npe \l_keys_path_str
-      { \exp_after:wN \@@_trim_spaces:n \exp_after:wN { \l_keys_path_str } }
     \tl_set_eq:NN \l_keys_path_tl \l_keys_path_str
   }
```
muzimuzhi, line 1866: `\@@_property_find_auxiv:w` still contains assignments, hence is not expandable.

I'll open a PR to fix this and add `\begin{macro}[EXP]` for now expandable `\@@_property_find_auxi:w`.

muzimuzhi: Ah, it seems by using brace trick, `\@@_property_find_auxiv:w` has to be expandable.

Skillmon: Correct, and it's only ever called inside of a context where the first step was already `protected`, so nothing bad really. You could (in theory) define it `protected` and then trick the system to expand it anyways (e.g., `\exp_after:wN \prg_do_nothing: \@@_property_find_auxiv:w`) but I don't see any gain there (it would however slow this solution down...).

FrankMittelbach: I agree that unnecessary protection that is then worked around for something that is only used as an internal function inside a protected function (and never exprected to be used elsewhere) is in my opinion the wrong approach. If you want, you should explicitly say that "it is not protected because..." but  the highest goal for such internal stuff is really to be fast

PR #1388, opened 2023-12-16
Update mark-ups and comment for `\__keys_property_find:n` and aux functions

muzimuzhi:


l3kernel/l3keys.dtx
```diff
@@ -1861,6 +1866,10 @@
     \s_@@_mark \s_@@_nil \@@_property_find_auxiv:w
   {
     \if_false: {{{ \fi: }}}
+    \@@_property_find_auxv:w { #1 }
+  }
+\cs_new_protected:Npn \@@_property_find_auxv:w #1
```
Skillmon, line 1871: I'd veto this. It doesn't add any clarity to the code, on the contrary, and adds an additional internal unnecessarily. Even if you split out the assignments from `auxiv` following coding style it still would have to be `protected` as it calls an unexpandable function. The `protected`/unprotected split is necessary for high-level things, but this is almost as low level as it gets, is nested inside a `protected` function, never used outside that `protected` scope, and, arguably, kernel code knows what it does and should be optimised for speed (within sane bounds).

Also, it should really be `\@@_property_find_auxv:n` here... (if it were in fact added, but see above).

Skillmon: Oh, and if it is decided to use this indirection, it should (for performance reasons) be:

```latex
\cs_new:Npn \@@_property_find_auxiv:w
  {
    \if_false: {{{ \fi: }}}
    \@@_property_find_auxv:w
  }
\cs_new_protected:Npn \@@_property_find_auxv:w
    #1 \s_@@_nil \@@_property_find_auxiii:w
    \s_@@_mark \s_@@_nil \@@_property_find_auxiv:w
  {
    \cs_set_nopar:Npe \l_@@_property_str { . #1 }
    \tl_set_eq:NN \l_keys_path_tl \l_keys_path_str
  }
```

l3kernel/l3keys.dtx
```diff
@@ -1813,12 +1813,17 @@
 % \end{macro}
 %
 % \begin{macro}{\@@_property_find:n}
-% \begin{macro}
+% \begin{macro}[EXP]
 %   {
 %     \@@_property_find_auxi:w   ,
+%     \@@_property_find_auxiii:w
```
Skillmon, line 1819: `\@@_property_find_auxiv:w` is missing from this list (if we want to be this precise here).

l3kernel/l3keys.dtx
muzimuzhi: Introduction of `\@@_property_find_auxv:w` is reverted now.

Skillmon (approved):

PR #1392, opened 2023-12-18
Document default l3doc class options

muzimuzhi:
For the typo,
- before
  image
  > `kernel`
  > Determines whether `l3doc` treats `\__kernel_` commands and `\(cgl)__kernel_` variables as allowable in code.
- after (initial proposal)
  image
- after (latest)
  > `kernel`
  > Determines whether `l3doc` treats internal functions and variables belong to `kernel` module as allowable in code, for instance `\__kernel_tl_to_str:w`, `\c__kernel_expl_date_tl`, and `\l__kernel_expl_bool`.

### DIFF
--- a/l3kernel/l3doc.dtx
+++ b/l3kernel/l3doc.dtx
@@ -272,6 +272,12 @@ and all files in that bundle must be distributed together.
 % class option. To disable hyphenation of control sequences entirely, use
 % \texttt{cs-break = false}.
 %
+% By default, class options
+% \begin{verbatim}
+%   full , kernel , check = false , checktest = false , lm-default
+% \end{verbatim}
+% are set.
+%
 % \subsection{Partitioning documentation and implementation}
 %
 % \pkg{doc} uses the \cs{OnlyDocumentation}/\cs{AlsoImplementation}
--- a/l3kernel/l3doc.dtx
+++ b/l3kernel/l3doc.dtx
@@ -237,8 +237,10 @@ and all files in that bundle must be distributed together.
 % encoding (the standard setting) or leaves the font setup unchanged.
 %
 % \DescribeOption{kernel}
-% Determines whether \pkg{l3doc} treats |\__kernel_| commands and
-% |\(c|g|l)__kernel_| variables as allowable in code. In general,
+% Determines whether \pkg{l3doc} treats internal functions and variables
+% belonging to |kernel| module as allowable in code, for instance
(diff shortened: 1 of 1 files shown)


l3kernel/l3doc.dtx
```diff
@@ -238,7 +238,7 @@ and all files in that bundle must be distributed together.
 %
 % \DescribeOption{kernel}
 % Determines whether \pkg{l3doc} treats |\__kernel_| commands and
-% |\(c|g|l)__kernel_| variables as allowable in code. In general,
+% "\(c|g|l)__kernel_" variables as allowable in code. In general,
```
FrankMittelbach, line 241: before it was simply wrong, now it is just odd, because everything is in typewriter. I guess most people will understand though, but perhaps one should do the extra step and use normal text font on the parentheses or just avoid this altogether and write them separately. Dunno

FrankMittelbach (approved):
just a thought

l3kernel/l3doc.dtx
muzimuzhi: Maybe just rewrite this to "treats functions and variables under `kernel` module".

FrankMittelbach: or something like this, yes. One may might still want to mention `|__kernel_|` in that sentence as that is just highlighting what "kernel module" means as this is correct for all of them and might make it clearer what is meant.

muzimuzhi: Oh, it should be "internal (functions and variables)".

l3kernel/l3doc.dtx
```diff
@@ -237,8 +237,10 @@ and all files in that bundle must be distributed together.
 % encoding (the standard setting) or leaves the font setup unchanged.
 %
 % \DescribeOption{kernel}
-% Determines whether \pkg{l3doc} treats functions and variables belong to
-% |kernel| module as allowable in code. In general,
+% Determines whether \pkg{l3doc} treats internal functions and variables belong
+% to |kernel| module as allowable in code, for instance
+% \cs{__kernel_tl_to_str:w}, \cs{c__kernel_expl_date_tl}, and
```
FrankMittelbach, line 242: my gut feeling is that it should say "or" not "and" in that case ... but what do I know :-)

davidcarlisle: I think `and` is OK but `belong to |kernel|` should say `belonging to the |kernel|`

muzimuzhi: Changed to `belonging to`, thanks @davidcarlisle.

PR #1394, opened 2023-12-19
l3regex performance tweaks

Skillmon:
This PR does a bunch of minor performance tweaks to `l3regex`. This is just a droplet on a hot stone, of course, but the changes aren't too drastic, imho, and should at least somewhat improve `l3regex`'s performance (though most likely not really measurable compared to the rest of the algorithms).


l3kernel/l3regex.dtx
```diff
@@ -1135,10 +1135,10 @@
 %   Empty a \tn{toks} or set it to a value, given its number.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_toks_clear:N #1
-  { \@@_toks_set:Nn #1 { } }
+  { \tex_toks:D #1 = { } }
```
josephwright, line 1138: How does this save time? `\@@_toks_set:Nn` _is_ `\tex_toks:D`

l3kernel/l3regex.dtx
```diff
@@ -1135,10 +1135,10 @@
...
 \cs_new_eq:NN \@@_toks_set:Nn \tex_toks:D
 \cs_new_protected:Npn \@@_toks_set:No #1
-  { \tex_toks:D #1 \exp_after:wN }
+  { \tex_toks:D #1 = \exp_after:wN }
```
josephwright, line 1141: We should likely tidy up the `:D` use one way or the other!

l3kernel/l3regex.dtx
```diff
@@ -1167,20 +1167,24 @@
 %   \cs{@@_toks_put_right:Ne} is provided because it is more
 %   efficient than \texttt{e}-expanding with \cs{exp_not:n}.
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_toks_put_left:Ne #1#2
+\cs_if_exist:NTF \tex_etokspre:D
```
josephwright, line 1170: This part I like - could we perhaps have a separate PR that only uses these primitives where available? I think that's easier to justify than using more `:D` functions across the board.

josephwright (changes requested):
Bit of a mixed bag: I'd like to know what @FrankMittelbach thinks

l3kernel/l3regex.dtx
FrankMittelbach: one less expansion and no failed search for = . In case of `_clear`perhaps not so essential as with `_set`but ...

l3kernel/l3regex.dtx
FrankMittelbach: This is one of the core data structures (even if we don't use it that much in expl3) and as such I think it warrants a fast as possible implementation, so I'm for the `:D` functions including the use of `=` in the code. If you are worried I suggest to add a note at the start of the code section that the implementation is  coded for speed as potentially the functions are used very often.

l3kernel/l3regex.dtx
muzimuzhi: Isn't the gain only from `=`, hence `\@@_toks_set:Nn #1 = { }` and `\tex_toks:D #1 = { }` are the same?
https://github.com/latex3/latex3/blob/e2f3be7d6dfe1395a80083cd68f5437cc66738dc/l3kernel/l3regex.dtx#L1136

FrankMittelbach: true as it is set equal, but I would still use :D throughout as it is more consistent and  while `\@@_toks_set:Nn #1 { }` is correct from the syntax point of view  `\@@_toks_set:Nn #1 = { }` wouldn't be.  What is the `=` doing here? picked up by 'n' or extra or ...

josephwright:
OK, @FrankMittelbach is happy so I am too. Can you addressed the conflicts then I can merge.

l3kernel/l3regex.dtx
Skillmon: Exactly @FrankMittelbach's reasoning was responsible to use `=`, which was the reason to use `:D` instead of `:Nn`.

Issue #1402, opened 2024-01-01
Passing \c_novalue_tl to a command, and testing for it


ysalmon:
I defined a versatile internal function, here simplified to :
```
\cs_new:Nn \example:nnnn {
  \tl_if_novalue:nTF {#1} {T} {F}
  % etc.
}
```

I have a more abstract layer of functions, of which some have to supply NoValue to ``\example`` :
```
\cs_new:Nn \second: {
  \example:nnnn {\c_novalue_tl} {things} {that} {depend}
}
```

But this does not work, as I get a F instead of a T.

More simply, ``\tl_if_novalue:nTF{\c_novalue_tl}{T}{F}`` gives an F.

I could not find a way to effectively pass the NoValue to a command, as if an optional argument in a user command had not been supplied.

The use case is typesetting an exercise sheet : questions can have a title, so the user command has a ``o`` argument ; tests cannot, so I do not want to have a ``o`` argument in the user command ; but at some point, this boils down to a generic function that typesets a Q or a T in the margin, and a title if there is one. I need my user command for tests to always give a NoValue as title to the internal function.

muzimuzhi:
> More simply, `\tl_if_novalue:nTF{\c_novalue_tl}{T}{F}` gives an F.

You need `\exp_args:No \tl_if_novalue:nTF {\c_novalue_tl} {T} {F}`. That is, feed the one-step expansion result of `\c_novalue_tl` to `\tl_if_novalue:nTF`.

ysalmon:
Thank you, I got to what I wanted with like ``\exp_args:No\tl_if_novalue:nF {#1} {\refstepcounter{#1}}`` and seems to work in all my cases, because expanding ``#1`` here is not a problem.

I am not closing this however, since the behaviour of expl3 seems counter-intuitive at least, and is not documented afaik.

Skillmon:
@ysalmon the wording might be misleading a bit, but it is documented for `\tl_if_novalue:nTF`:

> Tests if the ⟨*`token list`*⟩ is exactly equal to the special `\c_novalue_tl` marker.

doesn't mean that the ⟨*`token list`*⟩ contains the `\c_novalue_tl` marker, but is equal in the sense that if you do `\tl_set:Nn \l_tmpa_tl {#1}` the result then tests true with `\tl_if_eq:NNTF` to `\c_novalue_tl`. But I agree, the wording might be improved here, though I'm not sure how exactly.

ysalmon:
[...]

PR #1405, opened 2024-01-02
Clarify behavior of `\tl_if_novalue:n(TF)`

muzimuzhi:
See #1402.

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -330,8 +330,12 @@
 %     \cs{tl_if_novalue_p:n} \Arg{token list}
 %     \cs{tl_if_novalue:nTF} \Arg{token list} \Arg{true code} \Arg{false code}
 %   \end{syntax}
-%   Tests if the \meta{token list} is exactly equal to the special
-%   \cs{c_novalue_tl} marker. This function is intended to allow construction
+%   Tests if the \meta{token list} is exactly equal to the content of special
+%   \cs{c_novalue_tl} marker, hence it is functionally the same as
+%   \begin{quote}
+%     \cs{tl_if_eq:NnTF} \cs{c_novalue_tl} \Arg{token list} \Arg{true code} \Arg{false code}
+%   \end{quote}
+%   but expandable. This function is intended to allow construction
 %   of flexible document interface structures in which missing optional
 %   arguments are detected.
 % \end{function}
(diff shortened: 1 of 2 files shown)


l3kernel/l3tl.dtx
```diff
@@ -330,8 +330,12 @@
...
 %     \cs{tl_if_novalue:nTF} \Arg{token list} \Arg{true code} \Arg{false code}
 %   \end{syntax}
-%   Tests if the \meta{token list} is exactly equal to the special
-%   \cs{c_novalue_tl} marker. This function is intended to allow construction
+%   Tests if the \meta{token list} is exactly equal to the content of special
+%   \cs{c_novalue_tl} marker, hence it is functionally the same as
+%   \begin{quote}
+%     \cs{tl_if_eq:NnTF} \cs{c_novalue_tl} \Arg{token list} \Arg{true code} \Arg{false code}
+%   \end{quote}
+%   but expandable. This function is intended to allow construction
```
FrankMittelbach, line 338: sorry, but I find my explanation easier to understand. Yours is certainly a correct description but I think it is hard from your explanation to deduce that \tl_if_novalue:nTF \c_novalue_tl returns false.

But yes my explanation could most likely be improved in the wording.

car222222: The longer version here makes more sense to me (with minor enhancements:-):

`Tests whether the \meta{token list} is exactly equal to the content of the special  \cs{c_novalue_tl} marker token list.`

Not sure about the extra explanations, maybe both forms are needed?

muzimuzhi: @FrankMittelbach I replaced my `\tl_if_eq:NnTF \c_novalue_tl {<token list>}` with something adapted from your explanation in https://github.com/latex3/latex3/issues/1402#issuecomment-1873461713. The fact that `\c_novalue_tl` is not a quark is told can be seen in `\c_novalue_tl`'s doc, so I dropped that mention to quark.

@car222222 "Tests if ... " is used in documentation of all other `\tl_if_...` conditionals, so I decide to stick to this phrase. `\c_novalue_tl`'s name already tells it's a token list variable, thus I hesitate to add an extra "token list" tail.

"both in respect of character codes and category codes" is added, to be aligned with docs of `\tl_if_equal:(NN|Nn|nn)(TF)` functions.

car222222: > Tests if ... " is used in documentation of all 

Sure! (but I would change all of them:-)

car222222 (approved):
Latest version looks very good to me. Thanks!

FrankMittelbach (approved):

PR #1408, opened 2024-01-02
Use `.default:n` value for unknown key

### LINKED ISSUE #67: Default value of unknown key
I would expect

```
\keys_define:nn { my-keys }
  {
    unknown .default:n = x ,
    unknown .code:n = \tl_show:n {#1}
  }
\keys_set:nn { my-keys } { some-key }
```

to show the default value `x`. Instead, the given default value is ignored, and I see an empty token list.

Skillmon:
This PR changes the unknown-handler invocation to support a default value for unknown keys. This would effectively close #67.

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -3248,7 +3248,19 @@
       {
         \cs_if_exist:cTF
           { \c_@@_code_root_str \l_@@_module_str / unknown }
-          { \@@_execute:no { \l_@@_module_str / unknown } \l_keys_value_tl }
+          {
+            \bool_if:NT \l_@@_no_value_bool
+              {
+                \cs_if_exist:cT
+                  { \c_@@_default_root_str \l_@@_module_str / unknown }
+                  {
+                    \tl_set_eq:Nc
+                      \l_keys_value_tl
+                      { \c_@@_default_root_str \l_@@_module_str / unknown }
+                  }
+              }
+            \@@_execute:no { \l_@@_module_str / unknown } \l_keys_value_tl
+          }
           {
             \msg_error:nnee { keys } { unknown }
               \l_keys_path_str \l_@@_module_str
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -775,12 +775,14 @@
 % If a key has not previously been defined (is unknown), \cs{keys_set:nn}
 % looks for a special \texttt{unknown} key for the same module, and if this is
 % not defined raises an error indicating that the key name was unknown. This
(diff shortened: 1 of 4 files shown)


Skillmon:
Not sure whether documentation needs changes. A changelog entry is still missing.

muzimuzhi:
Maybe add a test to `m3keys001.lvt`? Currently it's the only test file that contains a test for `unknown` key:
https://github.com/latex3/latex3/blob/d1a7b2d83c7bb1816335358f171ba2b2fdbfcb43/l3kernel/testfiles/m3keys001.lvt#L270

l3kernel/l3keys.dtx
```diff
@@ -775,12 +775,14 @@
...
-% mechanism can be used for example to issue custom error texts.
+% mechanism can be used for example to issue custom error texts. The
+% \texttt{unknown} key also supports the \texttt{.default:n} property.
 % \begin{verbatim}
 %   \keys_define:nn { mymodule }
 %     {
 %       unknown .code:n =
-%         You~tried~to~set~key~'\l_keys_key_str'~to~'#1'.
+%         You~tried~to~set~key~'\l_keys_key_str'~to~'#1'. ,
+%       unknown .default:V = \c_novalue_tl
```
josephwright, line 785: Not keen on `\c_novalue_tl` here: that's really an `ltcmd`-helper so I'd rather see something different if we are showing this off.

josephwright (approved):
The one comment can have a little discussion, but otherwise looks good.

l3kernel/l3keys.dtx
Skillmon: @josephwright I think this is exactly a use case for `\c_novalue_tl`, from the current docs of `\tl_if_novalue:nTF` 

> This function is intended to allow construction of flexible document interface structures in which missing optional arguments are detected.

and

> A marker for the absence of an argument.

`key=value` input is a flexible document interface, and the marker is used for the absence of an argument.

FrankMittelbach: @josephwright I think I agree with @Skillmon here. If you think of, say `\foo` being defined with the sig `om` then `\foo{xyz}`would have `#1` being `--NoValue--`. If we extend `\foo`, say, for tagging, and the option argument would become a key value argument with the "default key" for the optional arg being `text` then `\foo{xyz}` would be (or should be) `text= --NoValue--` as would `\foo[bar=abc]{xyz}`, while `\foo[bar=abc,text=123]{xyz}` or \foo[123]{xyz}`would return `text=123`.

So using V of `\c_novalue_tl` makes sense to me in this case.

PR #1412, opened 2024-01-03
Faster align_safe_end

Skillmon:
I'm not sure why I didn't think of that when I sped up `\group_align_safe_begin:`, but this here speeds up `\group_align_safe_end:` in the same way.

### DIFF
--- a/l3kernel/l3prg.dtx
+++ b/l3kernel/l3prg.dtx
@@ -1789,9 +1789,10 @@
 \tex_catcode:D `\^^@ = 2 \exp_stop_f:
 \cs_new:Npn \group_align_safe_begin:
   { \exp:w \if_false: { \fi: `^^@ \exp_stop_f: }
-\group_end:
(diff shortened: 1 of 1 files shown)


Skillmon:
The speed increases by roughly 3% to 5%.

blefloch (approved):
The speed is the same as before, within error bars, but this is more symmetrical with the code of `\group_align_safe_begin:` (which was a significant speedup) so let's go with this change.

Skillmon:
@blefloch for me it was pretty consistently faster in pdfTeX (though it's not much, agreed). Also, this'll require a change in `unravel` (as did the last change, just a reminder for you).

blefloch:
Thanks for the heads up about `unravel`.  I had anticipated the problem there some time ago, where I redefine both `\group_align_safe_begin:` and `\group_align_safe_end:`, so it's all good.

PR #1415, opened 2024-01-10
Move \seq_set_filter:NNn to stable, etc.

josephwright:
As discussed on the team list


FrankMittelbach (approved):

blefloch:
- Somewhere in the documentation of `\seq_set_filter:NNn` we could give the example of «For instance, if you want to eliminate all terms in a sequence until a given item is found, you can use `\flag_clear:n { tmpa }` then use `\seq_set_filter:NNn` with the boolean expression `\str_if_eq:nnT {#1} { a given item } { \flag_raise:n { tmpa } } \flag_if_raised_p:n { tmpa }`».  Actually, I guess we should document a bunch of recipes somewhere; I think this was raised previously when discussing sequence intersection/union.

- By chance I came across a mention of the LaTeX-L mailing list in `expl3.dtx`.  Is that list still active?

- Since we will keep the old functions indefinitely, they will show up in old code, so the old names should be documented somewhere slightly better than the changelog, maybe just by setting `\cs_set_protected:Npn \keys_set_filter:nnn { \keys_set_exclude_groups:nnn }` instead of with `\cs_set_eq:NN`.

- The argument of the macro environment in `l3deprecation.dtx` should be `\keys_set_filter:nnn` etc, not the new name.

- Should the old functions be added to the right place in l3debug so that when the deprecation option is selected they will be made into errors?  I'm not sure about that one, maybe it needs a separate option.

josephwright:
>     * Since we will keep the old functions indefinitely, they will show up in old code, so the old names should be documented somewhere slightly better than the changelog, maybe just by setting `\cs_set_protected:Npn \keys_set_filter:nnn { \keys_set_exclude_groups:nnn }` instead of with `\cs_set_eq:NN`.

Changed 

>     * The argument of the macro environment in `l3deprecation.dtx` should be `\keys_set_filter:nnn` etc, not the new name.

Corrected

>     * Should the old functions be added to the right place in l3debug so that when the deprecation option is selected they will be made into errors?  I'm not sure about that one, maybe it needs a separate option.

Adjusted: I'd forgotten that the lines I copied as a model were non-standard :)

josephwright:
>     * By chance I came across a mention of the LaTeX-L mailing list in `expl3.dtx`.  Is that list still active?

I've been picking those off: the list still works but I don't think it can be recommended anymore

josephwright:
> * Somewhere in the documentation of `\seq_set_filter:NNn` we could give the example of «For instance, if you want to eliminate all terms in a sequence until a given item is found, you can use `\flag_clear:n { tmpa }` then use `\seq_set_filter:NNn` with the boolean expression `\str_if_eq:nnT {#1} { a given item } { \flag_raise:n { tmpa } } \flag_if_raised_p:n { tmpa }`».  Actually, I guess we should document a bunch of recipes somewhere; I think this was raised previously when discussing sequence intersection/union.

I think this is likely out-of-scope for the PR :)

PR #1419, opened 2024-01-12
Implement N-type flag variables

blefloch:
Flags are the only variable type that can be modified in an expansion-only context (without using LuaTeX).  They are currently named by an `n`-type string like `fp_overflow` and the present changes allow more LaTeX3 names such as `\l_fp_overflow_flag`.  (In fact, the implementation is such that both the `n`-type flag `fp_overflow` and the `N`-type flag `\l_fp_overflow_flag` have the same value since they use the same underlying csnames.)

There should be a slight speed up from having fewer tokens to move around, but the main motivation is to make flags more similar to other variables.  The goal here is to eventually deprecate `n`-type flag variables.  It is not currently possible to deprecate them since flags are used quite a lot despite the numerous warnings in the documentation that they are almost always a bad idea.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -9,6 +9,7 @@ this project uses date-based 'snapshot' version identifiers.
 
 ### Added
 - `\keys_set_exclude_groups:nnn(nN)` to replace `\keys_set_filter:nnn(nN)`
+- Flags with N-type names, like other variable types
 
 ### Changed
 - Set `l3doc` option `kernel` off as-standard (issue \#1403)
--- a/l3kernel/doc/source3body.tex
+++ b/l3kernel/doc/source3body.tex
@@ -321,7 +321,7 @@ following variable types:
   \item[\texttt{box}] Box register.
   \item[\texttt{coffin}] A \enquote{box with handles} --- a higher-level
     data type for carrying out \texttt{box} alignment operations.
-  \item[\texttt{flag}] Integer that can be incremented expandably.
+  \item[\texttt{flag}] Non-negative integer that can be incremented expandably.
   \item[\texttt{fparray}] Fixed-size array of floating point values.
   \item[\texttt{intarray}] Fixed-size array of integers.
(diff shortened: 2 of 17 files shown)


josephwright (approved):
Code looks good to me - my only query is do we *need* this as like color, messages, etc., scope is 'fixed'. I can see the difference, though, in that here we have a normal namespace.

Probably `flag` should get added to the list of variable types `expl3.dtx`.

blefloch:
Sorry, I don't understand what you mean by "the scope is fixed".  It's a good question whether we should stay with n-type flags or switch over.  Quite many types exist, some with N-type variables, others n-type, and I don't yet understand the logic.

In N-type variables with normal names, we have
- **Numerical data types** (bool, fp, int, dim, muskip, skip).
- **Lists and arrays** (bitset, clist, fparray, intarray, prop, seq).
- **Token storage** (str, token, tl).
- **Boxes** (box, coffin).
- **Other** (cctab, ior, iow, regex).

In the commit message of e1ad2c10797d75eab56a46a93a1eac1caa717860 I attempt to classify types that do not have such conventional names.
- **Constant markers** (quark, scan mark).
- **Functions** mostly constant (cs, conditional, \prg_new_map_functions).
- **Nested names** mostly-constant (msg, keys).
- **Used in expressions** (fp variables, fp functions, color, color_model).
- **Other** (pdf_object, flag).

I feel like these choices of N-type variables vs other kinds of names mostly make sense.  The main distinction is whether a given type is typically variable or constant.  This fails to explain cctab and regex (typically constant), fp variables, flags and colors (variable), and possibly pdf objects.  For fp variables and colors we don't have a choice since some of them like `pi` and `red` have to be n-type.

That leaves cctab, regex, flag, pdf objects.  I lean towards making all of these N-type (cctab and regex already are).

josephwright:
I'm happy but I'll ask the rest of the team at the next meeting

blefloch:
Thanks. Please include the case of pdf objects in the question.

Either way, it's not a big deal to reject this pull request.

PR #1421, opened 2024-01-15
Faster cs_if_exist

Skillmon:
This PR improves the performance of `\cs_if_exist:NTF`, `\cs_if_exist:cTF`, `\cs_if_free:NTF`, `\cs_if_free:cTF`, `\cs_if_exist_use:NTF`, and `\cs_if_exist_use:cTF`. The performance increases are mostly due to the `\prg_set_conditional:Npnn` optimisations being used by restructuring the checks (unfortunately at the cost of a few internals), but in the `c`-variants `\lastnamedcs` is also used if available.

Speed increases:

- `\cs_if_exist:NTF` 10 %
- `\cs_if_exist:cTF` 17 % (pdfTeX), 50 % (LuaTeX)
- `\cs_if_free:NTF` 11 %
- `\cs_if_free:cTF` 19 % (pdfTeX), 48 % (LuaTeX)
- `\cs_if_exist_use:NTF` marginally (only for `N`, `NT`, and `NF`, not `NTF` which is unchanged)
- `\cs_if_exist_use:cTF` 38 % (pdfTeX), 60 % (LuaTeX)


josephwright (commented):
I think something in the changelog would be useful. I'll leave it to Bruno to comment on the performance - I see the logic of using `\lastnamedcs` if available, certainly.

Skillmon:
At the cost of another internal one could further improve `\cs_if_exist:NTF` (which currently only uses a second -- technically first -- `\else:` to avoid that). Also `\__cs_if_exist_use_aux:Nnn` could profit from `\__prg_TF_true:w`, which I didn't use to keep the code clean (and if not used directly also to avoid a too messy setup for it involving a bit of expansion control and a technically wrong macro name).

There might be further small performance improvements possible (this doesn't claim to be optimal) but this looked like a good balance between complicated (and/or manual) setup and performance.

PR #1422, opened 2024-01-17
Update l3regex.dtx , Maybe a typo

rockyzhz:
To be consistent with the others.

Package: expl3 2024-01-04
TeX Live 2023
pdfTeX

### DIFF
--- a/l3kernel/l3regex.dtx
+++ b/l3kernel/l3regex.dtx
@@ -228,7 +228,7 @@
 %     Matches any of the specified tokens.
 %   \item[{[\char`\^\ldots{}]}] Negative character class.
 %     Matches any token other than the specified characters.
-%   \item[{x-y}] Within a character class, this denotes a range (can be
+%   \item[{[x-y]}] Within a character class, this denotes a range (can be
 %     used with escaped characters).
 %   \item[{[:\meta{name}:]}] Within a character class (one more set of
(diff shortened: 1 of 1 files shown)


l3kernel/l3regex.dtx
```diff
@@ -228,7 +228,7 @@
 %     Matches any of the specified tokens.
 %   \item[{[\char`\^\ldots{}]}] Negative character class.
 %     Matches any token other than the specified characters.
-%   \item[{x-y}] Within a character class, this denotes a range (can be
+%   \item[{[x-y]}] Within a character class, this denotes a range (can be
 %     used with escaped characters).
 %   \item[{[:\meta{name}:]}] Within a character class (one more set of
 %     brackets), this denotes the \textsc{posix} character class
```
muzimuzhi, line 234: As it's already "within a character class", I think the original `x-y` without brackets is alright.

For the POSIX character class, l3regex does require an extra pair of brackets `[:<name>:]` even within a character class.

See an example:
```tex
\regex_show:n {[0-9]}
\regex_show:n {[[:digit:]]}
```
write
```
> Compiled regex {[0-9]}:
+-branch
  range [48 (0),57 (9)].

> Compiled regex {[[:digit:]]}:
+-branch
  range [48 (0),57 (9)].
<recently read> }
```
to log.

Perhaps the l3regex doc needs more examples.

PR #1423, opened 2024-01-17
Deprecate global msg setting

josephwright:


josephwright:
We should have a consistent scope for messages - based on the usual 'local better than global' view, I've gone all-local except we always initialise globally. For almost all users, no change.

l3kernel/doc/l3obsolete.txt
```diff
@@ -53,6 +53,8 @@ Function                            Date deprecated
 \l_keys_path_tl                          2020-02-08
 \l_text_accents_tl                       2023-02-07
 \l_text_letterlike_tl                    2023-02-07
+\msg_gset:nnn                            2024-01-11
+\msg_gset:nnnn                           2024-01-11
```
muzimuzhi, line 57: Do the deprecation date here (and below in `l3deprecation.dtx`) need an update?

PR #1440, opened 2024-02-11
\if:w documentation

jlaurens:
Problems solved:
- `<token_1>` and `<token_2>` notations used only once
- `<true code>` not well defined
- typo
- legal construct not covered:
```
\def\x{xYES\else: NO\fi:}
\def\y{yNO\else: YES\fi:}
\if:w x\x
\if:w x\y
```

### DIFF
--- a/l3kernel/l3basics.dtx
+++ b/l3kernel/l3basics.dtx
@@ -1251,14 +1251,13 @@
 %
 % \begin{function}[EXP]{\if:w, \if_charcode:w, \if_catcode:w}
 %   \begin{syntax}
-%     "\if:w" <token(s)> <true code> "\else:" <false code> "\fi:" \\
-%     "\if_catcode:w" <token(s_> <true code> "\else:" <false code> "\fi:"
+%     "\if:w" <token_1> <token_2> <true code> "\else:" <false code> "\fi:" \\
+%     "\if_catcode:w" <token_1> <token_2> <true code> "\else:" <false code> "\fi:"
 %   \end{syntax}
-%   These conditionals expand \meta{token(s)} until two
+%   These conditionals expand their following tokens until two
 %   unexpandable tokens \meta{token_1} and \meta{tokens_2} are left;
-%   any further tokens become part of the \meta{true code}.
-%   If you wish to prevent this expansion,
-%   prefix the token in question with "\exp_not:N". "\if_catcode:w"
+%   any further tokens up to the next "\else" are the \meta{true code}.
+%   With "\exp_not:N", you can prevent the expansion of a token. "\if_catcode:w"
 %   tests if the category codes of the two tokens are the same whereas
 %   "\if:w" tests if the character codes are
 %   identical. "\if_charcode:w" is an alternative name for "\if:w".
--- a/l3kernel/l3basics.dtx
+++ b/l3kernel/l3basics.dtx
@@ -1251,18 +1251,23 @@
 %
 % \begin{function}[EXP]{\if:w, \if_charcode:w, \if_catcode:w}
 %   \begin{syntax}
-%     "\if:w" <token_1> <token_2> <true code> "\else:" <false code> "\fi:" \\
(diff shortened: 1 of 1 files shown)


l3kernel/l3basics.dtx
```diff
@@ -1251,14 +1251,13 @@
...
 % \begin{function}[EXP]{\if:w, \if_charcode:w, \if_catcode:w}
 %   \begin{syntax}
-%     "\if:w" <token(s)> <true code> "\else:" <false code> "\fi:" \\
-%     "\if_catcode:w" <token(s_> <true code> "\else:" <false code> "\fi:"
+%     "\if:w" <token_1> <token_2> <true code> "\else:" <false code> "\fi:" \\
+%     "\if_catcode:w" <token_1> <token_2> <true code> "\else:" <false code> "\fi:"
 %   \end{syntax}
-%   These conditionals expand \meta{token(s)} until two
+%   These conditionals expand their following tokens until two
 %   unexpandable tokens \meta{token_1} and \meta{tokens_2} are left;
```
car222222, line 1258: Should be:   token_2

Better:
. . . are found;

l3kernel/l3basics.dtx
```diff
@@ -1251,14 +1251,13 @@
...
-%   any further tokens become part of the \meta{true code}.
-%   If you wish to prevent this expansion,
-%   prefix the token in question with "\exp_not:N". "\if_catcode:w"
+%   any further tokens up to the next "\else" are the \meta{true code}.
+%   With "\exp_not:N", you can prevent the expansion of a token. "\if_catcode:w"
```
car222222, line 1260: Maybe better with a new paragraph after:  . . . token.

car222222 (changes requested):
Two extra adjustments noticed.

muzimuzhi:
Personally, since `\if:w`, `\if_charcode:w`, and `\if_catcode:w` are aliases of primitives, I think their detailed documentation can be omitted from `interface3.pdf`.

jlaurens:
I initially asked for a change in the documentation because it was not consistent with the description of the texbook. So removing the doc is an option
Cons:
- the file is less self contained
- we must have access to offline reliable original documentation to compensate
Pros
- the TeXbook.pdf is available online (there might be legal restrictions)
- TeX by topic is in TeXLive
- There are other expl APIs for the same purpose
- documenting only expl APIs promotes their usage
- send the message 'you can avoid complicated TeX primitives'

Skillmon:
>     * the TeXbook.pdf is available online (there might be legal restrictions)

There are, none of the PDFs you find on the internet are likely to be legally distributed, you might want to report those sources if you find them.

That said, the TeXbook sources are on CTAN and free and you might read them (but you're not allowed to build a PDF from them).

josephwright:
> Personally, since `\if:w`, `\if_charcode:w`, and `\if_catcode:w` are aliases of primitives, I think their detailed documentation can be omitted from `interface3.pdf`.

No - as they are given documented `expl3` names. Anything documented as part of the `expl3` API need documentation which covers the uses they can be put to. That doesn't necessarily mean all of the coverage as-in The TeXbook. For example, `\exp_not:n` is `\unexpanded` renamed, but it is documented as taking an `n`-type argument: that means that using say `\exp_not:n \exp_after:wN { ... }` is not supported (and indeed the reason we have `\__kernel_exp_not:w`).

So _appropriate_tightening of the docs is fine.

josephwright:
@jlaurens I'm confused here: you've asked to undone the tightening up I made at your request? For example,`\if:w` needs only to be followed by one token - what's important for the test is what it expands to. So `<token_1> <token_2>` is misleading.

jlaurens:
Changes: 
- left→found
- \if_charcode:w == \if:w put first to explain why there is no example of  \if_charcode:w usage
- \else: optional when in front of \fi:
- nesting is allowed
- <token_1> and <token_2> used twice

PR #1441, opened 2024-02-11
\ExplSyntax... documentation

jlaurens:
Problems solved:
- Incorrect documentation of `\ExplSyntaxOff`
- cctab cannot be used to implement `\ExplSyntaxOn` as it is

### DIFF
--- a/l3kernel/l3bootstrap.dtx
+++ b/l3kernel/l3bootstrap.dtx
@@ -69,18 +69,26 @@
 %   \begin{syntax}
 %     \cs{ExplSyntaxOn} \meta{code} \cs{ExplSyntaxOff}
 %   \end{syntax}
-%   The \cs{ExplSyntaxOn} function switches to a category code
-%   regime in which spaces and new lines are ignored, and in which the colon (|:|)
+%   The \cs{ExplSyntaxOn} function modifies the current category code
+%   regime such that spaces and new lines are ignored, the colon (|:|)
 %   and underscore (|_|) are treated as \enquote{letters}, thus allowing
 %   access to the names of code functions and variables. Within this
-%   environment, |~| is used to input a space. The \cs{ExplSyntaxOff}
-%   reverts to the document category code regime.
+%   environment, the tilde (|~|) is used to input a space.
+%   The \cs{ExplSyntaxOff} function undoes the modifications made by
+%   \cs{ExplSyntaxOn} to the current category code regime
+%   and reverts |\endlinechar| to its previous value.
+%
+%   Nesting "\ExplSyntaxOn...\ExplSyntaxOff" constructs is not supported.
 %   \begin{texnote}
 %     Spaces introduced by~|~| behave much in the same way as normal
 %     space characters in the standard category code regime: they are
 %     ignored after a control word or at the start of a line, and
 %     multiple consecutive~|~| are equivalent to a single one.  However,
 %     |~|~is \emph{not} ignored at the end of a line.
+%
+%     For performance reasons, all the changes made to the current category code
+%     regime while running \meta{code} are not handled by \cs{ExplSyntaxOff}
+%     and must be undone explicitly.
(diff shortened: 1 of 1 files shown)


l3kernel/l3bootstrap.dtx
```diff
@@ -69,12 +69,12 @@
...
 %   \end{syntax}
-%   The \cs{ExplSyntaxOn} function switches to a category code
-%   regime in which spaces and new lines are ignored, and in which the colon (|:|)
+%   The \cs{ExplSyntaxOn} function modifies the current category code
+%   regime such that spaces and new lines are ignored, the colon (|:|)
 %   and underscore (|_|) are treated as \enquote{letters}, thus allowing
 %   access to the names of code functions and variables. Within this
 %   environment, |~| is used to input a space. The \cs{ExplSyntaxOff}
-%   reverts to the document category code regime.
+%   reverts to their standard values what \cs{ExplSyntaxOn} has modified.
```
josephwright, line 77: This sentence doesn't make sense.

l3kernel/l3cctab.dtx
```diff
@@ -54,7 +54,8 @@
 % range. For other engines, only the $8$-bit range ($0$--$255$) is covered by
 % such tables. The implementation of category code tables in \pkg{expl3}
 % also saves and restores the \TeX{} \tn{endlinechar} primitive value, meaning
-% they could be used for example to implement \cs{ExplSyntaxOn}.
+% they could be used for example to implement
+% a command similar to \cs{ExplSyntaxOn}.
```
josephwright, line 58: I'm not sure I follow why you want to change this: as far as I can see, it would work (though we've re-thought this area).

l3kernel/l3bootstrap.dtx
jlaurens: Is it more clear?
```
The \cs{ExplSyntaxOff} reverts what \cs{ExplSyntaxOn} has modified to their standard values.
```

josephwright: I thing something like `reverts category codes modified by \cs{ExplSyntaxOn} to their previous values.` might be better.

davidcarlisle: yes as the point is it doesn't reset to a pre-defined default it just undoes the previous ExplSyntaxOn. Also we should probably mention that `\endlinechar` as well as catcodes are (re)set.

jlaurens: not only the category codes but also the value of `\endlinechar`.
A different option is to detail exactly what is changed ?

l3kernel/l3cctab.dtx
davidcarlisle: @jlaurens  `\obeylines` is affected as the `\endlinechar`  is reset not because of the number of catcodes changed.

josephwright: @jlaurens The number of catcodes reset is not documented for `\ExplSyntax...` and arguable should be 'the lot' - it's a tradeoff with performance

jlaurens: @davidcarlisle 
Exactly. Whereas, `\ExplSyntax...` only changes the value of  `\endlinechar`, selecting  a cctab will change both the value of  `\endlinechar *and* the category code of ^^M.

davidcarlisle: That's an implementation detail: ExplSyntax could set the catcode of ^^M it doesn't matter much as ^^M will hardly ever appear in a document anyway since endlinechar is -1 it is not added at ends of lines

jlaurens: Things are more clear to me: next commit includes
- change in cctab removed
- more reference to the "current category code regime"
- `\endlinechar` added
- nesting not supported
- precaution while changing category code due to performance reasons

PR #1448, opened 2024-02-12
Check missing `\endgroup` at the end of `\DocInclude`

muzimuzhi:
This incorporates latex3/latex2e@60a6a368 (Prevent time wasted debugging latex3/latex2e@50f5256, 2021-10-20) by @PhelypeOleinik, with some modification.

Following https://github.com/latex3/latex2e/pull/1270#discussion_r1485597939, I find this is one of l3doc fixes/improvements made only in latex2e, hence not yet incorporated into `l3doc.dtx`.

Questions:
- Does `\DocInput` need a similar check? Currently `\DocInclude` uses `\DocInput` internally.
- Would `\DocInclude`/`\DocInput` be used in groups, so we need to compare the values of `\tex_currentgrouplevel:D` at the beginning and end of command, instead of simply checking if it's zero at the end?
- I noticed `\DocInclude` is not documented, and it seems it's inherited from `ltxdoc.cls`. Is `\DocInclude` an official `l3doc` command, or it's added for backwards compatibility?
- According to the eTeX manual, assignments to `\interactionmode` are always global. Then which int function is preferable in assigning `\tex_interactionmode:D`, `\int_set:Nn` or `\int_gset:Nn`? Currently local set is used in this PR and in several other places in this repo.

### DIFF
--- a/l3kernel/l3doc.dtx
+++ b/l3kernel/l3doc.dtx
@@ -3727,10 +3727,22 @@ and all files in that bundle must be distributed together.
 %
 % \begin{macro}{\DocInclude}
 %   More or less exactly the same as \tn{include}, but uses
-%   \tn{DocInput} on a \file{.dtx} file, not \tn{input} on a \file{.tex}
-%   file.
+%   \tn{DocInput} on a \file{.fdd} or \file{.dtx} file (without file
+%   extension), not \tn{input} on a \file{.tex} file.
 %
 %    \begin{macrocode}
+\msg_new:nnn { l3doc } { missing-endgroup }
+  {
+    \str_if_eq:VnTF \@currenvir { document }
+      {
+        There~are~\int_use:N \tex_currentgrouplevel:D
+        \c_space_tl unclosed~groups~in~#1.
+      }
+      {
+        The~\@currenvir \c_space_tl environment~on~line~\@currenvline
+        \c_space_tl doesn't~have~a~matching~\iow_char:N\\end{\@currenvir}.
+      }
+  }
 \NewDocumentCommand \DocInclude { m }
   {
     \relax\clearpage
@@ -3741,6 +3753,13 @@ and all files in that bundle must be distributed together.
     \int_compare:nNnTF \@auxout = \@partaux
       { \@latexerr{\string\include\space cannot~be~nested}\@eha }
       { \@docinclude {#1} }
+    % check missing \endgroup, e.g., missing "\end{macro}" in time
+    \int_compare:nNnF { \tex_currentgrouplevel:D } = { 0 }
+      {
+        \int_compare:nNnT { \tex_interactionmode:D } = { 0 }
+          { \int_set:Nn \tex_interactionmode:D { 1 } }
+        \msg_fatal:nnn { l3doc } { missing-endgroup } { \currentfile }
+      }
   }
 %    \end{macrocode}
 %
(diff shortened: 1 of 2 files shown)


josephwright:
>     * Does `\DocInput` need a similar check? Currently `\DocInclude` uses `\DocInput` internally.

One for @FrankMittelbach 

>     * Would `\DocInclude`/`\DocInput` be used in groups, so we need to compare the values of `\tex_currentgrouplevel:D` at the beginning and end of command, instead of simply checking if it's zero at the end?

I think we can say no: this would be a poorly-formed source.

>     * I noticed `\DocInclude` is not documented, and it seems it's inherited from `ltxdoc.cls`. Is `\DocInclude` an official `l3doc` command, or it's added for backwards compatibility?

I think it's fair to say that nothing in `l3doc` is truly documented!

>     * According to the eTeX manual, assignments to `\interactionmode` are always global. Then which int function is preferable in assigning `\tex_interactionmode:D`, `\int_set:Nn` or `\int_gset:Nn`? Currently local set is used in this PR and in several other places in this repo.

`\int_gset:Nn`for clarity I think: I can see that this might need tightening up.

josephwright (approved):
I'm happy to mop up the scoping business separately

muzimuzhi:
Another question: Recommend declaring messages in one place, under subsection "Messages"? Currently there're four `\msg_new:nnn` hanging out alone.
https://github.com/latex3/latex3/blob/cf7b9fbf5072a82ea054e104c8c0a436d2c7ad3d/l3kernel/l3doc.dtx#L1408

josephwright:
Yes ideally that would be the case

muzimuzhi:
Since full message (except for holes `#1`, `#2`, ...) is usually hidden in the message declaration, when a message is used only once, I find declaring it right before its use improves readability. Otherwise I have to search for the message name to see what roles the provided `#<n>`s play in full message.

l3kernel/l3doc.dtx
```diff
@@ -3741,6 +3753,13 @@ and all files in that bundle must be distributed together.
     \int_compare:nNnTF \@auxout = \@partaux
       { \@latexerr{\string\include\space cannot~be~nested}\@eha }
       { \@docinclude {#1} }
+    % check missing \endgroup, e.g., missing "\end{macro}" in time
+    \int_compare:nNnF { \tex_currentgrouplevel:D } = { 0 }
+      {
+        \int_compare:nNnT { \tex_interactionmode:D } = { 0 }
+          { \int_set:Nn \tex_interactionmode:D { 1 } }
+        \msg_fatal:nnn { l3doc } { missing-endgroup } { \currentfile }
+      }
```
blefloch, line 3762: Expansion is missing, it should be `\msg_fatal:nne` since the msg functions stringify their argument.

josephwright:
> Since full message (except for holes `#1`, `#2`, ...) is usually hidden in the message declaration, when a message is used only once, I find declaring it right before its use improves readability. Otherwise I have to search for the message name to see what roles the provided `#<n>`s play in full message.

The idea is at 'some point' that we document all of the messages - the idea is one can redefine at-will with known/documented args, etc. May never happen, of course ...

l3kernel/l3doc.dtx
muzimuzhi: Thanks! I opened #1451 to sort this.

FrankMittelbach:
> > ```
> > * Does `\DocInput` need a similar check? Currently `\DocInclude` uses `\DocInput` internally.
> > ```
> 
> One for @FrankMittelbach
> 
> > ```
> > * Would `\DocInclude`/`\DocInput` be used in groups, so we need to compare the values of `\tex_currentgrouplevel:D` at the beginning and end of command, instead of simply checking if it's zero at the end?
> > ```
> 
> I think we can say no: this would be a poorly-formed source.


sorry have to disagree.  `\DocInput` is simply a form of `\input` and naturally input can appear in a group, for example in some environment for guiding formatting.

PR #1459, opened 2024-02-17
Fix cache workflow by checking out main repo

muzimuzhi:
Otherwise no `.github/tl_packages` is available thus the `cache.yaml` workflow caches a texlive installation with no packages installed (only ~4M size), and this cache will never be used by other workflows.

image


zauguin:
No objection to merging this, but alternatively we could consider just removing the workflow.
It's only really needed if we regularly don't push for more than a week to the primary branch which isn't really a problem in this repo.

muzimuzhi:
Or reduce the frequency to once a week, which from my practice in `CTeX-org/ctex-kit` (see [`test.yml`](https://github.com/CTeX-org/ctex-kit/blob/0095bccd9e6901bad52d88ec3c7b0d0bfcb48250/.github/workflows/test.yml#L6-L8) and the [cache page](https://github.com/CTeX-org/ctex-kit/actions/caches)), still keeps one living cache at any time.

https://github.com/latex3/latex3/blob/53f2e851b20ae81de52cc0e45012e4e3abccbeff/.github/workflows/cache.yaml#L11-L13

josephwright:
I've merged, but I'd be happy just dropping - I don't think we need to be forcing the cache to be maintained, honestly

PR #1469, opened 2024-02-21
Declare `\l_graphics_search_ext_seq` in l3backend in non-LaTeX

muzimuzhi:
This seq was declared in `l3graphics` but is set in a per-backend basis. Since backend driver is always loaded, even without `l3graphics`, declaring it in `l3backend-graphics` avoids a setting-before-declaration issue.

### DIFF
--- a/l3backend/l3backend-graphics.dtx
+++ b/l3backend/l3backend-graphics.dtx
@@ -61,20 +61,28 @@
 %    \end{macrocode}
 %
 % \begin{macro}{\@@_backend_loaded:n}
+% \begin{variable}[no-user-doc]{\l_graphics_search_ext_seq}
 %   To deal with file load ordering. Plain users are on their own.
+%   In \LaTeX{} format, setting \cs{l_graphics_search_ext_seq} is
+%   postponed until after loading \pkg{l3graphics}. In other formats
+%   it is set immediately, so we declare it right now.
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_backend_loaded:n #1
+\cs_if_exist:NTF \hook_gput_code:nnn
   {
-    \cs_if_exist:NTF \hook_gput_code:nnn
+    \cs_new_protected:Npn \@@_backend_loaded:n #1
       {
         \hook_gput_code:nnn
           { package / l3graphics / after }
           { backend }
           {#1}
       }
-      {#1}
+  }
+  {
+    \cs_new_eq:NN \@@_backend_loaded:n \use:n
+    \seq_new:N \l_graphics_search_ext_seq
   }
 %    \end{macrocode}
+% \end{variable}
 % \end{macro}
 %
 % \subsection{\texttt{dvips} backend}
--- a/l3backend/l3backend-graphics.dtx
+++ b/l3backend/l3backend-graphics.dtx
@@ -60,7 +60,7 @@
 %<@@=graphics>
 %    \end{macrocode}
 %
-% \begin{macro}{\@@_backend_loaded:n}
+% \begin{macro}[no-user-doc]{\@@_backend_loaded:n}
 % \begin{variable}[no-user-doc]{\l_graphics_search_ext_seq}
 %   To deal with file load ordering. Plain users are on their own.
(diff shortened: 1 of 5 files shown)


l3kernel/l3debug.dtx
```diff
@@ -787,6 +787,7 @@
       \prop_set_eq:NN
       \prop_set_from_keyval:Nn
       \seq_set_eq:NN
+      \seq_set_from_clist:Nn
```
muzimuzhi, line 790: Wondering why this change along won't fail the checks.

muzimuzhi: Oh, the problem of setting `\l_graphics_search_ext_seq` before declaration only appears in non-LaTeX formats:

https://github.com/latex3/latex3/blob/05c22772e096b48aac4addfbb8b3b93a70e803ee/l3backend/l3backend-graphics.dtx#L66-L76

One of the backends
https://github.com/latex3/latex3/blob/05c22772e096b48aac4addfbb8b3b93a70e803ee/l3backend/l3backend-graphics.dtx#L83-L89

muzimuzhi: And, since generic loader (`expl3-generic.tex`) loads both the expl3 kernel (`expl3-code.tex`) and backend driver, there's currently no way to turn on debugging between loading kernel and driver.

Currently (before this PR) with little tweak `l3graphics.sty` can be loaded in plain TeX:
```tex
\input expl3-generic.tex\relax

\ExplSyntaxOn
\cs_new_eq:NN \ProvidesExplPackage \use_none:nnnn
\input l3graphics.sty\relax
\ExplSyntaxOff

\end
```
Compile it and you'll get an error
```
! LaTeX Error: Control sequence \l_graphics_search_ext_seq already defined.
```

Thus the question becomes: does `l3graphics` aim to be used in non-LaTeX formats (in the future), or we live with the small flaw that a variable would be set before declaration when `l3graphics` is used in unsupported formats? @josephwright

muzimuzhi: Now `\l_graphics_search_ext_seq` is declared in `l3graphics` in LaTeX format, and in backend drivers otherwise.

muzimuzhi:
Not sure if a new test file loading `l3graphics` in plainTeX is needed.

muzimuzhi:
Not sure if a new test file loading `l3graphics` in plainTeX is needed.

l3backend/l3backend-graphics.dtx
```diff
@@ -61,20 +61,28 @@
 %    \end{macrocode}
 %
 % \begin{macro}{\@@_backend_loaded:n}
+% \begin{variable}[no-user-doc]{\l_graphics_search_ext_seq}
```
muzimuzhi, line 64: This results in an undefined label, because the outer most setting wins.

> End of definition for `\__graphics_backend_loaded:n` and `\l_graphics_search_ext_seq`. This function is documented on page ??.)

I think this is a special use case that need not be sorted in `l3doc`, hence I will push a workaround.

josephwright:
I need to think about this one a bit. The model is that nothing 'general' should be declared in `l3backend` at all: if it's not engine-specific in some way, it belongs in the 'parent' module. But I see the timing issue here, at least at present. (I do have a PR open to move `l3graphics` to `l3kernel`, which would nicely address that.)

muzimuzhi:
> (I do have a PR open to move `l3graphics` to `l3kernel`, which would nicely address that.)

Oh I forgot it (see #1338).

Since moving `l3graphics` to `l3kernel` automatically avoids the issue this PR tries to resolve and before that `l3graphics` is provided only as a LaTeX2e package, not officially loadable in plainTeX, I will close this PR.

josephwright:
@muzimuzhi  I fyou are sure :) I'll rebase the other PR to perhaps remind the rest of the team ;)

muzimuzhi:
> > Since moving `l3graphics` to `l3kernel` automatically avoids the issue this PR tries to resolve
>
> If you are sure :)

See https://github.com/latex3/latex3/pull/1338#discussion_r1504625942. 😄

PR #1474, opened 2024-02-28
Make `\opacity_fill:n` and `\opacity_stroke:n` work without `pdfmanagement`

### LINKED ISSUE #1473: Only `\opacity_select:n` fails silently without pdfmanagement
When pdfmanagement is not loaded, `\opacity_select:n` does nothing (hence fails silently) but `\opacity_fill:n` and `\opacity_stroke:n` both raise errors, `Undefined control sequence ... \pdfmanagement_add:nnn ...`.

https://github.com/latex3/latex3/blob/138e85fd8b6f4cb0df2c3e4027dfa29fb8d7a9b6/l3backend/l3backend-opacity.dtx#L195-L200

```tex
%\DocumentMetadata{}
\documentclass{article}
\usepackage{l3opacity}

\ExplSyntaxOn
% fails silently
\hbox:n { \opacity_select:n { 1 } }
% raises error, "Undefined control sequence ... \pdfmanagement_add:nnn ..."
\hbox:n { \opacity_fill:n { 1 } }

\ExplSyntaxOff

\begin{document}
\end{document}
```

Both silent failure and inconsistent behavior are odd to me.

If we want to keep the silent failure, `\opacity_(fill|stroke):n` needs to be set to do nothing too (by `gset`ting `\__opacity_backend_fill_stroke:nn`). Otherwise that `gset` to nothing can be dropped so using any `l3opacity` command raises the same error if pdfmanagement is not loaded.

Caught when digging into the strange diff of `m3opacity002.lvt` in [CI run](https://github.com/latex3/latex3/actions/runs/8072194087) triggered by https://github.com/latex3/latex3/commit/138e85fd8b6f4cb0df2c3e4027dfa29fb8d7a9b6.

muzimuzhi:
Sort of oversight of https://github.com/latex3/latex3/commit/996bad17c0123540a5f5d8f8a84a82efd37a74c9 (Only use \pdfmanagement_add:nnn if active, 2021-08-02).

Note the main functionality of `l3opacity` is NOT tested at all, because some config makes `l3build save/check` find no pdfmanagement supporting file.

Fixes #1473.


muzimuzhi:
> Note the main functionality of `l3opacity` is NOT tested at all, because some config makes `l3build save/check` find no pdfmanagement supporting file.

Solved in #1480.

PR #1475, opened 2024-02-28
Change arg types of `\__kernel_color_backend_stack_(push|pop):`, n -> N

muzimuzhi:
- `\__kernel_color_backend_stack_push:nn` -> `Nn`
- `\__kernel_color_backend_stack_pop:N` -> `N`

They are used as if their first arguments take an int variable, not expression. Thus the refactoring.


muzimuzhi:
I'm cherry-picking [fc01051](https://github.com/latex3/latex3/pull/1474/commits/fc010512bf2376abd1dd6d025c7b51c6f35c1e6c) from #1474 to pass the CI.

josephwright:
The `n`-type argument is by-design here. Wherever we need to pass an integer _value_, we define the argument as an `intexpr`. For convenience, we allow the use of an unbraced variable name where that variable will supply a complete integer. That's true even at the backend: although there is a performance impact, here it is tiny and to my mind not worth having a one-off difference. (_cf._ the balance bewteen `fp` and `tl`, which at the backend is worth making an exception for).

muzimuzhi:
Got it.

PR #1481, opened 2024-02-29
Speed up l3opacity

muzimuzhi:
Maybe the re-bounding (`\fp_eval:n { min(max(0,#1),1) }`) in backend should be kept.


l3backend/l3backend-opacity.dtx
```diff
@@ -207,26 +195,26 @@
...
+% \begin{macro}{\@@_backend_fill_stroke:nn}
 %   For separate fill and stroke, we need to work out if we need to do
 %   more work or if we can stick to a single setting.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_backend_fill:n #1
   {
-    \@@_backend_fill_stroke:ee
-      { \fp_eval:n { min(max(0,#1),1) } }
-      \l_@@_backend_stroke_tl
+    \exp_args:Nno \@@_backend_fill_stroke:nn
```
josephwright, line 204: V-type

l3backend/l3backend-opacity.dtx
```diff
@@ -207,26 +195,26 @@
...
+      { #1 }
+      { \l_@@_backend_stroke_tl }
   }
 \cs_new_protected:Npn \@@_backend_stroke:n #1
   {
-    \@@_backend_fill_stroke:ee
-      \l_@@_backend_fill_tl
-      { \fp_eval:n { min(max(0,#1),1) } }
+    \exp_args:No \@@_backend_fill_stroke:nn
```
josephwright, line 210: V-type

l3experimental/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ this project uses date-based 'snapshot' version identifiers.
 ### Added
 - `\draw_path_replace_bb:`
 
+### Changed
+- `l3opacity` is sped up a bit
```
josephwright, line 14: Do we need to document in the backend ChangeLog? I assume we are relying on a change-in-behaviour in teh frontend so should do.

l3experimental/l3opacity/l3opacity.dtx
```diff
@@ -106,9 +106,17 @@
   {L3 Experimental opacity support}
 %    \end{macrocode}
 %
+% \begin{variable}
+%   Temporary storage.
+%    \begin{macrocode}
+\fp_new:N \l_@@_tmpa_fp
```
josephwright, line 112: We only have one temp var so just `tmp`

l3experimental/l3opacity/l3opacity.dtx
```diff
@@ -117,14 +125,13 @@
...
 \cs_new_protected:Npn \@@_select:nN #1#2
-  { \exp_args:Ne \@@_select_aux:nN { \fp_eval:n {#1} } #2 }
-\cs_new_protected:Npn \@@_select_aux:nN #1#2
   {
+    \fp_set:Nn \l_@@_tmpa_fp { #1 }
     \bool_lazy_or:nnTF
-      { \fp_compare_p:nNn {#1} < { 0 } }
-      { \fp_compare_p:nNn {#1} > { 1 } }
+      { \fp_compare_p:nNn { \l_@@_tmpa_fp } < { \c_zero_fp } }
+      { \fp_compare_p:nNn { \l_@@_tmpa_fp } > { \c_one_fp } }
```
josephwright, line 132: Drop the braces here - we allow them to be elided. Personally I don't particularly like using `\c_..._fp` other than in the tightest of loops, but I can live with it here.

l3experimental/l3opacity/l3opacity.dtx
```diff
@@ -117,14 +125,13 @@
 \cs_new_protected:Npn \opacity_stroke:n #1
   { \@@_select:nN {#1} \@@_backend_stroke:n }
 \cs_new_protected:Npn \@@_select:nN #1#2
-  { \exp_args:Ne \@@_select_aux:nN { \fp_eval:n {#1} } #2 }
-\cs_new_protected:Npn \@@_select_aux:nN #1#2
   {
+    \fp_set:Nn \l_@@_tmpa_fp { #1 }
```
josephwright, line 129: Why set a var rather than `e`-type expand `\fp_eval:n` using an auxiliary?

l3experimental/l3opacity/l3opacity.dtx
```diff
@@ -117,14 +125,13 @@
...
     \bool_lazy_or:nnTF
-      { \fp_compare_p:nNn {#1} < { 0 } }
-      { \fp_compare_p:nNn {#1} > { 1 } }
+      { \fp_compare_p:nNn { \l_@@_tmpa_fp } < { \c_zero_fp } }
+      { \fp_compare_p:nNn { \l_@@_tmpa_fp } > { \c_one_fp } }
       { \msg_error:nnn { opacity } { out-of-range } {#1} }
-      { #2 {#1} }
+      { \exp_args:Ne #2 { \fp_use:N \l_@@_tmpa_fp } }
```
josephwright, line 134: Again, I'd expect to see the value here passed using an auxiliary.

josephwright (changes requested):

l3backend/l3backend-opacity.dtx
FrankMittelbach: Is this for clean coding or necessary? "o" is always faster than "v" so I would lean towards "o" in core code.

l3experimental/l3opacity/l3opacity.dtx
muzimuzhi: Because `\fp_eval:n { <fp expr> }` is slower than `\fp_eval:n { <fp var> }`, provided `<fp var>` is set by `\fp_set:Nn <fp var> { <fp expr> }`.

Since here `#1` will be evaluated two or three times, store the parsed form in an fp-var saves some time.

muzimuzhi:
Requested changes about backend Changelog, `tmpa` -> `tmp` renaming, and brace dropping are resolved.

Remaining are expansion type `o` -> `V` and introduction of the fp variable.

Issue #476, opened 2018-09-06
l3doc: split into class and package


wspr:
There is much in l3doc that doesn't "need" to be in a class — we've had requests, for example, to be able to load the structural parts of l3doc with a different document class such as `ltugboat.cls`.

This should be fairly easy to make a clean division...

car222222:
Very likely this is feasible . . .
BUT due to staff shortages this train may not depart!:-)

zauguin:
I'm also in favor of splitting `l3doc`, but I would prefer a separate name (`l3codedoc` looks good) to avoid confusion.

FrankMittelbach:
That very much depends on what is provided: if I can take a document currently written for  the `l3doc` and can put its content into a document with a different class, say, `book` and only have to load the package `l3doc` to make it run then I think it  would be okay to use the same name instead of inventing a new one. In my opinion there shouldn't be any confusion in that casebut I'm okay with a longer name if the majority thinks differently.

wspr:
I definitely like the idea of a split, but I think a more user-friendly package name than l3doc would be good. We have packages like l3benchmark and so on, but these are exclusively (?) programming tools. If we have a (now supported and maintained) package for users to load, I think it should have a name more like “xdoc” (but that name is already taken)...

josephwright:
A 'reverse-dupe' of https://github.com/latex3/latex3/issues/1136

PR #1497, opened 2024-03-03
Generate standalone package (`l3doc.sty`) from l3doc.dtx

huangyxi:
When creating a document class, my intention is to utilize the class itself for setting up custom styles while employing commands pertinent to documenting implementation, such as \begin{macrocode}. Currently, I am using doc.sty to fulfill this requirement; however, the package only offers basic documentation functionality, lacking some useful utilities like \begin{macro} and \begin{variable}. Therefore, I have adjusted the DocStrip instructions to enable DocTeX to generate a standalone package (.sty).

I am unsure if this modification is appropriate, so I have only made minimal changes to the file. If deemed suitable, I am prepared to continue updating the corresponding documentation related to this modification.

Minimal working example:
```tex
% \iffalse
\documentclass{article}
\usepackage{l3doc}
\begin{document}
	\DocInput{\jobname.dtx}
\end{document}
% \fi
%
% \section{Implementation}
% \begin{variable}{\l_demo_tl}
%    \begin{macrocode}
\tl_new:N \l_demo_tl
%    \end{macrocode}
% \end{variable}
```

### DIFF
--- a/l3kernel/l3doc.dtx
+++ b/l3kernel/l3doc.dtx
@@ -48,6 +48,7 @@ and all files in that bundle must be distributed together.
 \postamble
 \endpostamble
 \generate{\file{l3doc.cls}{\from{l3doc.dtx}{class,cfg}}}
+\generate{\file{l3doc.sty}{\from{l3doc.dtx}{package,cfg}}}
 %\generate{\file{l3doc.ist}{\from{l3doc.dtx}{docist}}}
 \ifx\fmtname\nameofplainTeX
   \expandafter\endbatchfile
@@ -71,9 +72,20 @@ and all files in that bundle must be distributed together.
 % This isn't included in the typeset documentation because it's a bit
 % ugly:
 %<*class>
-\ProvidesExplClass{l3doc}{2024-02-20}{}
+\ProvidesExplClass
+%</class>
+%<*package>
+\ProvidesExplPackage
+%</package>
+%<*class|package>
+  {l3doc}{2024-02-20}{}
+%</class|package>
+%<*class>
   {L3 Experimental documentation class}
 %</class>
+%<*package>
+  {L3 Experimental documentation package}
+%</package>
 % \fi
 %
 % \title{The \cls{l3doc} class -- experimental\thanks{%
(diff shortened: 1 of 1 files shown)


muzimuzhi:
Previous discussion: #476.

FrankMittelbach:
> Previous discussion: #476.

of limited relevance nowadays I would say, because that is before doc v3 came out.

FrankMittelbach:
> Currently, I am using doc.sty to fulfill this requirement; however, the package only offers basic documentation functionality, lacking some useful utilities like \begin{macro} and \begin{variable}. 

The doc package is not lacking "macro" and with doc version 3 there is a general mechanism to support other types of data, e.g., "variable" via `\NewDocElement`.

> Therefore, I have adjusted the DocStrip instructions to enable DocTeX to generate a standalone package (.sty).
> 
> I am unsure if this modification is appropriate, so I have only made minimal changes to the file. If deemed suitable, I am prepared to continue updating the corresponding documentation related to this modification.

The l3doc class was written as some internal hack to to provide some initial support for the documentation of the l3 kernel code. It was never meant (and still isn't) as a general purpose class even though people have used it as is given that it was around. Some of the methods and ideas inside aren't what we would do going forward and thus I don't think I like to see a package version providing that code standalone.

I agree wtih you that that there should probably be a package that loads `doc` and sets up a suitable set of doc elements for typical code documentation beyond  `macro` and `environment`  and possibly some other helper environments, but not exactly those currently in l3doc -- it is not a suitable place to start from.

huangyxi:
Thank you for your efficient and detailed response, and your continuous contribution to the great LaTeX project.

I acknowledge that generating a standalone package would increase code replication and maintenance workload. However, in my opinion, if a documentation class could handle more complex tasks well, including internal commands that may cause namespace conflicts, it should perform even better with normal classes or packages.

As an alternative to better reuse the handy utilities defined in `l3doc`, I have created a simple package called [`useclass`](https://github.com/huangyxi/useclass) to load classes (`.cls`) as packages (`.sty`). This solution works well with `l3doc.cls`.

Minimal working example:
```tex
% \iffalse
\documentclass{article}
\usepackage{useclass}
\useclass{l3doc}
\begin{document}
	\DocInput{\jobname.dtx}
\end{document}
% \fi
%
% \section{Implementation}
% \begin{variable}{\l_demo_tl}
%    \begin{macrocode}
\tl_new:N \l_demo_tl
%    \end{macrocode}
% \end{variable}
```

Anyway, your contributions to the LaTeX community are commendable, and I'm here to support in any way I can.

PR #1498, opened 2024-03-05
Reset active matrix marker in `\draw_transform_matrix_reset:`

muzimuzhi:
Need a Changelog entry?

### DIFF
--- a/l3experimental/l3draw/l3draw-transforms.dtx
+++ b/l3experimental/l3draw/l3draw-transforms.dtx
@@ -110,10 +110,10 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \draw_transform_matrix_reset:
   {
-    \fp_set:Nn \l_@@_matrix_a_fp { 1 }
+    \fp_set_eq:NN \l_@@_matrix_a_fp \c_one_fp
     \fp_zero:N \l_@@_matrix_b_fp
     \fp_zero:N \l_@@_matrix_c_fp
-    \fp_set:Nn \l_@@_matrix_d_fp { 1 }
+    \fp_set_eq:NN \l_@@_matrix_d_fp \c_one_fp
     \bool_set_false:N \l_@@_matrix_active_bool
   }
 \cs_new_protected:Npn \draw_transform_shift_reset:
--- a/l3experimental/l3draw/l3draw-transforms.dtx
+++ b/l3experimental/l3draw/l3draw-transforms.dtx
@@ -114,6 +114,7 @@
     \fp_zero:N \l_@@_matrix_b_fp
     \fp_zero:N \l_@@_matrix_c_fp
     \fp_set:Nn \l_@@_matrix_d_fp { 1 }
+    \bool_set_false:N \l_@@_matrix_active_bool
   }
 \cs_new_protected:Npn \draw_transform_shift_reset:
   {


l3experimental/l3draw/l3draw-transforms.dtx
```diff
@@ -110,10 +110,11 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \draw_transform_matrix_reset:
   {
-    \fp_set:Nn \l_@@_matrix_a_fp { 1 }
+    \fp_set_eq:NN \l_@@_matrix_a_fp \c_one_fp
     \fp_zero:N \l_@@_matrix_b_fp
     \fp_zero:N \l_@@_matrix_c_fp
-    \fp_set:Nn \l_@@_matrix_d_fp { 1 }
+    \fp_set_eq:NN \l_@@_matrix_d_fp \c_one_fp
```
josephwright, line 116: I see the thinking here but I'm not so keen. In general, the balance between `\<var>_set:Nn` and `\<var>_set_eq:NN` for numerical data is tricky - I tend to think `set` for places you have an explicit value, `set_eq` when you are doing save-and-restore or similar: makes it clear what's going on. Here, I get why you want to use `set_eq` as there is a definite performance gain. But `l3draw` isn't exactly fast and this is a relatively 'high level' function. So I think the current approach has clarity on it's side: but I am happy to hear counter-arguments.

l3experimental/l3draw/l3draw-transforms.dtx
```diff
@@ -110,10 +110,11 @@
...
+    \bool_set_false:N \l_@@_matrix_active_bool
```
josephwright, line 117: This part is nice and clear: definitely needed :)

josephwright (approved):
I'll see what others think before merging: as I say, my major comment is a matter of taste

FrankMittelbach:
> I'll see what others think before merging: as I say, my major comment is a matter of taste

I'm usually in favor of speed improvements, but here I guess I'm with Joseph in both cases

josephwright:
@FrankMittelbach favours speed: I'll merge

FrankMittelbach:
looks like I wasn't clear ... I "usually favor" but I agree with both your comments so I wouldn't do the first part

FrankMittelbach:
but I guess either are fine so ...

josephwright:
@FrankMittelbach  OK, I've backed out and sse can see if there are more comments :)

l3experimental/l3draw/l3draw-transforms.dtx
muzimuzhi: If these initial values are named constants like `\c__draw_matrix_scale_init_fp` and `\c__draw_matrix_slant_init_fp`, using `\fp_set_eq:NN` is more obvious.

josephwright: It's mainly used inside `l3fp` for return values, then also for comparisons

Skillmon: Then why document it? Do you see my point? I don't know how to further communicate it... I agree that I'd not use `\fp_set_eq:NN` for arbitrary initial values, but in the case of a constant following (like `\c_one_fp`) things are arguably different. It is instantly apparent, that if I set an `fp` equal to a constant I don't want to temporarily store the value of another variable to later restore it, but that I want to assign an initial value for something else!

PR #1523, opened 2024-03-27
Expand object names in backend layer

### LINKED ISSUE #1521: l3pdf feature request: providing variants for pdf-object-related functions that expand their object name argument
I think it would be useful if l3pdf functions that take a user provided object name, such as `\pdf_object_new:n`, `\pdf_object_write:nn[ne]`, `\pdf_object_ref:n` etc, generally expanded their object name argument or l3pdf at least provided corresponding `e` variants. I guess that in third-party packages such names for referencing PDF objects are often composed of macros and character tokens stringed together.

muzimuzhi:
Addresses comments I made in bcfbc26b (Expand object names (closes #1521), 2024-03-26). Refactoring is done in the last commit.

When refactoring, I found `\pdf_object_if_exist:n(TF)` is the only kernel function which makes direct use of int `\c__pdf_object_<object>_int` otherwise hidden in `l3backend-pdf.dtx`.
https://github.com/latex3/latex3/blob/a32300d602908dc6115f9d709c686db841593cd7/l3kernel/l3pdf.dtx#L349-L354

### DIFF
--- a/l3backend/CHANGELOG.md
+++ b/l3backend/CHANGELOG.md
@@ -6,6 +6,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+- Expand object names in `\pdf_object_...` functions (issue \#1521)
+
 ## [2024-03-14]
 
 ### Removed
--- a/l3backend/l3backend-pdf.dtx
+++ b/l3backend/l3backend-pdf.dtx
@@ -127,11 +127,11 @@
   {
     \int_gincr:N \g_@@_backend_object_int
     \int_const:cn
-      { c_@@_object_ \tl_to_str:n {#1} _int }
+      { c_@@_object_ #1 _int }
       { \g_@@_backend_object_int }
   }
 \cs_new:Npn \@@_backend_object_ref:n #1
-  { { pdf.obj \int_use:c { c_@@_object_ \tl_to_str:n {#1} _int } } }
+  { { pdf.obj \int_use:c { c_@@_object_ #1 _int } } }
 %    \end{macrocode}
 % \end{macro}
 % \end{macro}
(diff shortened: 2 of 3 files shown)


josephwright:
There are other things going to land here: I'll close this particular PR as it will conflict with them.

josephwright:
See https://github.com/latex3/latex3/compare/object-tracking for where things are going: it's a work-in-progress and may be a few days

josephwright:
The wider plan is to have two ways of generating objects: the current one and a method where they have a type and serial number, which then only needs us to map their serial number to the object number. Hence moving the tracking of object numbers out of the backend. (We need this for tagging, etc. )

muzimuzhi:
Loosely related: for backwards compatibility, expansions of object names in deprecated functions `\pdf_object_new:nn` and `\pdf_object_write:nn` may have to be disabled, like
```diff
diff --git a/l3kernel/l3deprecation.dtx b/l3kernel/l3deprecation.dtx
index 281a2f81..e24741ad 100644
--- a/l3kernel/l3deprecation.dtx
+++ b/l3kernel/l3deprecation.dtx
@@ -330,13 +330,13 @@
 \cs_new_protected:Npn \pdf_object_new:nn #1#2
   {
     \prop_gput:Nnn \g_@@_object_prop {#1} {#2}
-    \@@_backend_object_new:n {#1}
+    \@@_backend_object_new:n { \exp_not:n {#1} }
   }
 \__kernel_patch_deprecation:nnNNpn { 2022-08-30 } { [\pdf_object_write:n] }
 \cs_new_protected:Npn \pdf_object_write:nn #1#2
   {
     \exp_args:Nne \@@_backend_object_write:nnn
-      {#1} { \prop_item:Nn \g_@@_object_prop {#1} } {#2}
+      { \exp_not:n {#1} } { \prop_item:Nn \g_@@_object_prop {#1} } {#2}
     \bool_gset_true:N \g_@@_init_bool
   }
 \cs_generate_variant:Nn \pdf_object_write:nn { nx }
```

josephwright:
> Loosely related: for backwards compatibility, expansions of object names in deprecated functions `\pdf_object_new:nn` and `\pdf_object_write:nn` may have to be disabled, like

Don't see that: this is a refinement of the way object are handled in general, so includes the deprecated functions.

josephwright:
Should have added that as always many thanks for the PR - sorry that this one gets 'caught out' by some other changes I'm working on

PR #1524, opened 2024-03-28
Add index-based PDF object tracking

josephwright:


l3kernel/l3pdf.dtx
```diff
@@ -314,38 +360,59 @@
...
+  { \@@_object_new:e {#1} }
+\cs_new_protected:Npn \@@_object_new:n #1
   {
-    \exp_args:Ne \@@_backend_object_new:n {#1}
+    \@@_backend_object_new:
+    \int_const:cn
+      { c_@@_object_ #1 _int }
+      { \g_@@_backend_object_int }
   }
+\cs_generate_variant:Nn \@@_object_new:n { e }
```
muzimuzhi, line 389: Seems `\__pdf_object_new:n` can be merged to `\pdf_object_new:n`.

The `#1` is always fully expanded by the `c`-variant of `\int_const:cn` so `e`-type expansion is not necessary.

l3kernel/l3pdf.dtx
```diff
@@ -89,6 +91,58 @@
...
 %
+% \begin{function}[EXP, pTF, added = 2020-05-15]{\pdf_object_if_exist:n}
+%   \begin{syntax}
+%    \cs{pdf_object_if_exist_p:n} \Arg{object}
+%    \cs{pdf_object_if_exist:nTF} \Arg{object}
+%   \end{syntax}
+%   Tests whether an object with name \Arg{object} has been defined.
+% \end{function}
+%
+% \subsection{Indexed objects}
```
muzimuzhi, line 102: Apart from
- `\pdf_object_new_indexed:nn {<class>} {<index>}`
- `\pdf_object_write_indexed:nnnn {<class>} {<index>} {<type>} {<content>}`
- `\pdf_object_ref_indexed:nn {<class>} {<index>}`

for indexed objects, is a new conditional which checks existence of indexed object needed? For example

```tex
\pdf_object_index_if_exist:nnTF {<class>} {<index>} {<true code>} {<false code>}
```

Maybe even `\pdf_object_class_if_exist:nTF {<class>} {<true code>} {<false code>}`.

l3kernel/l3pdf.dtx
```diff
@@ -89,6 +91,58 @@
...
+%   referenced from this point on, and written later using
+%   \cs{pdf_object_write_indexed:nnnn}.
+% \end{function}
+%
+% \begin{function}[added = 2024-03-28]
+%   {\pdf_object_write_indexed:nnnn, \pdf_object_write_indexed:nnne}
+%   \begin{syntax}
+%     \cs{pdf_object_write_indexed:nnnn}  \Arg{class} \Arg{index} \Arg{type} \Arg{content}
+%   \end{syntax}
+%   Writes the \meta{content} as content of the onject of \meta{class} and
```
muzimuzhi, line 121:
```suggestion
%   Writes the \meta{content} as content of the object of \meta{class} and
```

l3kernel/l3pdf.dtx
josephwright: Agreed - changed

l3kernel/l3pdf.dtx
josephwright: I'd wondered about that but assume that given the use case it's unlikely: at least I'd want to see a use I think before adding.

muzimuzhi (changes requested):
Several typos and a refactor idea.

l3kernel/l3pdf.dtx
```diff
@@ -89,6 +91,58 @@
...
+%    \cs{pdf_object_if_exist_p:n} \Arg{object}
+%    \cs{pdf_object_if_exist:nTF} \Arg{object}
+%   \end{syntax}
+%   Tests whether an object with name \Arg{object} has been defined.
+% \end{function}
+%
+% \subsection{Indexed objects}
+%
+% Objects can be created using a \meta{class} and \emph{index}. For large
+% families of objects, this approach is more efficient than using individual names.
```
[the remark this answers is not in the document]
muzimuzhi: Nope, only `<index>` accepts an `<int expr>`. `<class>` accepts a string.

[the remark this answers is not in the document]
muzimuzhi: Updated suggestion

```suggestion
% Objects can be created using a \meta{class} (\meta{str}) and
% an \meta{index} (\meta{int expr}). For large families of objects,
% this approach is more efficient than using individual names.
```

l3kernel/l3pdf.dtx
```diff
@@ -354,12 +421,112 @@
...
+        \intarray_new:cn { g_@@_object_ #1 _ #2 _intarray }
+          \c_@@_object_block_int
+      }
+    \intarray_gset:cnn { g_@@_object_ #1 _ #2 _intarray }
+      {#3} \g_@@_backend_object_int
+  }
+\cs_new_protected:Npn \pdf_object_write_indexed:nnnn #1#2#3#4
+  {
+    \@@_object_indexed_split:nnN {#1} {#2}
+      \@@_object_new_write:nnnnn
```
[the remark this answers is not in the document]
muzimuzhi: This typo shown that indexed pdf object functions are not tested.

l3kernel/l3pdf.dtx
FrankMittelbach: I agree, we should wait with adding anything like that. So far the only real use case for indexed objects is tagging where they wouldn't be needed and I don't expect many others (and even if there are other types, you wouldn't ask for a random index if that exists or not imo). So let's wait and see.

car222222:
Surely the idea of “classes of int-arrays”, as an extension of single int-arrays, is not intended to be exclusive to numbered pdf objects.  
It should therefore be provided as an extension to general int-arrays, and only used here in the pdf module.

josephwright:
@car222222 Well it's a pretty simple idea - not sure it needs a generalised interface at this stage. Nothing internal is used, it's just set up in terms of making blocks.

l3kernel/l3pdf.dtx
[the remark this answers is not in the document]
josephwright: The class is expanded, so I will make that clear: but it's not a string.

l3kernel/l3pdf.dtx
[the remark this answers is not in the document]
josephwright: Oh, I know that :) The issue is that most testing for objects needs PDF management, and we can't do that in generic `expl3`.

l3kernel/l3pdf.dtx
```diff
@@ -354,12 +421,112 @@
...
+          {
+            \int_eval:n
+              { \int_div_truncate:nn {#2} \c_@@_object_block_int }
+          }
+          {
+            \int_eval:n
+              { \int_mod:nn {#2} \c_@@_object_block_int }
+          }
+      }
+  }
```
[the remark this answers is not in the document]
josephwright: Altered, and I've addressed an out-by-one issue :)

car222222:
@josephwright As you say:  it's a pretty simple idea 
and:  Nothing internal is used

so there are two good reasons for not confining it to within this module.  
Unless you mean that it is simple enough that every usage can set it up individually? That anyway seems a strange approach: reinventing wheels!

josephwright:
@car222222 

> so there are two good reasons for not confining it to within this module. Unless you mean that it is simple enough that every usage can set it up individually? That anyway seems a strange approach: reinventing wheels!

The point really is use cases: we have a particular need in tracking PDF objects to have a mechanism to allocate `intarray` storage in a way that can expand but where blocks make sense. I think a  more general requirement would be dynamically-allocated `intarray` storage - that would likely be approached in a similar way to the `prop` change, keeping the public API as-is but adding internally. Could of course still be based on blocks, but wouldn't be explicit.  In any case, that's a feature request for `intarray` structures, and doesn't directly need to be addressed here.

car222222:
Obviously that part is a feature request for intarray.  
But when (hopefully very soon) that request is dealt with, will the pdf module use it, or will it still do its own thing? (Aside:  There are other useful extensions to imtarray that could be also be added at this stage.)

FrankMittelbach (commented):
Only looked it over very briefly and  so this is not a real review. But I didn't see anything that stuck out (one minor comment though).

l3kernel/l3pdf.dtx
```diff
@@ -354,12 +411,113 @@
...
+            {
+              g_@@_object_ #1 _
+              \int_eval:n
+                { \int_div_truncate:nn {#2} \c_@@_object_block_int }
+              _intarray
+            }
+            {
+              \int_eval:n
+                { \int_mod:nn {#2} \c_@@_object_block_int } + 1
+            }
```
muzimuzhi, line 486: This group is the 2nd argument of `\exp_not:N #3`, thus indent it with the same level as `\exp_not:c`.

```suggestion
          {
            \int_eval:n
              { \int_mod:nn {#2} \c_@@_object_block_int } + 1
          }
```

l3kernel/l3pdf.dtx
```diff
@@ -354,12 +411,113 @@
...
+      {
+        \intarray_new:Nn #1
+          \c_@@_object_block_int
+      }
+    \intarray_gset:Nnn #1 {#2} \g_@@_backend_object_int
+  }
+\cs_new_protected:Npn \pdf_object_write_indexed:nnnn #1#2#3#4
+  {
+    \@@_object_indexed_split:nnN {#1} {#2}
+      \@@_object_new_write_indexed:Nnnn
```
muzimuzhi, line 450: No `new`.
```suggestion
      \@@_object_write_indexed:Nnnn
```

l3kernel/l3pdf.dtx
```diff
@@ -314,38 +362,47 @@
...
-\cs_new:Npn \pdf_object_ref:n #1 { \@@_backend_object_ref:n {#1} }
-\cs_new_protected:Npn \pdf_object_unnamed_write:nn #1#2
-  {
-    \exp_args:Ne \@@_backend_object_now:nn {#1} {#2}
-    \bool_gset_true:N \g_@@_init_bool
+\cs_new:Npn \pdf_object_ref:n #1
+ {
+    \exp_args:Nv \@@_backend_object_ref:n
+      { c_@@_object_ #1 _int }
   }
```
muzimuzhi, line 397:
```suggestion
\cs_new:Npn \pdf_object_ref:n #1
 {
   \exp_args:Nv \@@_backend_object_ref:n
     { c_@@_object_ #1 _int }
 }
```

l3backend/l3backend-pdf.dtx
```diff
@@ -156,7 +143,7 @@
 \cs_new_protected:Npn \@@_backend_object_write:nnn #1#2#3
   {
     \@@_backend_object_write_aux:nnn
-      { \@@_backend_object_ref:n {#1} }
+      { \pdf_object_ref:n {#1} }
```
muzimuzhi, line 146: Why this change? The `#1` here is already an int value, either value of `c_@@_object_ #1 _int`
```tex
\cs_new_protected:Npn \pdf_object_write:nnn #1#2#3
  {
    \exp_args:Nv \@@_backend_object_write:nnn
      { c_@@_object_ #1 _int } {#2} {#3}
    \bool_gset_true:N \g_@@_init_bool
  }
```
or the object number stored in intarray
```tex
\cs_new_protected:Npn \@@_object_write_indexed:Nnnn #1#2#3#4
  {
    \exp_args:Ne \@@_backend_object_write:nnn
      { \intarray_item:Nn #1 {#2} }
      {#3} {#4}
    \bool_gset_true:N \g_@@_init_bool
  }
```

muzimuzhi, line 146: Why this change? The `#1` here is already an int value, either value of `c_@@_object_ #1 _int`
```tex
\cs_new_protected:Npn \pdf_object_write:nnn #1#2#3
  {
    \exp_args:Nv \@@_backend_object_write:nnn
      { c_@@_object_ #1 _int } {#2} {#3}
    \bool_gset_true:N \g_@@_init_bool
  }
```
or the object number stored in intarray
```tex
\cs_new_protected:Npn \@@_object_write_indexed:Nnnn #1#2#3#4
  {
    \exp_args:Ne \@@_backend_object_write:nnn
      { \intarray_item:Nn #1 {#2} }
      {#3} {#4}
    \bool_gset_true:N \g_@@_init_bool
  }
```
------

```suggestion
      { \@@_backend_object_ref:n {#1} }
```

l3kernel/l3pdf.dtx
muzimuzhi, line 397:
```suggestion
\cs_new:Npn \pdf_object_ref:n #1
  {
    \exp_args:Nv \@@_backend_object_ref:n
      { c_@@_object_ #1 _int }
  }
```

l3kernel/l3pdf.dtx
muzimuzhi, line 450: No `new`.
```suggestion
      \@@_object_write_indexed:Nnnn
```

l3kernel/l3pdf.dtx
muzimuzhi, line 486: Reduce indent by one level, and move `+ 1` inside `\int_eval:n { ... }`.

```suggestion
          {
            \int_eval:n
              { \int_mod:nn {#2} \c_@@_object_block_int + 1 }
          }
```

muzimuzhi (changes requested):
Still some typos. I posted them as comments again... Now reposting them as a review.

l3kernel/l3pdf.dtx
```diff
@@ -89,6 +91,60 @@
...
+%    \cs{pdf_object_if_exist_p:n} \Arg{object}
+%    \cs{pdf_object_if_exist:nTF} \Arg{object}
+%   \end{syntax}
+%   Tests whether an object with name \Arg{object} has been defined.
+% \end{function}
+%
+% \subsection{Indexed objects}
+%
+% Objects can be created using a \meta{class} and \emph{index}; the
+% \meta{class} argument should expand to character tokens, whilst the
```
muzimuzhi, line 105:
```suggestion
% Objects can also be created using a pair of \meta{class} and \emph{index};
% the \meta{class} argument should expand to character tokens, whilst the
```

muzimuzhi (changes requested):
Reviewed documentation. All done.

l3kernel/l3pdf.dtx
josephwright: I think the indent level is right: this will be an argument of the function passed as `#3` so needs an additional level

josephwright:
@muzimuzhi  Hopefully I got all of your suggestions!

l3kernel/l3pdf.dtx
```diff
@@ -354,10 +411,111 @@
...
+            {
+              g_@@_object_ #1 _
+              \int_eval:n
+                { \int_div_truncate:nn {#2} \c_@@_object_block_size_int }
+              _intarray
+            }
+            {
+              \int_eval:n
+                { \int_mod:nn {#2} \c_@@_object_block_size_int + 1 }
+            }
```
muzimuzhi, line 486: (continued from https://github.com/latex3/latex3/pull/1524#discussion_r1544902762)

```tex
% expected
\use:e
  {
    \exp_not:N #3
      {...}
      {...}
  }

\use:e
  {
    \exp_not:N #3
      \exp_not:c {...}
      {...}
  }

\use:e
  {
    \exp_not:N #3
      \exp_not:c
        {
          ...
        }
      {
        ...
      }
  }

% current
\use:e
  {
    \exp_not:N #3
      \exp_not:c
        {
          ...
        }
        % extra level of indent
        {
          ...
        }
  }
```

------

```suggestion
        \exp_not:N #3
          \exp_not:c
            {
              g_@@_object_ #1 _
              \int_eval:n
                { \int_div_truncate:nn {#2} \c_@@_object_block_size_int }
              _intarray
            }
          {
            \int_eval:n
              { \int_mod:nn {#2} \c_@@_object_block_size_int + 1 }
          }
```

l3kernel/l3pdf.dtx
muzimuzhi: But currently this 2nd argument of `#3` has two extra levels compared to `\exp_not:N #3`.

```tex
% expected
\use:e
  {
    \exp_not:N #3
      {...}
      {...}
  }

\use:e
  {
    \exp_not:N #3
      \exp_not:c {...}
      {...}
  }

\use:e
  {
    \exp_not:N #3
      \exp_not:c
        {
          ...
        }
      {
        ...
      }
  }

% current
\use:e
  {
    \exp_not:N #3
      \exp_not:c
        {
          ...
        }
        % extra level of indent
        {
          ...
        }
  }
```

l3kernel/l3pdf.dtx
```diff
@@ -89,6 +91,60 @@
...
+%
+% \begin{function}[added = 2024-03-28]
+%   {\pdf_object_write_indexed:nnnn, \pdf_object_write_indexed:nnne}
+%   \begin{syntax}
+%     \cs{pdf_object_write_indexed:nnnn}  \Arg{class} \Arg{index} \Arg{type} \Arg{content}
+%   \end{syntax}
+%   Writes the \meta{content} as content of the object of \meta{class} and
+%   \meta{index}. Depending on the
+%   \meta{type} declared for the object, the format required for the
+%   \meta{data} will vary
```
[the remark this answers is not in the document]
muzimuzhi: > Docs of \pdf_object_write:nnn and \pdf_object_unnamed_write:nn have the same problem.

Can be addressed separately.

muzimuzhi (approved):
- Continued discussion about extra level of indent https://github.com/latex3/latex3/pull/1524#discussion_r1544923430. Not a barrier to merging.
- `\meta{data}` -> `\meat{content}` https://github.com/latex3/latex3/pull/1524#discussion_r1544925674, can be addressed separately.

l3kernel/l3pdf.dtx
```diff
@@ -354,12 +409,112 @@
   }
 %    \end{macrocode}
 % \end{macro}
+%
+% \begin{variable}{\c_@@_object_block_int}
+%   Sets the block size used for managing indexed objects.
```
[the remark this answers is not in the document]
car222222: A good change!

l3kernel/l3pdf.dtx
car222222: Maybe not ask for a “random” index, but in general there may be the possibility that the indexes are not simply a complete run of integers starting at 1?

l3kernel/l3pdf.dtx
```diff
@@ -101,8 +101,8 @@
 %
 % \subsection{Indexed objects}
 %
-% Objects can be created using a \meta{class} and \emph{index}; the
-% \meta{class} argument should expand to character tokens, whilst the
+% Objects can also be created using a pair of \meta{class} and \emph{index};
```
car222222, line 104: This does not look quite right: 
Is the “also” needed here (not looked at the previous sentence sentences)? 
Bette:
. . . using a class, index pair, as follows: . . .

muzimuzhi: > Is the “also” needed here (not looked at the previous sentence sentences)?

Yes, as this is the second way (`\subsection{Indexes objects}`) to create PDF objects. The first way is named objects (`\subsection{Named objects}`), see line 54 of `l3kernel/l3pdf.dtx` in this PR.

u-fischer:
I'm wondering if the indexed objects could/should be reserved in blocks in advance, e.g. 1000. That would mean that the object numbers of the next 1000 structures are already known and other objects, e.g. from filespec dictionaries or OBJR from links wouldn't shift that numbers around. Object numbers are sometimes stored in labels in the aux, e.g. for /Ref keys and if they stay more stable it could reduce the number of compilations needed.

l3kernel/l3pdf.dtx
car222222: Yes, I see the reasoning now.
But in fact “indexed” are just a special case of “named” so the whole se timon needs rejigging.

josephwright: 'Named' are entirely arbitrary: better ways of phrasing welcome

l3kernel/l3pdf.dtx
```diff
@@ -89,6 +91,60 @@
...
+%   \end{syntax}
+%   Tests whether an object with name \Arg{object} has been defined.
+% \end{function}
+%
+% \subsection{Indexed objects}
+%
+% Objects can also be created using a pair of \meta{class} and \emph{index};
+% the \meta{class} argument should expand to character tokens, whilst the
+% \meta{index} is an \meta{int expr}. For large families of objects, this
+% approach is more efficient than using individual names.
```
muzimuzhi, line 107: Mention that `<index>` starts at 1.

```suggestion
% \meta{index} is an \meta{int expr} and starts at $1$. For large
% families of objects, this approach is more efficient than using
% individual names.
```

muzimuzhi (changes requested):
A stray backtick and doc suggestion to mention `<index>` starts at 1.

l3kernel/l3pdf.dtx
```diff
@@ -354,12 +412,122 @@
...
+%    \begin{macrocode}
+\cs_new_protected:Npn \pdf_object_new_indexed:nn #1#2
+  {
+    \@@_object_indexed_split:nnN {#1} {#2}
+      \@@_object_new_indexed:Nn
+  }
+\cs_new_protected:Npn \@@_object_new_indexed:Nn #1#2
+  {
+    \@@_backend_object_new:
+    \intarray_if_exist:NF #1
```
u-fischer, line 441: `\intarray_if_exist:NF`  is not defined.

josephwright:
We good-to-go here? I think the functionality is now sorted, what about the naming?

car222222:
Yes, the naming may still be problematic!

car222222:
Why are these two wordings distinct? Do they mean exactly the same? :

1.  An \meta{object} name should fully expand to tokens suitable for use in a
 label-like context.
(this one sounds horribly vague!) 

2.  the \meta{class} argument should expand to character tokens

car222222:
This section of the file probably needs some rejigging, with further explanation of (or references for this) the concept and some details of PDF objects and the related LaTeX-side commands for these three separate stages: set up (including the naming possibilities), writing their contents, and referring to them. It should probably also outline what is not available in the current state of this stuff. 

I shall produce a draft outline of this approach as soon as possible.

car222222:
I understand that it may not be feasible to change these names at this stage, but here is an outline of a more suitable structure (and some names) for this section (further, more detailed, content suggestions will follow):

Subsections (not actual titles):
1.  Objects and object references (by number) in PDF (and their use and setup in LaTeX).
2.  LaTeX naming of objects (uses and methods), 
      with a subsubsection on using class+number as names, containing 
         details of why and how, including the importance of the underlying use of intarrays.
3.  Contents of objects: with details of what is provided and, importantly, what is not! 
         including unified description of the writing of all the types into the PDF;
        2 types of named objects (storing in LaTeX only  the 
         PDF ref number, not the contents); the unnamed objects with no storage so they
         cannot be referenced (even though each as a PDF reference number, it gets lost!)
 4.  (maybe a sub sub of 3.?) Types of PDF object, and details of the contents syntax    
          for (some) types (there are others!). 

Thus we have these concepts: 
“named objects that get there PDF reference stored (two types)”
“other objects that do not get anything stored”
The important distinction is thus between whether an object can be referenced at all from LaTeX, not so much whether or how it is named. 
﻿

u-fischer:
@car222222  

> 3\. the unnamed objects with no storage so they
>     cannot be referenced (even though each as a PDF reference number, it gets lost!)

You can reference unnamed objects with `\pdf_object_ref_last:`.  They would be rather useless without such an option.

car222222:
Possibly it would be useful to give some guidance on when only a single "class" is needed; and on how to determine how large each one of a collection classes should be (and how to name them).

car222222:
@muzimuzhi Only the one just written, surely!  
And: There is no provision for storing this reference number. 

So: Yes, sure, this is a detail that needs to be added to the text.

car222222:
Also, if one does wish to store the reference, then why not just use some type of "named object", rather than these?

u-fischer:
> Also, if one does wish to store the reference, then why not just use some type of "named object", rather than these?

because you sometimes need the reference only for a short time and so can store it in a tmp-macro or simply use it directly in the next command.

car222222:
@u-fischer Yes, sure!  But I was just responding to @muzimuzhi asserting that this was possible. 

Also, even if it is only needed temporarily, you probably still need to store it.

car222222:
The text currently says simply this: 

Inserts the appropriate information to reference the last \meta{object} created. This is particularly useful for anonymous objects.

Without saying what "last created" means, or what an "anonymous object" is 
(better than "unnamed":-). 

This all needs better explanation of the possibilities, the functions involved, and the requirements, not to mention the timing essentials (exactly when the reference number is available).

car222222:
To put it another way: an extra sub sub sec is needed to describe these possibilities (or maybe provide functions for them?).
Thus:
"creating" an object;
possibly storing its PDF reference in a local variable;
writing in the PDF the contents of the (very recently) "last-created object" without using a local variable;
the same if using a local variable (almost the same as for a "named object');
etc.

u-fischer:
> @u-fischer Yes, sure! But I was just responding to @muzimuzhi asserting that this was possible.

Imho you were answering my comment. At least I could find nothing from @muzimuzhi which seems related. 

> Also, even if it is only needed temporarily, you probably still need to store it.

Sure, but that is up to the user/author. I'm using unnamed objects quite a lot, and sometimes I store in a command and sometimes I use directly:

~~~~
\pdf_object_unnamed_write:ne {dict}{\pdfdict_use:n {l_pdfmeta/outputintent} }
     \pdfmanagement_add:nne {Catalog}{OutputIntents}{\pdf_object_ref_last:}
~~~~

But beside this I do not think that a general discussion about PDF objects belong into this PR so I would suggest to move that to some other place.

car222222:
OK, but the general stuff is needed: 
with a reference from here. 
Whatever you think! And do not tell me what I am answering in a response written 5 mins before you joined in! 

Where else would the general introduction go?

car222222:
No problems with what you do.

I am only saying that maybe provisions for doing so, and a description of how to do this (especially the crucial importance of the timing) is needed.

car222222:
Are existence conditionals now needed for "indexed objects" also? -- similar to: 

\pdf_object_if_exist...

car222222:
Some other possible queries/facts that may be useful for “indexed objects” (approximately):

\pdf_if_object_class_exists…
(also for the general intarray classes, when/if we get there!)

\pdf_last_index_int <class> 
(may already be available with a lower-level command?)

muzimuzhi:
> Are existence conditionals now needed for "indexed objects" also? -- similar to:
> 
> \pdf_object_if_exist...

It was raised in previous review too, see https://github.com/latex3/latex3/pull/1524#discussion_r1543358819. The decision received there was, in my understanding, not until it's needed by real use cases (currently mainly the pdf tagging project).

I too am not quite understand how pdf object reference can be used, as all things about pdf objects are so specific to (La)TeX. One lazy solution could be to refer to docs of related primitives, for example the sec. 4.6 "PDF objects" in The pdfTEX user manual (pdfTeX 1.40.26, 2024-02-22). Although the manual also provides no concrete example.

To me this PR doesn't make docs of pdf object functions worse (well a weak argument) and in some extent more comments makes it harder to follow, so I agree with @u-fischer in https://github.com/latex3/latex3/pull/1524#issuecomment-2029593009, "I would suggest to move that to some other place."  As far as no new release will be shipped between merging this PR and adding planned doc improvements, IMO it is ready to go.

car222222:
@muzimuzhi No problem in providing extra information elsewhere, so long as the documentation here is rejigged to fit, and there is a reference. 

The description in the PDFTeX Manual is indeed pretty much useless (I could not follow it at all at first) so some good description of these basics is badly needed.  But if l3pdf is to cover even a few aspects of PDF objects and their specification via TeX, then it needs clear linking to such basic descriptions.  Also, to somewhere that explains how the relevant parts of PDF work. 

These should be written now, wherever they may finally end up; they could easily have a temporary home here! Why not!

u-fischer:
> Are existence conditionals now needed for "indexed objects" also? -- similar to:
> 
> \pdf_object_if_exist...

I'm not sure. I'm using `\pdf_object_if_exist` in some places, but the indexed variant is meant for larger number of objects and should improve efficieny. It feels wrong to test existence by default. Instead one should ensure that there are no holes in the array. 

> \pdf_if_object_class_exists…

no that is not needed. We have currently one class for structure, and even if  I add another I will know that the classes exist. 

> \pdf_last_index_int

Not sure, perhaps some use case will come up, but currently it is imho not needed.

car222222:
It is not "ready to go" whilst it uses misleading names and probably also whilst it omits coverage (maybe only as examples) of features that are already in use (such as temporary names, and creation+contents-writing of never-named, and hence ultimately inaccessible).  Why the hurry? 

And I have a few more suggestions in the pipeline! :-)

car222222:
Not forgetting (as already explained) the structure is wrong in distinguishing, contrasting even, the "simple naming" of objects with using classes of intarrays to name them as co-equal subsections.  And there is no mention of using temporary names!

car222222:
Is it not better to provide reasonable things now, rather than always limiting to what one person is currently doing?  

Or am I entirely missing the point of the l3* packages! 

In other words, is it relevant to the real world what is currently being done in one very small corner of that world!

u-fischer:
@josephwright 

* Currently there is difference if you reference an unknown, indexed structure inside the range or outside the range. Imho it would make sense here to a test and give back the 0  structure always:

~~~~
\documentclass{article}
\begin{document}
\ExplSyntaxOn 
\pdf_object_new_indexed:nn{blub}{1} 

\pdf_object_ref_indexed:nn{blub}{1} ~% 1 0 R
\pdf_object_ref_indexed:nn{blub}{2000} ~% 0 0 R  
%\pdf_object_ref_indexed:nn{blub}{20000} % error
\par
\cs_set:Npn \__pdf_object_ref_indexed:Nn #1#2
  {
    \exp_args:Ne \__pdf_backend_object_ref:n
      { \intarray_if_exist:NTF #1{\intarray_item:Nn #1 {#2}}{0} }
  }
\par
\pdf_object_ref_indexed:nn{blub}{1} ~% 1 0 R
\pdf_object_ref_indexed:nn{blub}{2000} ~% 0 0 R  
\pdf_object_ref_indexed:nn{blub}{20000} % 0 0 R
\end{document}
~~~~

* For the structure destinations I need access to the actual object number of a structure (without the 0 R) so I would suggest to add something like the following. This could then serve also as an existence test: if the result is zero, the object doesn't exist.

~~~~
\documentclass{article}
\begin{document}
\ExplSyntaxOn 
\pdf_object_new_indexed:nn{blub}{1} 

\cs_new:Npn \__pdf_object_refnum_indexed:Nn #1#2
  {
     \intarray_if_exist:NTF #1{\intarray_item:Nn #1 {#2}}{0}
  }

\cs_new:Npn \pdf_object_refnum_indexed:nn #1#2
  {
      \__pdf_object_indexed_split:nnN {#1} {#2}
        \__pdf_object_refnum_indexed:Nn
  }
\par 
\pdf_object_refnum_indexed:nn{blub}{1} ~% 1 
\pdf_object_refnum_indexed:nn{blub}{2000} ~% 0 
\pdf_object_refnum_indexed:nn{blub}{20000} % 0 

\ExplSyntaxOff
\end{document}
~~~~

josephwright:
>     * For the structure destinations I need access to the actual object number of a structure (without the 0 R) so I would suggest to add something like the following. This could then serve also as an existence test: if the result is zero, the object doesn't exist.

How are you doing that at present? The `0 R` bit is backend-dependent so it's not obvious that we can have a general function that doesn't add it.

josephwright:
@car222222  I'm not sure which of your comments are specific to this PR: please could you clarify? If there are documentation issues, I'm happy to address. If we need a different interface approach to adding `_indexed`, again happy to change but need a suggestion or two.

u-fischer:
> > ```
> > * For the structure destinations I need access to the actual object number of a structure (without the 0 R) so I would suggest to add something like the following. This could then serve also as an existence test: if the result is zero, the object doesn't exist.
> > ```
> 
> @u-fischer How are you doing that at present? The `0 R` bit is backend-dependent so it's not obvious that we can have a general function that doesn't add it.

The exact method is backend dependent (the relevant code is in l3backend-testphase). With xelatex one can simply use the normal object reference `\pdf_object_ref:n{structure)`, with pdflatex + lualatex one need the object number and currently I use the internal int ` \int_use:c{ c_@@_object_ <structure> _int }` but that naturally will not work if the structure object are switched to the indexed version. 

But thinking about it, a public command is probably not good, as with xelatex it is obviously not the actual object number but something else. Instead some internal command that I can use in l3backend-testphase would be good.

muzimuzhi:
Maybe `l3candidates.dtx` is needed again (5d5f252e (Remove l3candidates, 2024-01-10)).

josephwright:
> Maybe `l3candidates.dtx` is needed again ([5d5f252](https://github.com/latex3/latex3/commit/5d5f252e9da70d6cc32ff4eea22b1de95e0c9152) (Remove l3candidates, 2024-01-10)).

Not a good idea: the reason for removing it was we need to make sure that stuff going into `l3kernbel` has been discussed and agreed. So PRs with focussed discussion are the way to go for relatively small changes, and experimental code aht people can load explicitly for larger stuff.

u-fischer:
> Also, as we don't have a 'non-indexed' approach that fits here, wouldn't this be separate from the PR at hand?

Well the current method is a bit ugly as it makes use of the internal integer name and I thought that if we improve the structure code then would be good to have  a clean access to the number too ;-)

josephwright:
@u-fischer See the latest addition - I'll look at the Lua stuff next

josephwright:
> I'm wondering if the indexed objects could/should be reserved in blocks in advance, e.g. 1000. That would mean that the object numbers of the next 1000 structures are already known and other objects, e.g. from filespec dictionaries or OBJR from links wouldn't shift that numbers around. Object numbers are sometimes stored in labels in the aux, e.g. for /Ref keys and if they stay more stable it could reduce the number of compilations needed.

Does this need a dedicated function? Couldn't you do it using
```latex
\int_step_inline:nnn { <start> } { <end> }
  {
    \pdf_new_object_indexed:nn { <class> } {#1}
  }
```

u-fischer:
> Does this need a dedicated function? Couldn't you do it using

yes, I was only wondering where such reservation should be done. But I think this is better done in the tagging code if needed, it feels too special and so nothing for this PR.

l3kernel/l3pdf.dtx
```diff
@@ -473,6 +481,16 @@
           { 0 }
       }
   }
+\cs_new:Npn \__kernel_pdf_object_id_indexed:nn #1#2
+  {
+    \@@_object_indexed_split:nnN {#1} {#2}
+      \@@_object_ref_id:Nn
```
u-fischer, line 487: shouldn't that be `\@@_object_id_indexed:Nn` ?

l3kernel/l3pdf.dtx
```diff
@@ -314,52 +363,248 @@
...
+  @@_objects[name] = n
+end, 'global')
+
+local function object_id(name)
+  return @@_objects[name] or 0
+end
+
+luacmd('@@_object_retrieve:n', function()
+  local name = scan_string()
+  return object_id(name)
```
zauguin, line 432: You probably want something like
```suggestion
  tex.cprint(12, tostring(object_id(name)))
```
here, otherwise the return value is ignored.

josephwright: Adjusted (I still don't really get all of the `print` variants Hans provides!)

l3kernel/l3kernel-functions.dtx
```diff
@@ -381,6 +381,17 @@
...
 % \end{function}
 %
+% \begin{function}[EXP]{\__kernel_pdf_object_id:n, \__kernel_pdf_object_id_indexed:nn}
+%   \begin{syntax}
+%     \cs{__kernel_pdf_object_id:n} \Arg{object}
+%     \cs{__kernel_pdf_object_id_indexed:nn} \Arg{class} \Arg{number}
+%   \end{syntax}
+%   Expands to the ID the \meta{object} (or object of \meta{number} within
+%   the \meta{class}), in for example page resource allocation. Depending on
+%   the backend, teh result may be the same as
```
muzimuzhi, line 391:
```suggestion
%   Expands to the ID of \meta{object} (or object of \meta{number} within
%   the \meta{class}), in for example page resource allocation. Depending on
%   the backend, the result may be the same as
```

muzimuzhi (changes requested):
doc typo

l3kernel/l3pdf.dtx
```diff
@@ -314,43 +367,106 @@
...
-% \begin{macro}
-%   {
-%     \pdf_object_unnamed_write:nn, \pdf_object_unnamed_write:ne,
-%     \pdf_object_unnamed_write:nx
-%   }
-% \begin{macro}{\pdf_object_ref_last:}
-% \begin{macro}[pTF]{\pdf_object_if_exist:n}
+% \begin{macro}{\__kernel_pdf_object_id:n}
+% \begin{macro}{ltx.pdf.object_id}
+% \begin{macro}{\@@_object_record:Nn}
```
muzimuzhi, line 383:
```suggestion
% \begin{macro}{\@@_object_record:nN}
```

muzimuzhi (changes requested):
Another typo.

car222222:
Re: "Document that non-existent objects return 0"
and document that 
"this works well since no PDF indirect object ever has reference 0"

car222222:
A query on the implementation (nothing vital):
is it "good/essential" that each "named object" has its own count register? 
Is this for time/space efficiency? 

Would it be better (or not much worse) to store each such PDF ref number in a tl variable? (\c_@@_object_ #1 _tl)
Or does some backend demand the use of a register?

PR #1534, opened 2024-04-15
Fix various misspelling

muzimuzhi:
Found with `codespell -L 'alph,co-ordinate,co-ordinates,filetest,inteval,openin' -S './texmf,*.eps,*.tlg'`.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -75,7 +75,7 @@ jobs:
   # GitHub automatically informs the initiator of any action about the result, but
   # we additionally want to keep the latex-commits mailing list informed about
   # test failures.
-  notifiy:
+  notify:
     name: Send notifications
     runs-on: ubuntu-20.04
     # Run after the `l3build` job in order to be able to react to it's output.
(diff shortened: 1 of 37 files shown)


josephwright:
Please exclude `xpackages`: this is all frozen stuff we really don't want to change

car222222:
Good stuff! 
Useful automation finally hits the documentation:-).

josephwright:
Nice tool - have just set it up

car222222:
What horrors did you find in siunitx:-)

car222222:
So little, congratulations!
I like "hightlight", you should have kept that:-).

muzimuzhi:
> Useful automation finally hits the documentation:-).

Got to know `codespell` for the first time in https://github.com/CTeX-org/zhlipsum/pull/8.

muzimuzhi:
Need to reduce false positive matches.

josephwright:
> Need to reduce false positive matches.

I was thinking for changed lines -we would then have an exclusions list that went with the repo as a whole, which would be added to be commits with 'odd' spellings

josephwright:
There is a Codespell action: only works on PRs at the moment - may be worth trying out

PR #1537, opened 2024-04-23
Register prefix `istqb`

Witiko:
This PR registers the prefix `istqb`. Furthermore, this PR also updates the registrar of prefixes `lt3luabridge` and `markdown` and sorts out-of-order rows in `l3prefixes.csv`.


josephwright:
Er, you seem to have over-done the sorting :) Capital letters are deliberately separated out from all lower-case as they come first in ASCII.

PR #1542, opened 2024-04-28
Define `\tl_set_from_cs:NN` and `\tl_gset_from_cs:NN`

Witiko:
This PR defines functions `\tl_set_from_cs:NN` and `\tl_gset_from_cs:NN` that assign the replacement text of a control sequence to a tl var. Compared to `\cs_replacement_spec:N`, these functions preserve all category codes. These functions only work for control sequences without delimited arguments.

The code has been developed [on TeX StackExchange][1] with contributions from @eg9, @gucci-on-fleek, and @Skillmon. It is currently [used][4] in the Markdown Package for TeX to implement appending to control sequences without the need to maintain copies of their current replacement text in separate tl vars.

Here is a demonstration from TeX StackExchange of how this function can be used:

> ``` tex
> \cs_new:Npn
>   \greet
>   #1#2
>   { Hello,~#1!~How~is~#2~doing? }
> \tl_set_from_cs:NN
>   \l_greet_tl
>   \greet
> \tl_put_right:Nn
>   \l_greet_tl
>   { ~Have~a~great~#3! }
> \cs_generate_from_arg_count:NNnV
>   \greet
>   \cs_set:Npn
>   { 3 }
>   \l_greet_tl
> ```
> 
> Now, I can write `\greet { world } { favorite~species~(humans,~presumably) } { evolutionary~leap }` and receive the following expansion:
> 
> > Hello, world! How is your favorite species (humans, presumably) doing? Have a great evolutionary leap!

 [1]: https://tex.stackexchange.com/q/716362/70941
 [2]: https://github.com/Witiko/markdown/pull/435
 [3]: https://github.com/Witiko/markdown/issues/437
 [4]: https://github.com/Witiko/markdown/pull/438

This PR is currently just a draft meant to kick off a discussion on the following topics:

1. Do we want to include control sequence patching in core expl3?
2. Is l3tl the proper place for these functions?
3. Are there more appropriate names for these functions?

After the discussion, if the answer to the first question is a yes, I will add regression tests and mark this PR as ready for a review.

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -215,6 +215,20 @@
 %   \meta{tl~var}.
 % \end{function}
 %
+% \begin{function}[added = 2024-04-28]
+%   {
+%     \tl_set_from_cs:NN, \tl_set_from_cs:cN, \tl_set_from_cs:Nc, \tl_set_from_cs:cc,
+%     \tl_gset_from_cs:NN, \tl_gset_from_cs:cN, \tl_gset_from_cs:Nc, \tl_gset_from_cs:cc
+%   }
+%   \begin{syntax}
+%     \cs{tl_set_from_cs:NN} \meta{tl var} \meta{control sequence}
+%   \end{syntax}
+%   This function assigns the replacement text of a \meta{control sequence}
+%   to \meta{tl var}. Compared to \cs{cs_replacement_spec:N}, this function
+%   preserves all category codes. This function only works for
+%   \meta{control sequence}s without delimited arguments.
+% \end{function}
+%
 % \section{Token list conditionals}
 %
 % \begin{function}[EXP,pTF, updated = 2019-09-04]
@@ -1717,6 +1731,96 @@
 \cs_generate_variant:Nn \tl_gput_right:No { c }
 \cs_generate_variant:Nn \tl_gput_right:Nn { Nx, cx }
 %    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\@@_from_cs:NNNn}
+% \begin{macro}{\@@_from_cs:NNN}
+% \begin{macro}
+%   {
+%     \tl_set_from_cs:NN, \tl_set_from_cs:cN, \tl_set_from_cs:Nc,
+%     \tl_set_from_cs:cc
+%   }
+% \begin{macro}
+%   {
+%     \tl_gset_from_cs:NN, \tl_gset_from_cs:cN, \tl_gset_from_cs:Nc,
+%     \tl_gset_from_cs:cc
+%   }
+%    \begin{macrocode}
+\cs_new_protected:Nn
+  \@@_from_cs:NNNn
+  {
+    \tl_set:Nn
+      \l_@@_internal_a_tl
+      { #2 }
+    \int_step_inline:nn
+      { #4 }
+      {
+        \exp_args:NNc
+          \tl_put_right:Nn
+          \l_@@_internal_a_tl
+          { @@_from_cs_parameter_ ##1 }
+      }
+    \exp_args:NNV
+      \tl_set:No
+      \l_@@_internal_b_tl
+      \l_@@_internal_a_tl
+    \regex_replace_all:nnN
+      { \cP. }
+      { \0\0 }
+      \l_@@_internal_b_tl
+    \int_step_inline:nn
+      { #4 }
+      {
+        \regex_replace_all:nnN
+          { \c { @@_from_cs_parameter_ ##1 } }
+          { \cP\# ##1 }
+          \l_@@_internal_b_tl
+      }
+    #3
+      #1
+      \l_@@_internal_b_tl
+  }
+\cs_new_protected:Nn
+  \@@_from_cs:NNN
+  {
+    \tl_set:Ne
+      \l_@@_internal_a_tl
+      { \cs_parameter_spec:N #2 }
+    \@@_from_cs:NNNf
+      #1
+      #2
+      #3
+      { \tl_count:N \l_@@_internal_a_tl / 2 }
+  }
+\cs_generate_variant:Nn
+  \@@_from_cs:NNNn
+  { NNNf }
+\cs_new_protected:Nn
+  \tl_set_from_cs:NN
+  {
+    \@@_from_cs:NNN
+      #1
+      #2
+      \tl_set:NV
+  }
+\cs_new_protected:Nn
+  \tl_gset_from_cs:NN
+  {
+    \@@_from_cs:NNN
+      #1
+      #2
+      \tl_gset:NV
+  }
+\cs_generate_variant:Nn
+  \tl_set_from_cs:NN
+  { cN, Nc, cc }
+\cs_generate_variant:Nn
+  \tl_gset_from_cs:NN
+  { cN, Nc, cc }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
 % \end{macro}
 % \end{macro}
 %
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -1747,8 +1747,8 @@
 %     \tl_gset_from_cs:cc
 %   }
 %    \begin{macrocode}
-\cs_new_protected:Nn
-  \@@_from_cs:NNNn
+\cs_new_protected:Npn
+  \@@_from_cs:NNNn #1#2#3#4
   {
     \tl_set:Nn
       \l_@@_internal_a_tl
@@ -1781,8 +1781,8 @@
       #1
       \l_@@_internal_b_tl
   }
-\cs_new_protected:Nn
-  \@@_from_cs:NNN
+\cs_new_protected:Npn
+  \@@_from_cs:NNN #1#2#3
   {
     \tl_set:Ne
       \l_@@_internal_a_tl
@@ -1796,16 +1796,16 @@
 \cs_generate_variant:Nn
   \@@_from_cs:NNNn
   { NNNf }
-\cs_new_protected:Nn
-  \tl_set_from_cs:NN
+\cs_new_protected:Npn
+  \tl_set_from_cs:NN #1#2
   {
     \@@_from_cs:NNN
       #1
       #2
       \tl_set:NV
   }
-\cs_new_protected:Nn
-  \tl_gset_from_cs:NN
+\cs_new_protected:Npn
+  \tl_gset_from_cs:NN #1#2
   {
     \@@_from_cs:NNN
       #1


gucci-on-fleek:
> 2. Is l3tl the proper place for these functions?

As someone _completely uninvolved_ with expl3 development, my recommendation would be to:

1. Soft-deprecate `\cs_parameter_spec:N` and `\cs_replacement_spec:N`; that is, recommend against its use, but still keep it around for a bit. (I counted 52 files in `$TEXMFDIST` that use them, so removing them would be needlessly disruptive).

2. Rename the current `\cs_parameter_spec:N` and `\cs_replacement_spec:N` to `\cs_parameter_spec_to_str:N` and `\cs_replacement_spec_to_str:N`. This matches the naming conventions of `\cs_to_str:N`, `\token_to_str:N`, `\tl_to_str:N`, and `\bool_to_str:N`, and makes it clear that these functions ignore any catcodes.

3. Name the new function (the one in this PR) `\cs_replacement_spec_to_tl:NN`. The first half of the name matches the current function `\cs_replacement_spec:N`, and the second half matches the functions `\fp_to_tl:N` and `\fparray_item_to_tl:Nn`. The new names also make the differences between `\cs_replacement_spec_to_str:N` and `\cs_replacement_spec_to_tl:NN` fairly self-explanatory.

> 1. Do we want to include control sequence patching in core expl3?

I see this less as “patching” and more as “introspection”, or perhaps as “a more robust `\cs_replacement_spec:N`”. The biggest issue that I see is that l3regex isn't used anywhere in the kernel outside of l3regex itself, and I'm not sure if this use case is strong enough to justify its first internal use.

I would lean towards its inclusion though, since the function seems to be both generally useful and fairly tricky to implement robustly.

josephwright:
>     1. Soft-deprecate `\cs_parameter_spec:N` and `\cs_replacement_spec:N`; that is, recommend against its use, but still keep it around for a bit. (I counted 52 files in `$TEXMFDIST` that use them, so removing them would be needlessly disruptive).

As a policy position, nothing will be removed from `l3kernel` in future - we have reached the point where that doesn't make sense.

Witiko:
> As a policy position, nothing will be removed from `l3kernel` in future - we have reached the point where that doesn't make sense.

Soft deprecation does not need to mean the future removal of an API. For example, [Python][1] documentation discourages the use of soft-deprecated APIs and lists preferred alternative APIs. We may also make soft-deprecated APIs less prominent in the documentation by using visual cues such as gray text or by moving them to a separate section. However, they would still be implemented and regression-tested.

 [1]: https://docs.python.org/es/dev/glossary.html#term-soft-deprecated

davidcarlisle:
I'm not sure the use case here.

You can  only know that the command has undelimited arguments if it's an expl3 names csname with `:nnn`  (unlike the `\greet` example), or if you know its definition in advance.

In either case you also know how many arguments it has so can add to it without all the complicated introspection involved in the more general case.

In many cases an application of a cmd hook with arguments may be enough?

Also it's not clear if the use case is really to generate a tl with just the replacement text or if that is just a step in the intended final use case of modifying the replacement text of a csname?

As a general rule "patching" has so far intentionally been not added to expl3 as unrestricted code changes is sort of an anti-pattern.

In an ideal world code that potentially needed different behaviour in different contexts would have a hook or a socket or some other structured access to change points and not require patching.

Witiko:
> I'm not sure the use case here.

My use case is the appending operator in the Markdown package, which patches a user command by appending to the end of its replacement text. The uses of the appending operator are the incremental definition of token renderers, [especially those that handle HTML attributes][1]. I will discuss more on this topic in my TUG 2024 talk.

 [1]: https://github.com/latex3/latex3/assets/603082/c54fc980-6bf6-40d2-938a-955530450936

> You can only know that the command has undelimited arguments if it's an expl3 names csname with `:nnn` (unlike the `\greet` example), or if you know its definition in advance.
>
> In either case you also know how many arguments it has so can add to it without all the complicated introspection involved in the more general case.

In the Markdown package, we control both the prefix spec and the parameter spec of control sequences, whereas the users control the replacement spec. Therefore, we know that the control sequences have undelimited parameters and also the number of parameters. As you mention, this allows us to pass the number of parameters directly to the functions, [which we do in the Markdown package][2] to keep the function simple. This corresponds to the private function `\@@_from_cs:NNNn` from this PR, since keeping the interface simple seemed preferable to premature speed optimizations.

> Also it's not clear if the use case is really to generate a tl with just the replacement text or if that is just a step in the intended final use case of modifying the replacement text of a csname?
> 
> As a general rule "patching" has so far intentionally been not added to expl3 as unrestricted code changes is sort of an anti-pattern.

As I explained above, my use is patching, although not necessarily in the usual sense of the word, where one package patches the commands of another. Rather, it allows the programmer to treat their own control sequences as mutable.

> In an ideal world code that potentially needed different behaviour in different contexts would have a hook or a socket or some other structured access to change points and not require patching.

This seems true for public functions, whose definition (or at least the expected behavior) should likely not change during the runtime of a program. However, even such functions may be easier to define incrementally.

Furthermore, even if the usefulness of the patching and the incremental definitions of functions in the LaTeX3 kernel is unclear, the functions seem general enough to enable other uses such as introspection. However, it's possible that `\cs_replacement_spec:N` is sufficient for that purpose.

 [2]: https://github.com/Witiko/markdown/blob/8a9ef823558604ef87c3890ea1db0830bc872762/markdown.dtx#L20207-L20266

josephwright:
> > [...N]othing will be removed from `l3kernel` in future [–] we have reached the point where that doesn't make sense.
> 
> That seems for the best. However, @gucci-on-fleek mentioned _soft deprecation_, which does not need to mean the future removal of an API. For example, [Python](https://docs.python.org/es/dev/glossary.html#term-soft-deprecated) documentation discourages the use of soft-deprecated APIs and lists preferred alternative APIs. We may also make soft-deprecated APIs less prominent in the documentation by using visual cues such as lesser opacity or by moving them to a separate section. However, they would remain implemented and regression-tested.

The approach in `l3kernel` is that only currently-recommended functions are documented: deprecated material is moved and the documentation is removed, but the code will never be dropped. That's been the case for a couple of years now. (Deprecated functions are listed in `l3obsolete.txt`.)

Witiko:
@josephwright: That makes a lot of sense, I learnt something new. To be clear, I don't necessarily agree with @gucci-on-fleek that `\cs_replacement_spec:N` should be deprecated, since it may be sufficient for cases, where it is useful in the LaTeX kernel.

Witiko:
Let's close this PR as not-planned then?

josephwright:
@Witiko I think we should see what @FrankMittelbach, @car222222, @eg9, etc., think

Witiko:
To be clear, I only mentioned @eg9 in the commit message to acknowledge their contribution. I am not sure whether they approve of this PR or how active they are on GitHub.

josephwright:
@Witiko Enrico looks at 'bigger picture' API questions for the team as he does a lot of `expl3` 'promotion' :)

car222222:
I am not here commenting on these ideas as a "method of patching”.
Thus I ﻿am unsure how pertinent these comments are to the original post.

These ideas do seem to be worth considering:
@gucci-on-fleek 's suggestion in his #2. looks sensible.
And also, therefore, his #3 also (i.e., the concept of this new function, but not necessarily his suggested name) looks both sensible and useful. 

(I had not previously understood that these two functions currently produce strings!).

Witiko:
In commit e00ac608c from this draft PR, I fixed the CI. Apparently, we can't use the function `\cs_new_protected:Nn` in `l3tl.dtx` and have to use the lower-level function `\cs_new_protected:Npn`. Who knew?!

From the discussion here, I get the sense that the new functions `\tl_set_from_cs:NN` and `\tl_gset_from_cs:NN` from this PR don't seem as strong contenders for becoming part of the standard library by at least @davidcarlisle and perhaps others in the LaTeX team who haven't given their opinion. Suggestion (3) from @gucci-on-fleek to drop `\tl_gset_from_cs:NN` and rename `\tl_set_from_cs:NN` to `\cs_replacement_spec_to_tl:NN` seems more in line with the rest of the current interfaces and therefore perhaps more acceptable than my original proposal.

I am happy to work with you to get this PR to a point where it is good to be merged.

muzimuzhi:
> Apparently, we can't use the function `\cs_new_protected:Nn` in `l3tl.dtx` and have to use the lower-level function `\cs_new_protected:Npn` instead. Who knew?! 🤔

That's tricky.

In [`l3.ins`](https://github.com/latex3/latex3/blob/727c6876ff04864577b7c3d0f7cf92507c2c9825/l3kernel/l3.ins#L64-L73), `l3names.dtx` is loaded before `l3tl.dtx`, the latter is then loaded before `l3prg.dtx`.

But implementation of `\cs_new_protected:Nn` and friends (in `l3basics.dtx`) uses `\bool_if:NTF` which is defined in `l3prg.dtx`, hence is not defined when `l3tl.dtx` is loaded.
https://github.com/latex3/latex3/blob/727c6876ff04864577b7c3d0f7cf92507c2c9825/l3kernel/l3basics.dtx#L3105-L3107

This explains the error in the [log of CI](https://github.com/latex3/latex3/actions/runs/8870781998/job/24352962129#step:5:56) triggered by https://github.com/latex3/latex3/pull/1542/commits/cd2d1212b49cdffab6eddab56335f3c82ddff25a
```
! Undefined control sequence.
\__cs_generate_from_signature:nnNNNn ...ol_if:NTF 
                                                  #3{\cs_set_nopar:Npx \__cs...
l.3716   }
```

For speed considerations, the implementation of `l3kernel` always prefers `:Npn` than `:Nn` style `\cs_new...` functions, I think.

FrankMittelbach:
> For speed considerations, the implementation of `l3kernel` always prefers `:Npn` than `:Nn` style `\cs_new...` functions, I think.

yes!

 In my opinion the whole `:NN` was a misguided idea of an abstraction. Beside being slow it is not even better readable.

Witiko:
> Beside being slow it is not even better readable.

It gets worse with many parameters: `\cs:new:Npn \some_function:NNNNNNN #1#2#3#4#5#6#7 { ... }`. The `l3kernel` code can likely be trusted to keep the plain TeX parameter spec consistent with the expl3 argument specifiers. However, it seems like a needless footgun in library and user code, where the ease of writing and maintainability is usually more important than speed.

FrankMittelbach:
> > Beside being slow it is not even better readable.
> 
> The readability seems improved with a large number of parameters. Compare `\cs_new:Nn \some_function:NNNNNNNNN { ... }` and `\cs_new:Npn \some_function:NNNNNNNNN #1#2#3#4#5#6#7#8#9 { ... }`.
> 

that's not my understanding of readability. In the second form  I immediately see that this has 9 parameters in the first have to painfully count. Now the counting has to be made once for sure (when you write the definition) but for reading code I think the other form is better. but it is a borderline example in any case

> The `l3kernel` code can likely be trusted to keep the plain TeX parameter spec consistent with the expl3 argument specifier. However, it can be a needless footgun in library and user code, where the ease of writing and maintainability often seem more important than the runtime speed.

again, not convinced sorry. I know that @wspr thought (or thinks) differently about that in line with you, so I'm not claiming that my position is the right one, but I honestly don't see footguns lurking here --- to me the other form has mainly downsides.

Witiko:
Sorry for the edits!

FrankMittelbach:
> This can be a needless footgun in library and user code, where the ease of writing and maintainability often seem more important than the runtime speed. Understandably, this is less the case for the `l3kernel` code, where speed is a priority. Perhaps the ideal language for writing `l3kernel` and the ideal language for writing library and user code are two slighly different languages.

it is actually the other way around: how long it would take for the kernel to compile into a fmt file doesn't matter at all where it matters is in code that is processing such constructs at runtime. But once more, I don't see your argument concerning maintainibility: The given spec has to match `#1#2#3...` yes (a tiny and not complicated extra step) but it also has to match the individual `#1`... inside the definition. So if you make a mistake in the spec then your code blows up in some form or another.

Agreed, you can't do `:nnn #1#2` by mistake while now you can, but I haven't seen many invalids with gunshots in their feet due to that. On the other hand you can also not do this deliberately, which is often a noticeable speed improvement in itself and that is something worth both in kernel as well as package code.

Witiko:
> The given spec has to match `#1#2#3...` yes (a tiny and not complicated extra step) but it also has to match the individual `#1`... inside the definition. So if you make a mistake in the spec then your code blows up in some form or another.

The issue for me is the violation of the DRY principle, but I agree that I would rather keep the `#1#2#3...` rather than the `:NNN...`!

Witiko:
> it is actually the other way around: how long it would take for the kernel to compile into a fmt file doesn't matter at all where it matters is in code that is processing such constructs at runtime.

That is true, although it still matters when loading `expl3-generic.tex` in other formats.

Witiko:
Now that I have successfully derailed my own PR with unrelated discussion, I feel inspired to create a LaTeX3 library called `lt3fancycs` that will implement fancy control sequences with Python-style key–value parameters. Importantly, they will be pig slow both at definition and at call time. However, they will be quite beautiful.

FrankMittelbach:
> > 
> 
> The footgun for me is the violation of the DRY principle due to the duplication between the parameter spec and the argument specification. If the programmer makes a mistake, the code will _eventually_ blow but not every library has high-coverage continuous tests to catch it early like `l3kernel` does.

yes, that principle :-) but remember we aren't dealing with a compiled language but with an macro interpreter and there we have to stay reasonably close to what the interpreter does out of the box, which is why a lot of things like the `__` conventions, `\l_`, `\g_`, as well as the arg spec are for human consumption not enforced syntax (unless in checking mode) which, by the way, one could check there if anybody cares to write the necessary check routines.

So yes it is not DRY due to the need to offer the real parameters to TeX in an efficient way without overhead but that's the way the cookie crumbles in my opinion.

Witiko:
> which, by the way, one could check there if anybody cares to write the necessary check routines

I am currently working on a static expl3 linter.

[The last update][1] is from May but I am planning to post the next update by the end of this week.

> So yes it is not DRY due to the need to offer the real parameters to TeX [...]

Yes, although as we agreed, the real parameters are perhaps not the issue, the `:NN` abstraction is.

> [...] in an efficient way without overhead but that's the way the cookie crumbles in my opinion.

This is definitely true for `l3kernel` but not all library and user code needs to be efficient. The time-to-market can be more important than whether a document compiles in 10 or 15 seconds. There are definitely different considerations and tradeoffs for `l3kernel` code and library and user code.

 [1]: https://x.com/tvitecnik/status/1787148258176881024

Witiko:
> Experience with TeX/`expl3` generally suggests that issues in matching arg. specs are low down the bug cause list.

I was making a more general observation that the language seems to serve a double duty as a base programming layer for the LaTeX team and also a high-level programming language for package authors. Both groups have somewhat different goals and might make somewhat different design choices for expl3. For example, this draft PR proposes a new interface that may be useful to package authors but seems useless to `l3kernel`.

As an aside, I am hoping that the expl3 checker can provide a more data-driven overview of the common causes of bugs in expl3 code. When the tool is more mature, it would seem useful to run it on all packages in TeX Live and see what it says.

wspr:
> again, not convinced sorry. I know that @wspr thought (or thinks) differently about that in line with you, so I'm not claiming that my position is the right one, but I honestly don't see footguns lurking here --- to me the other form has mainly downsides.

I do wonder if I argued too vehemently to keep it :) Personally I do still find it a help when I am writing expl3 code, and I guess we are stuck with it now anyway, but I've probably softened with age about whether I think it's a "good" feature of the language.

Skillmon:
> Now that I have successfully derailed my own PR with unrelated discussion, I feel inspired to create a LaTeX3 library called `lt3pigcs` that will implement fancy control sequences with Python-style key–value parameters. Importantly, they will be _pig_ slow both at definition and at call time. However, they will be quite beautiful.

I played with that thought multiple times as well. But ultimately I always think that it isn't worth the hassle. Also, `expkv-cs` comes close so why bother, and I have a PR to get something similar into `ltcmd` (splitting an argument with key=value syntax in separate arguments).

Skillmon:
Oh, and only because it seems to be en vogue to state ones stance towards `:Nn` vs `:Npn`: I'm absolutely and totally in favour of `:Npn` and against `:Nn`, mostly because my code curries wherever possible (so I'm going to upset your linter).

Witiko:
> Also, expkv-cs comes close so why bother

I appreciate that you are doing my research of related work for me. I will acknowledge you in the upcoming pigcs article.

> I'm going to upset your linter

Hopefully not. My linter will be like me – happy to ignore things it does not understand.

Witiko:
> users increasingly complain about page-per-second rather than seconds-per-page rates, and that is affected by whatever code is being used.

I wonder if it would make sense to turn some of the slower expl3 functions into engine primitives? This should likely be informed by profiling a large set of diverse documents, which might include e.g. all arXiv preprints, to see where most time is spent and how much potential speed-up there is. I imagine that sequences, property lists, and key–values might benefit the most from such acceleration.

josephwright:
@Witiko Quite apart from the fact that this is not easy, adding primitives has a lead time of at least 5 years ...

Witiko:
> @Witiko Quite apart from the fact that this is not easy

The profiling, the implementation, convincing the engine developers and maintainers, or all of the above?

> adding primitives has a lead time of at least 5 years ...

All the more reason to start now. Of course, the functions would still be available in engines where the primitives would be unavailable, they would just be as slow as they are now. In LuaTeX, acceleration might be possible without altering the engine.

josephwright:
> > @Witiko Quite apart from the fact that this is not easy
> 
> The profiling, the implementation, convincing the engine developers and maintainers, or all of the above?

Have you ever written any WEB code ;)

More seriously, 'slow' is really quite a tricky thing - what's slow is not any specific function per se but some combinations. But then are those the most used things. For example, the floating point code is definitely slow, but it's not used in tight loops in the main.

gucci-on-fleek:
@Witiko

> > users increasingly complain about page-per-second rather than seconds-per-page rates, and that is affected by whatever code is being used.
> 
> I wonder if it would make sense to turn some of the slower expl3 functions into engine primitives?

This is fairly simple for LuaTeX—the easiest would be to just write it in Lua, but adding something to the engine in C is fairly simple too. pdfTeX and XeTeX are a bit harder though—they're written in WEB which not very many people are familiar with, and their codebases are rather convoluted. Depending on exactly what you're adding, pdfTeX and XeTeX could use the same code (change file), but if you need different code for each, then this gets tricky since XeTeX is mostly unmaintained and there's really no one active who actually understands it.

For LuaTeX though, if you can think of something specific, I might have time to take a look and maybe add it.

Witiko:
> Have you ever written any WEB code ;)

Not personally, but we go over [examples of WEB code][1] in our TeX course at the Masaryk University in Brno. Furthermore, in his 2019 Q&A sessions in Brno, Donald Knuth mentioned that his favorite programming language is CWEB, so it can't be all bad!

 [1]: https://www.fi.muni.cz/~xruzick7/pb029-elektronicka-priprava-dokumentu/literarni-programovani/

> More seriously, 'slow' is really quite a tricky thing - what's slow is not any specific function per se but some combinations. But then are those the most used things. For example, the floating point code is definitely slow, but it's not used in tight loops in the main.

That's why I suggested profiling the code to see where the most time is spent in an average document. I am happy to do this initial research with my students. However, it might just take a year, since I have my hands full with other projects at the moment, as I am sure you are. Therefore, I am probing how actionable and interesting you think such research would be for the LaTeX team.

josephwright:
> That's why I suggested profiling the code to see where the most time is spent in an average LaTeX document. I am happy to do this initial research with my students. However, it might just take a year, since I have my hands full with other projects, as I am sure you have. Therefore, I am probing how actionable and interesting such research would be for the LaTeX team.

Things that are relatively slow tend to link to TeX's basic model for variables: we have only macros and registers plus a few oddities (fontdimens for example),  and that's not really something that's particularly tractable to solve. In any case, what we don't see is lots of slow stuff caused by `expl3` in general - @FrankMittelbach is very live to anything that needs performance tweaks.

Witiko:
> Things that are relatively slow tend to link to TeX's basic model for variables: we have only macros and registers plus a few oddities (fontdimens for example), and that's not really something that's particularly tractable to solve.

I understand that, at least to a degree. I would assume that adding engine-level support for these higher-level expl3 variable types would be faster than emulating them using macros, registers, fontdimens, etc. The higher-level the concept, the bigger speed advantage in translating it from TeX to C, I would imagine. Regexps, key–values, and floating point arithmetic seem especially foreign from TeX's standpoint.

> @FrankMittelbach is very live to anything that needs performance tweaks.

So I noticed. :)

Witiko:
> For LuaTeX though, if you can think of something specific, I might have time to take a look and maybe add it. (I've done lots of programming in Lua, and I've wrote a tiny bit of C code in the engine itself)

Thanks, I appreciate it!

FrankMittelbach:
> Things that are relatively slow tend to link to TeX's basic model for variables: we have only macros and registers plus a few oddities (fontdimens for example), and that's not really something that's particularly tractable to solve. In any case, what we don't see is lots of slow stuff caused by `expl3` in general - @FrankMittelbach is very live to anything that needs performance tweaks.

There are a few things which I guess could be speeded up quite a lot, though I doubt it is feasible to do any of them as engine primitives. Be it as it may, they could be attempted as lua functions first. That would mean they are only producing an improvement in lualatex, but if that should enough  performance gain then as a second step one could think of lobbying for a general primitive also available in other engines.

What I would think falls into that category are "seq"  and "prop" because their storage is done in macros and that doesn't scale well because every operation essentially traverse the whole list of tokens storing the "prop" variable and that possibly multiple times. On the other hand these data types are well capsulated, so implementing them on lua level should be transparent to anything outside the LaTeX kernel --- and if not too bad :-)

josephwright:
>  Regexps, key–values, and floating point arithmetic seem especially foreign from TeX's standpoint to my eye. However, I'd rather profile the code first than trust my intuition.

Regex and FP, whilst slow, are not that heavily used (regex in particular are really the 'best' plan for only a very small number of cases in TeX). Keyval processing at the core level is not too bad: `keyval` itself is fast, `l3keys` and `expkv` also pretty quick, and higher-level stuff is best handled using pre-compilation in many cases.

josephwright:
> There are a few things which I guess could be speeded up quite a lot, though I doubt it is feasible to do any of them as engine primitives. Be it as it may, they could be attempted as lua functions first. 

All true, but LuaTeX is already relatively slow, so the sort of user who is particularly bothered by speed won't necessarily consider LuaTeX in the first place.

Witiko:
> All true, but LuaTeX is already relatively slow, so the sort of user who is particularly bothered by speed won't necessarily consider LuaTeX in the first place.

Perhaps all the more reason to offset the slowness of LuaTeX by some speediness.

FrankMittelbach:
> > There are a few things which I guess could be speeded up quite a lot, though I doubt it is feasible to do any of them as engine primitives. Be it as it may, they could be attempted as lua functions first.
> 
> All true, but LuaTeX is already relatively slow, so the sort of user who is particularly bothered by speed won't necessarily consider LuaTeX in the first place.

for one we are likely to shift towards suggesting luatex anyway (at least when accessibility is needed) and so counteracting some of luatex's slowness can be good.

secondly, it allows to see how much gain you have implementing such a data structure not using the macro-level. that may then lead eventually to providing it via engine primitives in other engines

Witiko:
> What I would think falls into that category are "seq" and "prop" because their storage is done in macros and that doesn't scale well because every operation essentially traverse the whole list of tokens storing the "prop" variable and that possibly multiple times. [...]

@FrankMittelbach: I assume that's also why there is an option to use the new constant-time "linked" property lists since February. However, according to the docs, they have their own warts (memory leakage, more difficult to copy, less efficient for short property lists). Using a Lua table for both "linked" and "flat" property lists seems like a possible improvement.

u-fischer:
> Regex and FP, whilst slow, are not that heavily used

well regex are imho what really slows down tabularray.

josephwright:
> > Regex and FP, whilst slow, are not that heavily used
> 
> well regex are imho what really slows down tabularray.

True - but back with 'its generally not a good idea' - it's simply not how TeX coding generally works.

dbitouze:
> higher-level stuff is best handled using pre-compilation in many cases.


@josephwright Could you clarify what you mean by "pre-compilation"?

josephwright:
> @josephwright Could you clarify what you mean by "pre-compilation"?

The obvious would be using templates: keyval settings which are converted to a simple list of assignments for point-of-use. That means you only parse the keyvals rarely, so they don't impact if used 100s of time in a document.

car222222:
As @FrankMittelbach wrote:

*it allows to see how much gain you have implementing such a data structure not using the macro-level*

I think that this is an important reason for experimentally adding capabilities to luatex using lua. 

Another is of course to provide better access to processes that are, at best, difficult to programme in a macro-based language. 

I shall look for further areas where this should be investigated.

car222222:
The use of lua tables for many aspects of data storage would improve many aspects.

But, yes, some thought is needed on the data types used in the table and their relationship to different types of "TeX data".

Witiko:
> For LuaTeX though, if you can think of something specific, I might have time to take a look and maybe add it. (I've done lots of programming in Lua, and I've wrote a tiny bit of C code in the engine itself)

@gucci-on-fleek I wonder if you'd be interested in defining the data structures from the l3seq and l3prop modules using Lua tables as discussed by @FrankMittelbach, @car222222, and me in https://github.com/latex3/latex3/pull/1542#issuecomment-2196436363, https://github.com/latex3/latex3/pull/1542#issuecomment-2196518178, https://github.com/latex3/latex3/pull/1542#issuecomment-2196606744, and https://github.com/latex3/latex3/pull/1542#issuecomment-2197094470.

Initially, we would need a proof-of-concept that we can benchmark against the current implementation. Therefore, it seems sufficient to define just the basic functions from both modules using Lua tables:

- `\seq_new:N` and `\prop_new:N` to initialize the data structures
- `\seq_put_right:Nn` and `\prop_put:Nnn` to store items
- `\seq_item:Nn` and `\prop_item:Nn` to retrieve the stored items

Neither the storage nor the retrieval should change the category codes in the items.

For the proof-of-concept, we don't need to consider proper scoping (`put` should be rolled back at the end of the current TeX group, likely using Lua metatables) or memory leaks (`\cs_undefine:N` should allow the Lua table to be garbage-collected unless other control sequences reference this sequence / property list), as these can be tackled later (the scoping) or left for future generations to deal with (the memory leaks). Furthermore, it seems sufficient to have the definitions in a separate `.tex` file; we can add them to the proper places in the `.dtx` files and document them later.

gucci-on-fleek:
@Witiko

> @gucci-on-fleek I wonder if you'd be interested in defining the data structures from the l3seq and l3prop modules using Lua tables as discussed by @FrankMittelbach, @car222222, and me in [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2196436363), [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2196518178), [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2196606744), and [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2197094470).
> 
> Initially, we would need a proof-of-concept that we can benchmark against the current implementation. Therefore, it seems sufficient to define just the basic functions from both modules using Lua tables:
> 
>  * `\seq_new:N` and `\prop_new:N` to initialize the data structures
> 
>  * `\seq_put_right:Nn` and `\prop_put:Nnn` to store items
> 
>  * `\seq_item:Nn` and `\prop_item:Nn` to retrieve the stored items

Sure, I should have some time later this week to take a look at this.

Skillmon:
@Witiko and @gucci-on-fleek I'd argue that part of your test should also be a `\seq_map_inline:Nn` equivalent, as that's another quite frequently used function (and one which is rather efficient in TeX).

gucci-on-fleek:
@Witiko

> @gucci-on-fleek I wonder if you'd be interested in defining the data structures from the l3seq and l3prop modules using Lua tables as discussed by @FrankMittelbach, @car222222, and me in [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2196436363), [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2196518178), [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2196606744), and [#1542 (comment)](https://github.com/latex3/latex3/pull/1542#issuecomment-2197094470).
> 

Ok, I've got an initial implementation available at gucci-on-fleek/latex3#1. Any suggestions/comments/criticism are welcome.

> Initially, we would need a proof-of-concept that we can benchmark against the current implementation. Therefore, it seems sufficient to define just the basic functions from both modules using Lua tables:
> 
>  * […] `\prop_new:N` to initialize the data structures
> 
>  * […] `\prop_put:Nnn` to store items
> 
>  * […] `\prop_item:Nn` to retrieve the stored items
> 
> 
> Neither the storage nor the retrieval should change the category codes in the items.

I've already got all of the above, albeit in a very preliminary form.

> For the proof-of-concept, we don't need to consider proper scoping (`put` should be rolled back at the end of the current TeX group, likely using Lua metatables)

This is likely to be the hardest problem. I can only really think of two solutions: copy-on-write with `\chardef`ed references, or something with `\aftergroup`. Neither is likely to have good performance characteristics (although maybe it will be fine).

> or memory leaks (`\cs_undefine:N` should allow the Lua table to be garbage-collected unless other control sequences reference this sequence / property list)

I'm not really worried about this at all: the Lua token userdata objects use 8 bytes of memory each, and a hash entry uses ~40 bytes. Unlike the classical TeX engines, LuaTeX dynamically allocates memory, so unless you have absurdly large property lists, this shouldn't be an issue. (And I can [personally attest](https://tex.stackexchange.com/a/715598/270600) that LuaTeX is totally okay with allocating 10 GiB+ of memory, although your computer might not be too happy about it).

Witiko:
> Ok, I've got an initial implementation available at https://github.com/gucci-on-fleek/latex3/pull/1 [...] in a very preliminary form.

Seems pretty complete to me! I am experimenting with the implementation and will post benchmark results later today.

> This PR is **not** intended to be merged. Rather, this is intended as somewhere to discuss Lua-based implementations of some expl3 functions *so as to not further derail PR [latex3#1542](https://github.com/latex3/latex3/pull/1542).*

At this point, I would suggest posting at least occasional summaries of progress here for visibility.

gucci-on-fleek:
How much faster would you (the LaTeX team) want a Lua implementation to be in order to justify including it in the kernel? I know that there's a fairly large maintenance burden for including new code (especially engine-dependent code), so I want to get a rough idea of what is expected so that I don't put too much effort into something impossible.

Right now, I've got a fairly complete implementation of `l3prop`. It currently supports all non-branching functions (`T`/`F` variants) and all the basic functionality seems to work correctly, although I haven't tested it too extensively.

I haven't optimized it very much yet, but some benchmarking shows that it's many times faster than the “flat” property lists, and somewhat faster than the new “linked” property lists. Here's a rough overview of my results:

<table>
    <tr>
        <td rowspan="3"></td>
        <th colspan="6" align="center">Kilo-Operations</th>
    </tr>
    <tr>
        <th colspan="3" align="center">Local</th>
        <th colspan="3" align="center">Global</th>
    </tr>
    <tr>
        <th align="right">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
        <th align="left">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
    </tr>
    <tr>
        <th align="right"><code>put-same</code></th>
        <td align="right">1320</td>
        <td align="right">34</td>
        <td align="right">21</td>
        <td align="right">1310</td>
        <td align="right">35</td>
        <td align="right">26</td>
    </tr>
    <tr>
        <th align="right"><code>put-rand</code></th>
        <td align="right">2960</td>
        <td align="right">38</td>
        <td align="right">45</td>
        <td align="right">3000</td>
        <td align="right">39</td>
        <td align="right">46</td>
    </tr>
    <tr>
        <th align="right"><code>get</code></th>
        <td align="right">905</td>
        <td align="right">28</td>
        <td align="right">16</td>
        <td align="right">936</td>
        <td align="right">28</td>
        <td align="right">16</td>
    </tr>
    <tr>
        <th align="right"><code>pop</code></th>
        <td align="right">470</td>
        <td align="right">12</td>
        <td align="right">9</td>
        <td align="right">470</td>
        <td align="right">12</td>
        <td align="right">7</td>
    </tr>
    <tr>
        <th align="right"><code>map</code></th>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">7</td>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">9</td>
    </tr>
    <tr>
        <th align="right"><code>item</code></th>
        <td align="right">1680</td>
        <td align="right">12</td>
        <td align="right">7</td>
        <td align="right">1620</td>
        <td align="right">12</td>
        <td align="right">7</td>
    </tr>
    <tr>
        <th align="right"><code>count</code></th>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">0</td>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">0</td>
    </tr>
    <tr>
        <th align="right"><code>concat</code></th>
        <td align="right">1390</td>
        <td align="right">14</td>
        <td align="right">1</td>
        <td align="right">1330</td>
        <td align="right">14</td>
        <td align="right">1</td>
    </tr>
    <tr>
        <th align="right"><code>clear</code></th>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
    </tr>
    <tr>
        <th align="right"><code>keyval</code></th>
        <td align="right">181</td>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">194</td>
        <td align="right">2</td>
        <td align="right">2</td>
    </tr>
    <tr>
        <th align="right"><code>grouping</code></th>
        <td align="right">3830</td>
        <td align="right">16</td>
        <td align="right">10</td>
        <td align="right">9430</td>
        <td align="right">16</td>
        <td align="right">7</td>
    </tr>
    <tr>
        <th align="right"><code>set-eq</code></th>
        <td align="right">0</td>
        <td align="right">2</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">2</td>
        <td align="right">0</td>
    </tr>
    <tr>
        <th align="right">Geometric Mean</th>
        <td align="right">46</td>
        <td align="right">5</td>
        <td align="right">2</td>
        <td align="right">50</td>
        <td align="right">5</td>
        <td align="right">1</td>
    </tr>
</table>

I've got the code + a few comments at gucci-on-fleek/latex3#1, so any comments/criticism are welcome, either there or here. (I can move that PR to this repo if you want; I opened it in my repo so that I didn't spam you with notifications in case you weren't interested.)

Skillmon:
Does your Lua code preserve category codes of the input? So are the tokens returned identical to the ones read in?

gucci-on-fleek:
@Skillmon 

> Does your Lua code preserve category codes of the input? So are the tokens returned identical to the ones read in?

Yes, at least from the testing that I've done, although really weird stuff like frozen relax tokens might cause issues (but should probably work fine too).

https://github.com/gucci-on-fleek/latex3/blob/9a8dc83eada3390b0014f61cc6fe12fc231f6f6a/lua-tests/expl-lua-test.tex#L347-L367

```log
=== prop: weird-catcodes ===

=== prop: gset_eq ===
The token list contains the tokens:
>  s (begin-group character s)
>  e (end-group character e)
>  c (math shift character c)
>  o (alignment tab character o)
>  n (macro parameter character n)
>  d (superscript character d)
>    (subscript character  )
>  v (blank space v)
>  a (the letter a)
>  l (the character l)
>  u (begin-group character u)
>  e (end-group character e)
>  \scan_stop: (control sequence=\relax)
```

josephwright:
@gucci-on-fleek I suggest you make a separate PR - things have rather wandered here :) I don't see an issue with a separate code path for LuaTeX provided the API is the same - we have that for `intarray` for example.

gucci-on-fleek:
@josephwright 

> I suggest you make a separate PR - things have rather wandered here :)

Ok, done; see latex3/latex3#1557.

> I don't see an issue with a separate code path for LuaTeX provided the API is the same - we have that for `intarray` for example.

:+1:

Witiko:
@gucci-on-fleek: Well done! I wonder if you are so fast because you are accelerated by Lua?!

> I haven't optimized it very much yet, but some benchmarking shows that it's many times faster than the “flat” property lists, and somewhat faster than the new “linked” property lists.

I find it surprising how fast the linked property lists are. Intuitively, I would have expected a much bigger speed-up with Lua.

josephwright:
Looking back here, I think we resolved against this - anyone think differently?

Witiko:
@josephwright: I think so. I maintain [my own version of the function `\tl_set_from_cs:NN`][1] in the Markdown Package for TeX, since the consensus here seems to be that such a function is not a good fit for the expl3 standard library.

 [1]: https://github.com/Witiko/markdown/blob/89791c4d06bb08546d50e5d21e4cab66b2ca599c/markdown.dtx#L21499-L21532

Besides that, this PR has included a discussion @gucci-on-fleek's reimplementation of parts of l3kernel in Lua in https://github.com/gucci-on-fleek/latex3/pull/1 but a separate PR has since been opened in https://github.com/latex3/latex3/pull/1557.

Therefore, please feel free to close this PR.

PR #1557, opened 2024-07-07
Exploring Lua-based data structures/functions

gucci-on-fleek:
> [!IMPORTANT]  
> This PR is **not** intended to be merged in its current state. Rather, this is intended as somewhere to discuss Lua-based implementations of some expl3 functions so as to not further derail PR latex3/latex3#1542.

Experimental, Lua-based reimplementations of a few expl3 data structures and functions. 

### Progress

- [x] **l3prop** (partially):
    - ☑ All global, non-branching functions.
    - ☑ Local functions.
    - ☐ Branching functions (`T`/`F` variants).
    - ☐ Additional profiling/performance enhancements.
- [ ] **l3seq** (planned):
    This will be implemented quite similarly to the l3prop tests.
- [ ] **l3fp** (maybe).
    A minimal implementation should be fairly simple, but perfect compatibility is probably impossible due to precision issues. The speedup here would likely be fairly large, although floating point code is used rarely enough that I doubt that it would be noticeable in an average document.
- [ ] **l3regex** (maybe):
    There's good potential here for a major speedup, but this one would be _hard_ to implement. I've already written an LPEG-based parser for a subset of TeX (for a project to extract metadata from the TUGboat back issues) which helps a little, but I still suspect that this is beyond what I'm capable of.
- [ ] **l3expan** (maybe):
    With LuaTeX, _everything_ is expandable if you try hard enough. The methods to do this aren't terribly fast themselves, but it means that you can replace a slow expandable macro with a fast non-expandable macro while still keeping it expandable. Not sure if there's any good uses for this though.
- [ ] **Others?**:
    If there's anything else that might be a good candidate, let me know and I'll add it to this list.

### Logistics
In PR latex3/latex3#1542, the discussions switched to the (mostly unrelated) topic of Lua-based implementations of some slower expl3 functions after the comment https://github.com/latex3/latex3/pull/1542#issuecomment-2196458740. To explore this idea somewhere less disruptive, I started gucci-on-fleek/latex3#1, and am now moving the discussion here as per https://github.com/latex3/latex3/pull/1542#issuecomment-2212465283. 

Right now, this current PR is not intended to be merged; instead, I'm using this PR as a thread for code + discussions. To make the development easier for me, I've placed all the new code in the folder `lua-tests/`, but once we're closer to a final implementation, I'll squash + rebase the commits and move the code to the appropriate files.


gucci-on-fleek:
### Current status (July 7th)

(mostly copied from various comments in gucci-on-fleek/latex3#1)

#### Implementation

Right now, the code is a bit of a mess, but for the most part, I'm storing the property lists as Lua tables, with the keys as strings and the values as a table of tokens (`luatex.token` userdata). At the moment, the local and global functions each use completely indepenent implementations since I was worried that correctly accounting for the grouping would have a significant performance penalty, but some initial benchmarking shows that this isn't the case, so I'll likely be able to combine these and reduce the code size by half.

#### Correctness

Right now, I have complete implementations of every non-branching function in `l3prop` (everything but `T`/`F` variants). From my initial tests, the Lua implementations behave identically to the TeX implementations, including maintaining the correct catcodes of all arguments.

To test the correctness, I've got some code in [`expl-lua-test.tex`](https://github.com/gucci-on-fleek/latex3/blob/lua-seq-prop/lua-tests/expl-lua-test.tex). To test this, disable the benchmarking code by uncommenting the `\iffalse`/`\fi` lines, then run:

```sh
$ diff --side-by-side --color=always --minimal --ignore-all-space <(lualatex expl-lua-test flat) <(lualatex expl-lua-test lua)
```

This should show that both versions give identical output, with one key difference: `\prop_map_…` maps through the lists in insertion order while Lua tables have a random insertion order. The expl3 documentation explicitly states that

> The order in which `⟨entries⟩` are returned is not defined and should not be relied upon.

so this should be fine.

(Actually, testing it right now shows that there are a few differences, so one of my changes must have broken something. It did work correctly earlier though, so I'll push out a fix shortly.)

#### Performance

If you leave the `\iffalse`/`\fi` lines commented out, the code runs in benchmarking mode, which you can test as follows:

```sh
$ lualatex expl-lua-test flat
$ lualatex expl-lua-test linked
$ lualatex expl-lua-test lua
```

I've compiled all of my LuaTeX binaries with `-g3 -O0` (no optimizations and full debug symbols), so my results likely won't generalize very well. Nevertheless, here are my results:


<table>
    <tr>
        <td rowspan="3"></td>
        <th colspan="6" align="center">Kilo-Operations</th>
    </tr>
    <tr>
        <th colspan="3" align="center">Local</th>
        <th colspan="3" align="center">Global</th>
    </tr>
    <tr>
        <th align="right">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
        <th align="left">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
    </tr>
    <tr>
        <th align="right"><code>put-same</code></th>
        <td align="right">1320</td>
        <td align="right">34</td>
        <td align="right">21</td>
        <td align="right">1310</td>
        <td align="right">35</td>
        <td align="right">26</td>
    </tr>
    <tr>
        <th align="right"><code>put-rand</code></th>
        <td align="right">2960</td>
        <td align="right">38</td>
        <td align="right">45</td>
        <td align="right">3000</td>
        <td align="right">39</td>
        <td align="right">46</td>
    </tr>
    <tr>
        <th align="right"><code>get</code></th>
        <td align="right">905</td>
        <td align="right">28</td>
        <td align="right">16</td>
        <td align="right">936</td>
        <td align="right">28</td>
        <td align="right">16</td>
    </tr>
    <tr>
        <th align="right"><code>pop</code></th>
        <td align="right">470</td>
        <td align="right">12</td>
        <td align="right">9</td>
        <td align="right">470</td>
        <td align="right">12</td>
        <td align="right">7</td>
    </tr>
    <tr>
        <th align="right"><code>map</code></th>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">7</td>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">9</td>
    </tr>
    <tr>
        <th align="right"><code>item</code></th>
        <td align="right">1680</td>
        <td align="right">12</td>
        <td align="right">7</td>
        <td align="right">1620</td>
        <td align="right">12</td>
        <td align="right">7</td>
    </tr>
    <tr>
        <th align="right"><code>count</code></th>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">0</td>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">0</td>
    </tr>
    <tr>
        <th align="right"><code>concat</code></th>
        <td align="right">1390</td>
        <td align="right">14</td>
        <td align="right">1</td>
        <td align="right">1330</td>
        <td align="right">14</td>
        <td align="right">1</td>
    </tr>
    <tr>
        <th align="right"><code>clear</code></th>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">0</td>
    </tr>
    <tr>
        <th align="right"><code>keyval</code></th>
        <td align="right">181</td>
        <td align="right">2</td>
        <td align="right">3</td>
        <td align="right">194</td>
        <td align="right">2</td>
        <td align="right">2</td>
    </tr>
    <tr>
        <th align="right"><code>grouping</code></th>
        <td align="right">3830</td>
        <td align="right">16</td>
        <td align="right">10</td>
        <td align="right">9430</td>
        <td align="right">16</td>
        <td align="right">7</td>
    </tr>
    <tr>
        <th align="right"><code>set-eq</code></th>
        <td align="right">0</td>
        <td align="right">2</td>
        <td align="right">0</td>
        <td align="right">0</td>
        <td align="right">2</td>
        <td align="right">0</td>
    </tr>
    <tr>
        <th align="right">Geometric Mean</th>
        <td align="right">46</td>
        <td align="right">5</td>
        <td align="right">2</td>
        <td align="right">50</td>
        <td align="right">5</td>
        <td align="right">1</td>
    </tr>
</table>
<table>
    <tr>
        <td rowspan="3"></td>
        <th colspan="6" align="center">Seconds</th>
    </tr>
    <tr>
        <th colspan="3" align="center">Local</th>
        <th colspan="3" align="center">Global</th>
    </tr>
    <tr>
        <th align="right">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
        <th align="left">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
    </tr>
    <tr>
        <th align="right"><code>put-same</code></th>
        <td align="left">0.3250</td>
        <td align="left">0.0085</td>
        <td align="left">0.0052</td>
        <td align="left">0.3240</td>
        <td align="left">0.0085</td>
        <td align="left">0.0064</td>
    </tr>
    <tr>
        <th align="right"><code>put-rand</code></th>
        <td align="left">0.7290</td>
        <td align="left">0.0094</td>
        <td align="left">0.0111</td>
        <td align="left">0.7400</td>
        <td align="left">0.0095</td>
        <td align="left">0.0114</td>
    </tr>
    <tr>
        <th align="right"><code>get</code></th>
        <td align="left">0.2260</td>
        <td align="left">0.0069</td>
        <td align="left">0.0040</td>
        <td align="left">0.2370</td>
        <td align="left">0.0069</td>
        <td align="left">0.0038</td>
    </tr>
    <tr>
        <th align="right"><code>pop</code></th>
        <td align="left">0.1170</td>
        <td align="left">0.0030</td>
        <td align="left">0.0021</td>
        <td align="left">0.1190</td>
        <td align="left">0.0030</td>
        <td align="left">0.0017</td>
    </tr>
    <tr>
        <th align="right"><code>map</code></th>
        <td align="left">0.00038</td>
        <td align="left">0.00064</td>
        <td align="left">0.00165</td>
        <td align="left">0.00038</td>
        <td align="left">0.00064</td>
        <td align="left">0.00233</td>
    </tr>
    <tr>
        <th align="right"><code>item</code></th>
        <td align="left">0.4180</td>
        <td align="left">0.0029</td>
        <td align="left">0.0018</td>
        <td align="left">0.4020</td>
        <td align="left">0.0029</td>
        <td align="left">0.0016</td>
    </tr>
    <tr>
        <th align="right"><code>count</code></th>
        <td align="left">0.000393</td>
        <td align="left">0.000664</td>
        <td align="left">0.000087</td>
        <td align="left">0.000380</td>
        <td align="left">0.000669</td>
        <td align="left">0.000045</td>
    </tr>
    <tr>
        <th align="right"><code>concat</code></th>
        <td align="left">0.34400</td>
        <td align="left">0.00345</td>
        <td align="left">0.00022</td>
        <td align="left">0.32800</td>
        <td align="left">0.00348</td>
        <td align="left">0.00023</td>
    </tr>
    <tr>
        <th align="right"><code>clear</code></th>
        <td align="left">0.0000010</td>
        <td align="left">0.0000022</td>
        <td align="left">0.0000942</td>
        <td align="left">0.0000011</td>
        <td align="left">0.0000023</td>
        <td align="left">0.0000005</td>
    </tr>
    <tr>
        <th align="right"><code>keyval</code></th>
        <td align="left">0.04470</td>
        <td align="left">0.00057</td>
        <td align="left">0.00062</td>
        <td align="left">0.04780</td>
        <td align="left">0.00057</td>
        <td align="left">0.00061</td>
    </tr>
    <tr>
        <th align="right"><code>grouping</code></th>
        <td align="left">0.9490</td>
        <td align="left">0.0040</td>
        <td align="left">0.0025</td>
        <td align="left">2.5500</td>
        <td align="left">0.0040</td>
        <td align="left">0.0018</td>
    </tr>
    <tr>
        <th align="right"><code>set-eq</code></th>
        <td align="left">0.00000089</td>
        <td align="left">0.00037300</td>
        <td align="left">0.00000003</td>
        <td align="left">0.00000088</td>
        <td align="left">0.00038500</td>
        <td align="left">0.00000038</td>
    </tr>
    <tr>
        <th align="right">Geometric Mean</th>
        <td align="left">0.01148</td>
        <td align="left">0.00124</td>
        <td align="left">0.00047</td>
        <td align="left">0.01253</td>
        <td align="left">0.00125</td>
        <td align="left">0.00035</td>
    </tr>
</table>
<table>
    <tr>
        <td rowspan="3"></td>
        <th colspan="6" align="center">Rank</th>
    </tr>
    <tr>
        <th colspan="3" align="center">Local</th>
        <th colspan="3" align="center">Global</th>
    </tr>
    <tr>
        <th align="right">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
        <th align="left">Flat</th>
        <th align="left">Linked</th>
        <th align="left">Lua</th>
    </tr>
    <tr>
        <th align="right"><code>put-same</code></th>
        <td align="left">6</td>
        <td align="left">3</td>
        <td align="left">1</td>
        <td align="left">5</td>
        <td align="left">4</td>
        <td align="left">2</td>
    </tr>
    <tr>
        <th align="right"><code>put-rand</code></th>
        <td align="left">5</td>
        <td align="left">1</td>
        <td align="left">3</td>
        <td align="left">6</td>
        <td align="left">2</td>
        <td align="left">4</td>
    </tr>
    <tr>
        <th align="right"><code>get</code></th>
        <td align="left">5</td>
        <td align="left">4</td>
        <td align="left">2</td>
        <td align="left">6</td>
        <td align="left">3</td>
        <td align="left">1</td>
    </tr>
    <tr>
        <th align="right"><code>pop</code></th>
        <td align="left">5</td>
        <td align="left">3</td>
        <td align="left">2</td>
        <td align="left">5</td>
        <td align="left">4</td>
        <td align="left">1</td>
    </tr>
    <tr>
        <th align="right"><code>map</code></th>
        <td align="left">1</td>
        <td align="left">3</td>
        <td align="left">5</td>
        <td align="left">2</td>
        <td align="left">3</td>
        <td align="left">6</td>
    </tr>
    <tr>
        <th align="right"><code>item</code></th>
        <td align="left">6</td>
        <td align="left">3</td>
        <td align="left">2</td>
        <td align="left">5</td>
        <td align="left">3</td>
        <td align="left">1</td>
    </tr>
    <tr>
        <th align="right"><code>count</code></th>
        <td align="left">4</td>
        <td align="left">5</td>
        <td align="left">2</td>
        <td align="left">3</td>
        <td align="left">6</td>
        <td align="left">1</td>
    </tr>
    <tr>
        <th align="right"><code>concat</code></th>
        <td align="left">6</td>
        <td align="left">3</td>
        <td align="left">1</td>
        <td align="left">5</td>
        <td align="left">4</td>
        <td align="left">2</td>
    </tr>
    <tr>
        <th align="right"><code>clear</code></th>
        <td align="left">2</td>
        <td align="left">5</td>
        <td align="left">6</td>
        <td align="left">3</td>
        <td align="left">4</td>
        <td align="left">1</td>
    </tr>
    <tr>
        <th align="right"><code>keyval</code></th>
        <td align="left">5</td>
        <td align="left">1</td>
        <td align="left">4</td>
        <td align="left">6</td>
        <td align="left">2</td>
        <td align="left">3</td>
    </tr>
    <tr>
        <th align="right"><code>grouping</code></th>
        <td align="left">5</td>
        <td align="left">4</td>
        <td align="left">2</td>
        <td align="left">6</td>
        <td align="left">3</td>
        <td align="left">1</td>
    </tr>
    <tr>
        <th align="right"><code>set-eq</code></th>
        <td align="left">4</td>
        <td align="left">5</td>
        <td align="left">1</td>
        <td align="left">3</td>
        <td align="left">6</td>
        <td align="left">2</td>
    </tr>
    <tr>
        <th align="right">Arithmetic Mean</th>
        <td align="left">4.5</td>
        <td align="left">3.3</td>
        <td align="left">2.6</td>
        <td align="left">4.6</td>
        <td align="left">3.7</td>
        <td align="left">2.1</td>
    </tr>
</table>

Based off of this, the Lua implementation appears to be many times faster than the “flat” property lists, and about 50% faster than the “linked” property lists. However, for a few operations the Lua implementation is slower than the the TeX implementations, although generally not by much. Also, the global-only Lua implementation is slightly faster than the version that properly handles grouping, but not by very much, so it's probably safe to use the grouped version for both local and global assignments.

I've also ran this through a profiler, and here are the results (you'll likely need to click to expand it):

[profile](https://tug.org/~mseven/files/expl3-lua-profile.svg)

From this, it looks like 40% of the time is spent in Lua. Drilling into this, 3.0% of the total time is spent in `next` (`pairs`/`ipairs`), 2.0% in the garbage collector, 0.6% checking the current group level, 0.2% in `token.get_mode`, 2.8% in `tex.sprint`, 2.8% in `token.get_next`, 2.2% in `token.scan_csname`, 11.2% in `token.scan_toks`, 1.6% in table indexing, 0.5% in table allocation, 1.0% in token indexing, 2.6% in adding a new key to a table, 0.4% in the length operator `#`, and 9.0% in other Lua stuff. On top of this, there's an extra 12.9% of the total time also spent in Lua. This time is tracked differently, but it's probably safe to just add 10% to all of the above numbers.

It's unclear how closely this would match the numbers with an optimized binary, but given how much time is spent in the `token` module C functions, it's unclear here how much room there is to further optimize things from Lua, but I can still think of a few potential places. I can also see a few places to optimize the C function `run_scan_toks`, although it's likely that the compiler would do lots of this for me if I were to compile with `-O3`.

#### Next steps:

* Fix the functions that I broke so that this behaves identically to the TeX implementation again.
* Add the branching `T`/`F` variant functions.
* Do some more testing to make sure that this gives identical output to the TeX implementation.
* Try and speed up the code some more with additional benchmarking/profiling.
* Add the `l3seq` functions using a similar implementation.

---

That should be it for now; if anyone has any comments/requests/criticism/etc., feel free to leave a comment or review. Thanks!

zauguin:
(Leaving low level code comments aside for now and focusing on the higher level)

I really like the approach to emulate local assignments you used here. I am a bit worried though about memory leaks from bit property lists in local scopes. Probably it's not too much of an issue but something we should be aware of.

Also we should have a stable iteration order. Not in the sense of well-documented and consistent across engines, but we should not reorder between runs of the same unchanged document. That makes compilation too nondeterministic. I am wondering if keeping track of the list of entries might even make iteration even.

zauguin:
Some comments for the list of candidate modules:

>     * [x]  **l3prop** (partially):

:+1: 

> 
>     * [ ]  **l3seq** (planned):

I would expect much better random access performance but potentially worse sequential access / iteration. It will be interesting how that affects real documents where I would expect l3seq to be used primarily for iteration.

> 
>     * [ ]  **l3fp** (maybe).
>       A minimal implementation should be fairly simple, but perfect compatibility is probably impossible due to precision issues. The speedup here would likely be fairly large, although floating point code is used rarely enough that I doubt that it would be noticeable in an average document.

I started some experiments there at some point but it's tricky. Of course it's relatively easy to implement it based on Lua's floating point numbers, but given that they are binary floats instead of decimal floats they behave very differently and are not applicable for many usecases of l3fp. So to avoid completely breaking this it needs to implement decimal floating point calculations in Lua. The advantage being that this should allow perfect compatibility.

> 
>     * [ ]  **l3regex** (maybe):
>       There's good potential here for a major speedup, but this one would be _hard_ to implement. I've already written an LPEG-based parser for a subset of TeX (for a project to extract metadata from the TUGboat back issues) which helps a little, but I still suspect that this is beyond what I'm capable of.

I agree that it would be hard. Also I don't see how LPEG helps here since it's not enough to match in strings but the actual tokens need to be taken into account.
In any case I'm not sure that it would have much impact anyway since it's probably used much less than the other entries in this list.

> 
>     * [ ]  **l3expan** (maybe):
>       With LuaTeX, _everything_ is expandable if you try hard enough. The methods to do this aren't terribly fast themselves, but it means that you can replace a slow expandable macro with a fast non-expandable macro while still keeping it expandable. Not sure if there's any good uses for this though.

I would be surprised if l3expan could be made much faster. Most of what it does already maps rather closely to primitives.

gucci-on-fleek:
Ok, I've reran the benchmarks with optimized LuaTeX binaries, and things look _much_ better for the Lua implementation:

<table><thead>
<tr>
    <th rowspan="3"></th>
    <th colspan="6">Kilo-operations</th>
</tr>
<tr>
    <th colspan="3">Local</th>
    <th colspan="3">Global</th>
</tr>
<tr>
    <th align="right">Flat</th>
    <th align="right">Linked</th>
    <th align="right">Lua</th>
    <th align="right">Flat</th>
    <th align="right">Linked</th>
    <th align="right">Lua</th>
</tr></thead>
<tbody>
<tr>
    <th align="right"><code>put-same</th>
    <td align="right">1280.0</td>
    <td align="right">31.5</td>
    <td align="right">14.5</td>
    <td align="right">1340.0</td>
    <td align="right">32.7</td>
    <td align="right">11.8</td>
</tr>
<tr>
    <th align="right"><code>put-rand</th>
    <td align="right">3180.0</td>
    <td align="right">34.7</td>
    <td align="right">20.7</td>
    <td align="right">3180.0</td>
    <td align="right">35.1</td>
    <td align="right">20.0</td>
</tr>
<tr>
    <th align="right"><code>get</th>
    <td align="right">718.0</td>
    <td align="right">25.8</td>
    <td align="right">10.6</td>
    <td align="right">781.0</td>
    <td align="right">26.5</td>
    <td align="right">10.6</td>
</tr>
<tr>
    <th align="right"><code>pop</th>
    <td align="right">331.0</td>
    <td align="right">11.4</td>
    <td align="right">7.4</td>
    <td align="right">475.0</td>
    <td align="right">11.5</td>
    <td align="right">5.9</td>
</tr>
<tr>
    <th align="right"><code>map</th>
    <td align="right">1.2</td>
    <td align="right">2.2</td>
    <td align="right">7.6</td>
    <td align="right">1.2</td>
    <td align="right">2.1</td>
    <td align="right">8.0</td>
</tr>
<tr>
    <th align="right"><code>item</th>
    <td align="right">1370.0</td>
    <td align="right">11.1</td>
    <td align="right">5.1</td>
    <td align="right">1310.0</td>
    <td align="right">11.0</td>
    <td align="right">5.0</td>
</tr>
<tr>
    <th align="right"><code>count</th>
    <td align="right">1.3</td>
    <td align="right">2.4</td>
    <td align="right">0.2</td>
    <td align="right">1.3</td>
    <td align="right">2.5</td>
    <td align="right">0.1</td>
</tr>
<tr>
    <th align="right"><code>concat</th>
    <td align="right">955.0</td>
    <td align="right">12.9</td>
    <td align="right">0.6</td>
    <td align="right">1070.0</td>
    <td align="right">12.1</td>
    <td align="right">0.5</td>
</tr>
<tr>
    <th align="right"><code>clear</th>
    <td align="right">0.0</td>
    <td align="right">0.0</td>
    <td align="right">2.6</td>
    <td align="right">0.0</td>
    <td align="right">0.0</td>
    <td align="right">0.0</td>
</tr>
<tr>
    <th align="right"><code>keyval</th>
    <td align="right">116.0</td>
    <td align="right">2.0</td>
    <td align="right">1.4</td>
    <td align="right">103.0</td>
    <td align="right">2.0</td>
    <td align="right">1.2</td>
</tr>
<tr>
    <th align="right"><code>grouping</th>
    <td align="right">1900.0</td>
    <td align="right">15.3</td>
    <td align="right">7.9</td>
    <td align="right">4310.0</td>
    <td align="right">15.1</td>
    <td align="right">6.0</td>
</tr>
<tr>
    <th align="right"><code>set-eq</th>
    <td align="right">0.0</td>
    <td align="right">1.4</td>
    <td align="right">0.0</td>
    <td align="right">0.0</td>
    <td align="right">1.5</td>
    <td align="right">0.0</td>
</tr>
<tr>
    <th align="right">Geometric Mean</th>
    <td align="right">36.0</td>
    <td align="right">4.6</td>
    <td align="right">1.6</td>
    <td align="right">40.5</td>
    <td align="right">4.6</td>
    <td align="right">0.8</td>
</tr>
</tbody></table>

The Lua implementation is faster than the linked implementation for every test except for `\prop_clear:N` and `\prop_map_…`. Using geometric means for the local tests, the Lua implementation is 2.8× faster than the linked implementation and 22.5× faster than the flat implementation.

Here's the results of the profiler when running the full benchmarks of the Lua implementation with the optimized binaries:

[profile](https://tug.org/~mseven/files/expl3-lua-profile-opt.svg)

I'll do some further investigations and see if I can speed up the `\prop_map_…` functions.

zauguin:
> Hmm, LuaTeX already links with an arbitrary precision decimal arithmetic library (`texk/web2c/mplibdir/decNumber.c`), so it should actually be pretty easy to add support for this to the engine. If we're lucky, this might already be easy to use from `mplib`, so we might not even need to modify the engine.

This is not currently easy to use from mplib since it's not exposed, so the only way to use it would be to generate actual MetaPost equivalents for expressions and then that them be evaluated. LuaMetaTeX is providing Lua access to decNumber though, so my current experiments are using that.

Skillmon:
> Not as a replacement for the current l3expan functions (which I agree would be much slower), but as some special-purpose additions so that more things become expandable. For example, `l3sort` § Expandable sorting says
> 
> > Sorting expandably is very different from sorting and assigning to a variable. Since tokens cannot be stored, they must remain in the input stream, and be read through at every step. It is thus necessarily much slower (at best O(n² ln n)) than non-expandable sorting functions (O(n ln n)).
> 
> and the description for `\seq_map_tokens:Nn` says
> 
> > The function `\seq_map_inline:Nn` is typically faster but it is not expandable.
> 
> but with `\immediateassignment` (and maybe some Lua), we can make the fast versions expandable too, then when LuaTeX is used, we could make the slow functions just wrappers around the fast functions. There's probably fairly little benefit to this though, since most performance-sensitive code uses the quick versions already.

Bad example, expandably sorting is kind of fixed already, see my (painfully still open) PR: https://github.com/latex3/latex3/pull/842 Using `\expanded` and a mergesort instead of quicksort we can drastically reduce time complexity and are actually faster than the unexpandable code for small (one could say typically sized -- small is in the order of a few hundred items in this case) datasets, and faster on already sorted input for arbitrary size.

gucci-on-fleek:
Ok, I've optimized `\lprop_map_function:NN`. Here are the results:

|           Test | Milliseconds | Kilo-operations |
| -------------: | ------------:| ---------------:|
|           Flat |        0.373 |            2.30 |
|         Linked |        0.707 |            4.58 |
| Lua (TL 2024) |        0.796 |            5.24 |
| Lua (Patched Engine) |        0.694 |            4.59 |

With a patch to the LuaTeX engine (included), the Lua implementation of `\prop_map_function:NN` is now as fast as the “linked” implementation, and without the patch, it's about 15% slower than the “linked” implementation. Sorting the property list before mapping (to give a stable iteration order) adds a 25% speed penalty, so I've commented that out for now.

Based off of the results from the profiler:

[profile](https://tug.org/~mseven/files/expl3-lua-profile-prop-map-2.svg)

I don't think that there's much room left here to optimize this any further, although if anyone has any ideas, I can try it out.

I've also combined the local and global implementations since the speed difference between the two was negligible. This also means that you can now mix local and global property lists in functions like `\lprop_concat:NNN`; however, with the current implementation, setting a value locally on a global property list will lead to unexpected behaviour. Doing so is unsupported by expl3 though, so I don't think that this should be an issue.

gucci-on-fleek:
Sorry, I seem to have forgotten about this, and I don't think that I'll have time to finish this up in the near future, so I'll close this for now, since I can always open a new PR if/when I get some time.

PR #1559, opened 2024-07-14
Redo the fix for LaTeX2e key properties

muzimuzhi:
Fixes the flawed previous fix 76bdd69d (Avoid an error if a key prop. is lacking a value, 2024-06-19), see my analysis https://github.com/latex3/latex2e/issues/1415#issuecomment-2227205603.

Unlike what @Skillmon suggested in latex3/latex2e#1415, this PR still treats `\DeclareKeys{foo .code}` the same as `\DeclareKeys{foo .code = {}}`, rather than raising an error.

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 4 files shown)


josephwright (approved):

josephwright:
I've rebased and merged the commits - thanks

PR #1564, opened 2024-07-20
Support 'opt' cmds in \GetDocumentCommandArgSpec

### LINKED ISSUE #1550: `\GetDocumentCommandArgSpec` and friends cannot handle optimized m-type cmds and envs
```tex
\documentclass{article}
\usepackage{xparse}

\DeclareDocumentCommand { \null } { } { }
\GetDocumentCommandArgSpec { \null } \ShowCommand { \ArgumentSpecification }

\begin{document}
\end{document}
```

gives
```
> \ArgumentSpecification=macro:
->\null code .
<argument> \ArgumentSpecification 
```
while since `\null` takes no arguments the `\ArgumentSpecification` here should be an empty macro.

Example is adapted from tests in `xparse009.lvt`.

https://github.com/latex3/latex3/blob/2e6909e5374f85062d8309836a3eb7301d2c5fca/l3packages/xparse/testfiles/xparse009.lvt#L17-L34

josephwright:
Fixes #1550.

### DIFF
--- a/l3packages/xparse/xparse.dtx
+++ b/l3packages/xparse/xparse.dtx
@@ -1442,7 +1442,22 @@
   {
     \__kernel_cmd_if_xparse:NTF #1
       {
-        \tl_set:Ne \ArgumentSpecification { \tl_item:Nn #1 { 2 } }
+        \tl_set:Ne \ArgumentSpecification
+          {
+            \exp_args:No \tl_if_head_eq_meaning:nNTF {#1} \@@_start_optimized:
+              {
+                \prg_replicate:nn
+                  {
+                    \str_count:e
+                      {
+                        \exp_args:Nc \cs_parameter_spec:N
+                          { \cs_to_str:N #1 \c_space_tl code }
+                      } / 2
+                  }
+                  { m }
+              }
+              { \tl_item:Nn #1 { 2 } }
+          }
         #2
       }
       {#3}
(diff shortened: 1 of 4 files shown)


l3packages/xparse/xparse.dtx
```diff
@@ -1442,7 +1442,22 @@
...
+        \tl_set:Ne \ArgumentSpecification
+          {
+            \exp_args:No \tl_if_head_eq_meaning:nNTF {#1} \@@_start_optimized:
+              {
+                \prg_replicate:nn
+                  {
+                    \str_count:e
+                      {
+                        \exp_args:Nc \cs_parameter_spec:N
+                          { \cs_to_str:N #1 \c_space_tl code }
```
muzimuzhi, line 1454: Not sure which way is better, making use of the implementation detail of `\<cmdname>~code` made from `\<cmdname>`, or the content of optimized command, in which the second (and last) token is exactly `\<cmdname>~code`.

Using the second way, we can get `\cs_parameter_spec:N \<cmdname>~code` from
```tex
\exp_last_unbraced:NNo \use_i:nn
  \cs_parameter_spec:N \mycmda
```

muzimuzhi (changes requested):
Doc of `\@@_get_arg_spec:NTF` needs an update.

l3packages/xparse/xparse.dtx
josephwright: We can rely on implementation here - this is code that is clearly part of  `ltcmd`/`xparse`.

muzimuzhi (approved):

PR #1586, opened 2024-09-02
Increase outdent of long function names

### LINKED ISSUE #1585: l3doc: long function names have smaller outdent
The function name blocks are typeset like inside `\llap` when its width is short, and when its width exceeds some value (`\marginparwidth`), a constant outdent is used.

But currently the constant outdent for long function names is smaller than the largest outdent for short names. See for example the p. 19, `interface3.pdf` (2024-08-16).

image

I'll provide a PR to increase the outdent for long names.

muzimuzhi:
Fixes #1585.

Output of `./l3kernel/testfiles-l3doc/test.lvt`:

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/l3kernel/testfiles-l3doc/test.tlg
+++ b/l3kernel/testfiles-l3doc/test.tlg
@@ -17,6 +17,89 @@ LaTeX Font Info:    External font `lmex10' loaded for size
 LaTeX Font Info:    External font `lmex10' loaded for size
 (Font)              <7> on input line ....
 LaTeX Font Info:    Trying to load font information for T1+lmss on input line ....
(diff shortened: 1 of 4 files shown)


l3kernel/testfiles-l3doc/test.tlg
muzimuzhi, line 1: Why it resulted in more deleted lines than added lines in this `.tlg`?

muzimuzhi: Because with the newly added `function` envs, `test.lvt` now produce two pages, and the old insertion position of `\showoutput` only shows the last page.

PR #1591, opened 2024-09-06
Register prefix "suanpan"

muzimuzhi:
https://ctan.org/pkg/suanpan-l3
https://gitee.com/nwafu_nan/suan-pan/blob/master/suanpan-l3.sty


josephwright:
We've typically only added stuff to the database if the author asks - did they request a PR here?

muzimuzhi:
And unfortunately all the previous registered prefixes proposed by me were actually done _without_ authors' requests. 😢
https://github.com/latex3/latex3/pulls?q=is%3Apr+author%3Amuzimuzhi+Register

FrankMittelbach:
I think we should ask or not do this in the future.

josephwright:
@FrankMittelbach  Indeed - must have missed that earlier request were not from the code maintainer(s)

PR #1594, opened 2024-09-24
fix arg type for \clist_use:nnnn

mbertucci47:
Changes `\meta` to `\Arg` for `\clist_use:nnnn` and `\clist_use:nn`.

Also fixes the referenced arg in `\clist_if_empty:nTF` from `clist var` to `comma list`.

I also noticed there's not consistency in how the `N` and `n` clist commands are documented. Some of them are documented together but only the `N` type is shown, e.g. `\clist_if_in:...`, `\clist_map_function:...`, etc. Some are documented together and both are shown, e.g. `\clist_map_tokens:...`, `\clist_rand_item:...`. And some are documented separately, e.g. `\clist_use:...`, `\clist_show:...`. Maybe this is intentional, just thought I'd point it out.

### DIFF
--- a/l3kernel/l3clist.dtx
+++ b/l3kernel/l3clist.dtx
@@ -620,8 +620,8 @@
 %
 % \begin{function}[EXP, added = 2021-05-10]{\clist_use:nnnn, \clist_use:nn}
 %   \begin{syntax}
-%     \cs{clist_use:nnnn} \meta{comma~list} \Arg{separator~between~two} \Arg{separator~between~more~than~two} \Arg{separator~between~final~two}
-%     \cs{clist_use:nn} \meta{comma~list} \Arg{separator}
+%     \cs{clist_use:nnnn} \Arg{comma~list} \Arg{separator~between~two} \Arg{separator~between~more~than~two} \Arg{separator~between~final~two}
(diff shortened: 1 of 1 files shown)


muzimuzhi (approved):

muzimuzhi:
> I also noticed there's not consistency in how the `N` and `n` clist commands are documented.

I think when the functionalities are similar, they can be documented together.

josephwright:
It's true things are not always documented together - to some extent that's because it's not clear what is 'best', as formally `n` and `N` type arguments are not the same, etc.

muzimuzhi:
Another case is when they are not added/updated on the same day. For example `\clist_use:Nnnn` was added on 2013-05-26 and the `n` version added on 2021-05-10`. But that's a limitation of `l3doc`.

josephwright:
> Another case is when they are not added/updated on the same day. For example `\clist_use:Nnnn` was added on 2013-05-26 and the `n` version added on 2021-05-10`. But that's a limitation of `l3doc`.

Well yes but over time the dates become less important - at some stage I may well 'clean up' stuff so we have a 'base date' for what we are covering. At present, I'd say documenting anything pre-2018 is simply 'always there' is not a bad position. But that's a separate discussion ...

PR #1593, opened 2024-09-16
correct global boolean, use TU with lualatex

u-fischer:
Fixes:
* in l3doc two global bool are set locally. 
* With lualatex T1-encoding is used, this switches to TU-encoding. 

(dates are not changed as I think that is done at a release?)

PR #1596, opened 2024-09-27
Fix inconsistent local/global assignments in `l3doc` and test against them

muzimuzhi:
Address problems I caught in https://github.com/latex3/latex3/pull/1593/files#r1761360349.


josephwright:
Looks good - can you fix the conflict then I can merge?

PR #1604, opened 2024-10-10
Add `\keyval_map_inline:nnn`

Skillmon:
This PR adds the function `\keyval_map_inline:nnn` and the corresponding `break`-functions.

-[ ] Unit tests

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -1096,6 +1096,76 @@
 %   \end{texnote}
 % \end{function}
 %
+% \begin{function}[added=2024-10-10]{\keyval_map_inline:nnn}
+%   \begin{syntax}
+%     \cs{keyval_map_inline:nnn} \Arg{key--value list} \Arg{inline function_1} \Arg{inline function_2}
+%   \end{syntax}
+%   Applies the \meta{inline function_1} to every key in the \meta{key--value
+%   list} which got no value, and \meta{inline function_2} to every key in the
+%   \meta{key--value list} which got a value. The \meta{inline function_1}
+%   should consist of code that receives keys without a value as |#1|, and
+%   \meta{inline function_2} receives each key as |#1| and each value as |#2|.
+% \end{function}
+%
+% \begin{function}[added=2024-10-10, rEXP]{\keyval_map_break:}
+%   Used to terminate a \cs[no-index]{keyval_map_\ldots{}} function before all
+%   entries in the \meta{key--value list} have been processed. This
+%   normally takes place within a conditional statement, for example
+%   \begin{verbatim}
+%     \keyval_map_inline:nnn { foo, bar = baz, bingo, boom, ... }
+%       {
+%         \str_if_eq:nnTF { #1 } { bingo }
+%           { \clist_map_break: }
+%           {
+%             % Do something useful
+%           }
+%       }
+%       {
+%         % Do more useful things
+%       }
+%   \end{verbatim}
+%   See also \cs{keyval_map_break:n}.
+%   Use outside of a \cs[no-index]{keyval_map_\ldots{}} scenario leads to low
+%   level \TeX{} errors.
+%   \begin{texnote}
+%     When the mapping is broken, additional tokens may be inserted
+%     before further items are taken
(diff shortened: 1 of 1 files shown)


muzimuzhi:
Comparing with existing `\tl_map_...` functions which takes a token list and `\keyval_parse:...` functions:

| <code>tl</code> | <code>keyval</code> |
|--------|--------|
| <code>\tl_map_function:nN {\<token list\>} \<function\></code> | <code>\keyval_parse:NNn \<function 1\> \<function 2\> {\<key-value list\>}</code> |
| <code>\tl_map_inline:nn {\<token list\>} {\<inline function\>}</code> | - |
| <code>\tl_map_tokens:nn {\<token list\>} {\<code\>}</code> | <code>\keyval_parse:nnn {\<code 1\>} {\<code 2\>} {\<key-value list\>}</code> |
| <code>\tl_map_variable:nNn {\<token list\>} \<variable\> {\<code using variable\>}</code> | - | 

- `\keyval_parse:NNn` is analogue to `\tl_map_function:nN` by using a function as map body,
- `\keyval_parse:nnn` is analogue to `\tl_map_tokens:nn`by using code receiving trailing brace group(s) as map body,

and the proposed `\keyval_map_inline:nnn` is analogue to `\tl_map_inline:nn` by using an inline function containing `#<n>` as map body.

- When is breaking from key-val parsing usable?
- About function names: `map` seems better in supporting `\keyval_map_break:(n)`, otherwise `parse` is OK to me.
- About argument order and language consistency: maybe there can also be `\keyval_parse_function:nNN`, `\keyval_parse_tokens:nnn` which take the `<key-value list>` as their first argument, not the last.

Skillmon:
@muzimuzhi imho, there is good reason to provide reordered versions of `\keyval_parse:...` functions. And due to the way those functions work (returning individual elements inside `\exp_not:n`) they can't possibly support `break` functions. The `\keyval_map_inline:nnn` contained here is the only function for which this is possible, and hence the only one using `map` as part of its name.

And since it's called `_map_inline:...` it has the same argument order as the other map-functions and not those of this module's `parse` functions.

Aside: I'd much prefer if every map-function in `expl3` had the same argument order like `\keyval_parse:nnn`, but that ship has sailed before I could write...

josephwright:
> Aside: I'd much prefer if every map-function in `expl3` had the same argument order like `\keyval_parse:nnn`, but that ship has sailed before I could write...

In general, we favour putting `<foo>` first for `\<foo>_<dostuff>:...` - the keyval stuff is non-standard as really these are `clist` functions :)

Skillmon:
> In general, we favour putting `<foo>` first for `\<foo>_<dostuff>:...` - the keyval stuff is non-standard as really these are `clist` functions :)


I know the convention on argument ordering (btw. one could consider `regex` to also break that rule of thumb), and as you see this PR abides to this rule.

However, `keyval` is not `clist`, only closely related. `{a=b},c` is not equivalent to `a=b,c` in `keyval`.

l3kernel/l3keys.dtx
```diff
@@ -1592,12 +1662,47 @@
...
+%   definition in a group and use \cs{keyval_parse:NNn} which we fully expand
+%   twice before the group is closed. This has the drawback that any breaking of
+%   the map happens after the entire list was already parsed, so has a
+%   performance penalty in that case. However, these should be rather rare
+%   occasions and the simplicity of the code should warrant them. Another
+%   drawback of this approach is that the expansion of the internal macros is
+%   left in the input stream for every element.
+%
+%   The break point can remain outside of the braces containing the loop, since
+%   by the time user code is executed that might actually
```
mrpiggi, line 1683: This sentence seems incomplete

car222222: It is indeed! 
And to what does "that" refer?

car222222: And what "user code" is this?

Skillmon: If only I knew, it's been a while :)

car222222: @ :-)

PR #1612, opened 2024-11-06
Add \sys_if_engine_opentype:TF

josephwright:


josephwright:
Name OK here? Do we need more docs?

FrankMittelbach (approved):
I wonder if this should have a 2e name as well and then replace our `\Umath...` tests in the 2e kernel - @davidcarlisle ?

l3kernel/l3sys.dtx
```diff
@@ -514,6 +524,18 @@
 % \end{variable}
 % \end{macro}
 %
+% \begin{variable}[pTF]{\sys_if_engine_opentype:}
```
muzimuzhi, line 527: `function`, not `variable`
```suggestion
% \begin{function}[pTF]{\sys_if_engine_opentype:}
```

l3kernel/l3sys.dtx
```diff
@@ -514,6 +524,18 @@
...
+%    \begin{macrocode}
+\@@_const:nn
+  { sys_if_engine_opentype }
+  {
+    \bool_lazy_or_p:nn
+      { \sys_if_engine_luatex_p: }
+      { \sys_if_engine_xetex_p: }
+  }
+%    \end{macrocode}
+% \end{variable}
```
muzimuzhi, line 537:
```suggestion
% \end{function}
```

l3kernel/l3unicode.dtx
```diff
@@ -390,9 +388,7 @@
   { \exp_args:Ne \@@_to_nfd:n { \int_eval:n {#1} } }
 \cs_new:Npn \@@_to_nfd:n #1
   { \@@_to_nfd:nn {#1} { \char_value_catcode:n {#1} } }
-\bool_lazy_or:nnF
-  { \sys_if_engine_luatex_p: }
-  { \sys_if_engine_xetex_p: }
+  \sys_if_engine_opentype:F
```
muzimuzhi, line 391: extra indent
```suggestion
\sys_if_engine_opentype:F
```

l3kernel/l3sys.dtx
```diff
@@ -107,6 +107,16 @@
 %   |\sys_if_engine_ptex_p:| is true for \epTeX{} but false for \eupTeX{}.
 % \end{function}
 %
+% \begin{function}[added = 2024-11-05, EXP, pTF]{\sys_if_engine_opentype:}
```
muzimuzhi, line 110: `EXP` here is option, as "`pTF` implies `EXP`".
```suggestion
% \begin{function}[added = 2024-11-05, pTF]{\sys_if_engine_opentype:}
```

muzimuzhi:
To me `opentype` is a bit indirect than `unicode`. The whole latex3 repository hardly uses the word "OpenType".

```
$ rg --ignore-case opentype
l3kernel/doc/l3news05.tex
43:\item[Dec.] Lua\TeX\ 0.65. We are happy to see Lua\TeX\ development steadily continuing. \LaTeX\ users may use Lua\TeX\ with the \verb|lualatex| program. Like \verb|xelatex|, this allows \LaTeX\ documents to use multilingual OpenType fonts and Unicode text input.

texmf/tex/generic/unicode-data/BidiMirroring.txt
51:# The OpenType Mirroring Pairs List (OMPL) is frozen to match the

texmf/tex/latex/base/ltluatex.lua
366:  find_opentype_file = data,
379:  read_opentype_file = exclusive,
```

josephwright:
> To me `opentype` is a bit indirect than `unicode`. 

We had quite a discussion about this :) The issue with 'unicode' was that there are multiple meanings - general Unicode input, math layout, font usage, and any engine that uses an OpenType font approach will definitely use Unicode input, but the opposite does not apply (upTeX).

josephwright:
@muzimuzhi  Moreover, outside of the `expl3` loader, the test here is almost always used to say 'can I load `fontspec`', which is about OpenType not about Unicode _per se_.

u-fischer:
I think the documentation should be more explizit that this currently is equivalent to a luatex+xetex test but can/will be adapted to other engines/latex formats (e.g. based on luametatex ....) when needed. And even if the list is fuzzy, one could add a few hints about some of the implicit requirements like, e.g. unicode input and support of `\Umathchardef` (needed by unicode-math fonts).

l3kernel/l3sys.dtx
```diff
@@ -514,6 +524,18 @@
 % \end{variable}
 % \end{macro}
 %
+% \begin{function}[pTF]{\sys_if_engine_opentype:}
+%    \begin{macrocode}
+\@@_const:nn
+  { sys_if_engine_opentype }
+  {
+    \bool_lazy_or_p:nn
```
davidcarlisle, line 532: This is defining `_engine_opentype` as being luatex or xetex, which isn't what I expected. I thought this would be a feature test testing for a suitable `\U` primitive such as `\Uchardef` not a named engine test

davidcarlisle (changes requested):

davidcarlisle:
@FrankMittelbach 

> I wonder if this should have a 2e name as well and then replace our `\Umath...` tests in the 2e kernel - @davidcarlisle ?

possibly but I think the definition (if not the name) should be a feature test (like `\IfTUTeX` from `iftex`)  not defined as xetex or luatex

l3kernel/l3sys.dtx
josephwright: Well yes, but that's back with it's hard to decouple the ideas anyway

josephwright:
> @FrankMittelbach
> 
> > I wonder if this should have a 2e name as well and then replace our `\Umath...` tests in the 2e kernel - @davidcarlisle ?
> 
> possibly but I think the definition (if not the name) should be a feature test (like `\IfTUTeX` from `iftex`) not defined as xetex or luatex

Well I can change the underlying definition, but I suspect we'd want to extend for any new engine anyway (there is the question of whether LuaMetaTeX counts as LuaTeX - I guess not based on pTeX vs upTeX).

l3kernel/l3sys.dtx
davidcarlisle: no test is perfect but I think making the test a feature test like choosing the name makes it clear that it is a test for the font support being there and isn't just a named engine list, and practically you want this to default as true in luatex or xetex variants (which may or may not be close enough to have the luatex or xetex tests being `true`) But mainly I think it is for documentation and messaging rather than practical difference

josephwright:
No-one has commented yet, but is `engine_opentype` OK - the other `engine_...` are engine names - I therefore wondered about `\sys_if_opentype:TF` or `\sys_if_<something>_opentype:TF`

davidcarlisle:
> No-one has commented yet, but is `engine_opentype` OK - the other `engine_...` are engine names - I therefore wondered about `\sys_if_opentype:TF` or `\sys_if_<something>_opentype:TF`

well the name depends on the definition, the current definition is more `\if_engine_luatex_or_xetex`  but I'd prefer the `opentype` name (probably defined as `\ifcsname \ifUcharder` so it picks up  extended text font support without necessarily opentype math) (\iftutex` is \ifcsname\Umathchardef`  so intentionally testing for the math support)

josephwright:
@davidcarlisle  I've adjusted

PR #1615, opened 2024-11-12
Return result of `\clist_use:N` in `\unexpanded`

muzimuzhi:


josephwright:
Interesting question - should we do what `\tl_use:N` does or what `\clist_use:Nnnn` does - hopefully others will chime in.

Skillmon:
Either be `\tl_use:N` or this, don't be `\clist_use:Nnnn`. As it currently is this can be used to serialise a clist, `\clist_use:Nnnn` can't do that as it strips braces. With brace stripping this function has zero use and is only a shorthand for `\clist_use:Nn #1 { , }`.

muzimuzhi:
> As it currently is this can be used to serialise a clist, `\clist_use:Nnnn` can't do that as it strips braces.

So the `\clits_use:N` exposes the re-bracing behavior of `\__clist_sanitize:n`. Worth documenting?

https://github.com/latex3/latex3/blob/4ea6ddab36a3255db77ee36c234e90e3f2260389/l3kernel/l3clist.dtx#L936-L944

josephwright:
> > As it currently is this can be used to serialise a clist, `\clist_use:Nnnn` can't do that as it strips braces.
> 
> So the `\clits_use:N` exposes the re-bracing behavior of `\__clist_sanitize:n`. Worth documenting?
> 
> https://github.com/latex3/latex3/blob/4ea6ddab36a3255db77ee36c234e90e3f2260389/l3kernel/l3clist.dtx#L936-L944

Not sure what you mean: all `clist` operations sanitise, so any odd entries will simply not be there.

muzimuzhi:
After `\clist_set:Nn \l_tmpa_clist { ~, a, ~b, ~{~c}, {{d}}, {}, {{}} }`, `\clist_use:N \l_tmpa_clist` gives
```
a,b,{ c},{{d}},{},{{}}
```
where odd entries are braced. This behavior is different from that of `\clist_use:Nn(nn)`, and that's why Skillmon said `\clist_use:N` "can be used to serialise a clist". Maybe this special behavior (or output format) is worth being documented.

Skillmon:
Exactly what @muzimuzhi said. That's another reason why `\exp_not:n` would've been preferable. Because then you can say something like "The result of `\clist_use:N` can be reparsed by any `n`-type taking `clist` function without a change to the data." (which is just a cumbersome way of stating "serialised"). With the current situation you're left with an unspecified number of expansion steps to reach something that's serialised. The current definition is practically useless, imho. It's not useful for typesetting (`\clist_use:Nnnn` is the better alternative there), and not useful for data manipulation because you have no idea how many expansion steps are necessary (well, `f`-type works on it, as the first entry can't start with a space, but to know that you need knowledge over the internal structure).

josephwright:
I'm going to prod the team about this - but once we have some 2e stuff done today/tomorrow

FrankMittelbach:
I think the suggested code is fine but it could perhaps do with a little be more documentation of what it does and doesn't

josephwright:
> Exactly what @muzimuzhi said. That's another reason why `\exp_not:n` would've been preferable. Because then you can say something like "The result of `\clist_use:N` can be reparsed by any `n`-type taking `clist` function without a change to the data." (which is just a cumbersome way of stating "serialised"). With the current situation you're left with an unspecified number of expansion steps to reach something that's serialised. The current definition is practically useless, imho. 

As a reminder, the reason for an addition here is to tidy up the question of `V`-type expansion - I see your point, but I also see that with the def as set now, `\clist_use:N` gives exactly the same as `V`-type expanding a `clist`. I suggest something in the docs to that effect might be sensible.

muzimuzhi:
> I suggest something in the docs to that effect might be sensible.

FYI, see d1284d1a (Extend \clist_use:N  docs, 2024-12-08) by Joseph.

PR #1625, opened 2024-12-01
Typo fixed [ci skip]

dbitouze:
On page 169 of the documentation `interface3.pdf`, the command `\int_gset:MV` is listed with `\int_set:Nn` and friends: I guess it is a typo and it should be `\int_gset:NV` instead.


muzimuzhi:
Already fixed in dev (to provide both stable and dev releases, developments now happen on branch `develop`), by https://github.com/latex3/latex3/commit/20fa9ac27aa5439fa5f5f6cfc913ec67efcaeeba.

https://github.com/latex3/latex3/blob/67c6ffc60e1396d7aecb81e87753c77d382916e0/l3kernel/l3int.dtx#L297-L301

PR #1628, opened 2024-12-05
Run CI on ubuntu-latest

muzimuzhi:
Ubuntu 20 image is closing down
https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#ubuntu-20-image-is-closing-down

### DIFF
--- a/.github/workflows/cache.yaml
+++ b/.github/workflows/cache.yaml
@@ -14,7 +14,7 @@ on:
 
 jobs:
   cache:
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


josephwright:
I think this should be shifted to `ubuntu-22.04`, or `ubuntu-24.04`, no?

zauguin:
I'm in favor of just going with `ubuntu-latest`. Nothing we run should depend on the system anyway, so in the unlikely event that having a non-fixed version breaks anything then that would probably be an actual bug.

PR #1652, opened 2024-12-30
Update l3basics.dtx

jlaurens:
\cs_if_exist:..., \cs_if_free:...: better documentation

### DIFF
--- a/l3kernel/l3basics.dtx
+++ b/l3kernel/l3basics.dtx
@@ -1186,9 +1186,10 @@
 %     \cs{cs_if_exist:NTF} \meta{control sequence} \Arg{true code} \Arg{false code}
 %   \end{syntax}
 %   Tests whether the \meta{control sequence} is currently defined
-%   (whether as a function or another control sequence type). Any
-%   definition of \meta{control sequence} other than \tn{relax}
-%   evaluates as \texttt{true}.
+%   (whether as a function or another control sequence type),
+%   and its meaning is not the primitive \tn{relax} token.
+%   This is different from \cs{if_cs_exist:N}, which evaluates to
+%   \texttt{true} if passed the token \tn{relax} as an argument.
 % \end{function}
 %
 % \begin{function}[EXP,pTF]{\cs_if_free:N, \cs_if_free:c}
@@ -1196,10 +1197,7 @@
 %     \cs{cs_if_free_p:N} \meta{control sequence}
 %     \cs{cs_if_free:NTF} \meta{control sequence} \Arg{true code} \Arg{false code}
 %   \end{syntax}
-%   Tests whether the \meta{control sequence} is currently free to
-%   be defined. This test is \texttt{false} if the
-%   \meta{control sequence} currently exists (as defined by
-%   \cs{cs_if_exist:NTF}).
+%   This test is the negation of the above \cs{cs_if_exist:NTF}.
 % \end{function}
 %
 % \subsection{Primitive conditionals}
@@ -1281,7 +1279,7 @@
 %   Check if <cs> appears in the hash table or if the control sequence
 %   that can be formed from <tokens> appears in the hash table. The
 %   latter function does not turn the control sequence in question into
-%   "\scan_stop:"! This can be useful when dealing with control
+%   the primitive \tn{relax} token. This can be useful when dealing with control
 %   sequences which cannot be entered as a single token.
 %   \begin{texnote}
 %     These are the \TeX{} primitives \tn{ifdefined} and \tn{ifcsname}.
@@ -2457,7 +2455,7 @@
 %
 % \subsection{Exist or free}
 %
-% A control sequence is said to \emph{exist} (to be used) if has an entry in
+% A control sequence is said to \emph{exist} (to be used) if it has an entry in
 % the hash table and its meaning is different from the primitive
 % \tn{relax} token. A control sequence is said to be \emph{free}
 % (to be defined) if it does not already exist.


l3kernel/l3basics.dtx
```diff
@@ -1186,9 +1186,9 @@
 %     \cs{cs_if_exist:NTF} \meta{control sequence} \Arg{true code} \Arg{false code}
 %   \end{syntax}
 %   Tests whether the \meta{control sequence} is currently defined
-%   (whether as a function or another control sequence type). Any
-%   definition of \meta{control sequence} other than \tn{relax}
-%   evaluates as \texttt{true}.
+%   (whether as a function or another control sequence type),
+% and its meaning is not the primitive \tn{relax} token
+%(in which it differs from \cs{if_cs_exist:N}).
```
muzimuzhi, line 1191: Keep indent.
```suggestion
%   (whether as a function or another control sequence type),
%   and its meaning is not the primitive \tn{relax} token
%   (in which it differs from \cs{if_cs_exist:N}).
```

l3kernel/l3basics.dtx
```diff
@@ -1197,9 +1197,7 @@
 %     \cs{cs_if_free:NTF} \meta{control sequence} \Arg{true code} \Arg{false code}
 %   \end{syntax}
 %   Tests whether the \meta{control sequence} is currently free to
-%   be defined. This test is \texttt{false} if the
-%   \meta{control sequence} currently exists (as defined by
-%   \cs{cs_if_exist:NTF}).
+%   be defined, meaning the opposite of the above.
```
muzimuzhi, line 1200: Perhaps expand "above" to "\cs{cs_if_exist:NTF}"?
```suggestion
%   be defined, meaning the opposite of the \cs{cs_if_exist:NTF}.
```

muzimuzhi (changes requested):

l3kernel/l3basics.dtx
muzimuzhi: Another aspect: doc of `\if_cs_exist:w` mentions `\scan_stop:` instead of `\relax`. Do we want to improve consistency and also use `\scan_stop:` here?

FrankMittelbach: I think it is better to talk about `\relax` in such a case because that *is* what is generated even if that is equiv to `\scan_stop:` so it should probably be corrected in  the documentation of `\f_cs_exist:w` instead.

l3kernel/l3basics.dtx
jlaurens: Changes done, I will merge the commits when everything is OK

l3kernel/l3basics.dtx
```diff
@@ -1186,9 +1186,9 @@
 %     \cs{cs_if_exist:NTF} \meta{control sequence} \Arg{true code} \Arg{false code}
 %   \end{syntax}
 %   Tests whether the \meta{control sequence} is currently defined
-%   (whether as a function or another control sequence type). Any
-%   definition of \meta{control sequence} other than \tn{relax}
-%   evaluates as \texttt{true}.
+%   (whether as a function or another control sequence type),
+%   and its meaning is not the primitive \tn{relax} token
+%   (in which it differs from \cs{if_cs_exist:N}).
```
FrankMittelbach, line 1191: I would change this to something like this:

... token.  This is different from \cs{if_cs_exist:N}, which evaluates to \texttt{true} if \tn{relax} if passed the token \tn{relax} as an argument.

FrankMittelbach (changes requested):
please also change the date in \ProvidesFile  (but not the version number as it is only a docu change).

l3kernel/l3basics.dtx
```diff
@@ -1197,9 +1197,7 @@
 %     \cs{cs_if_free:NTF} \meta{control sequence} \Arg{true code} \Arg{false code}
 %   \end{syntax}
 %   Tests whether the \meta{control sequence} is currently free to
-%   be defined. This test is \texttt{false} if the
-%   \meta{control sequence} currently exists (as defined by
-%   \cs{cs_if_exist:NTF}).
+%   be defined, meaning the opposite of the above \cs{cs_if_exist:NTF}.
```
FrankMittelbach, line 1200: I can't really see that this is clearer than the original text, on the contrary. I would suggest not to make this change.

jlaurens: The meaning of "free to be defined" is unclear. For example, temporary variables may be free to be defined despite they still contain material.
According to some comment in the implementation, `\cs_if_free:NTF` is the "logical reversal" of `\cs_if_exist:NTF`, this fact does not appear clearly in the actual documentation.

FrankMittelbach: sorry opinions differ then, it is clearly defined in the  (...) in my opinion, while  "meaning the opposite" leaves more question marks.

FrankMittelbach: if you want to strengthen the statement then I would do

This test is \texttt{false} if the \meta{control sequence} currently exists as defined by
\cs{cs_if_exist:NTF}, i.e., is the negation of that conditional.

jlaurens: Why not simply say
"""
This test is the negation of \cs{cs_if_exist:NTF}.
"""

FrankMittelbach: also fine

josephwright:
We sorted here?

jlaurens:
Rebased as one single commit.
One more change: a missing "it" in the implementation part.

FrankMittelbach (approved):

PR #1665, opened 2025-01-14
Document c-type behaviour

josephwright:


l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,14 @@ \section{Naming functions and variables}
     lists (with access at both ends) and stacks.
 \end{description}
 
+\subsection{Behavior of \texttt{c}-type expansion for undefined arguments}
+
+When \texttt{c}-type expansion is applied to with and argument that is
```
FrankMittelbach, line 340: to with and argument?

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,14 @@ \section{Naming functions and variables}
...
+undefined, the result is a token equal to |\scan_stop:| (|\relax|). This will
+likely lead to low-level errors if applied to variables. As such, the
+programmer should ensure that \texttt{c}-type expansion is only applied where
+resulting variable names will definitely exist.
```
FrankMittelbach, line 344: definitely -> already ?

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,14 @@ \section{Naming functions and variables}
     lists (with access at both ends) and stacks.
 \end{description}
 
+\subsection{Behavior of \texttt{c}-type expansion for undefined arguments}
```
FrankMittelbach, line 338: the arguments aren't undefined the produced token is, but I have no idea how to say that

FrankMittelbach (changes requested):

l3kernel/doc/source3body.tex
Skillmon: No "definitely" is correct, imho. Think of `\tl_new:c`.

l3kernel/doc/source3body.tex
Skillmon:
```suggestion
\subsection{Behavior of \texttt{c}-type expansion for undefined control sequence names}
```

FrankMittelbach: yes, better

l3kernel/doc/source3body.tex
FrankMittelbach: that's a case I haven't considered, but "definitely" doesn't  give the right vibe to me. I think it would be better to explicitly spell it out

 ...  already exists or is being defined by the command, e.g., \tl_new:c

with better wording of course.

l3kernel/doc/source3body.tex
car222222: Agreed!

l3kernel/doc/source3body.tex
car222222: Probably: 
to with and  ==>  to an

l3kernel/doc/source3body.tex
car222222: Try this: keep "definitely" and add to the end of the sentence further explanation, such as:

: i.e., each of the names either already exists, or gets defined by using this command. 
. . . not exactly correct wording! 

I did not understand the "e.g., \tl_new:c" so left it out.  What is it an example of?

car222222, line 344: And now we have "variable names": 
which seems to imply that these names vary!! 

So again: what is "a variable"?

Are the variable and its name not the same thing?
Can the variable exist but not its name, or vice versa?

car222222 (changes requested):
A lot of problems for such a short PR.
With apologies for my lack of understanding:  much of the language used seems strange to me.

l3kernel/doc/source3body.tex
josephwright: I was avoiding 'control sequence' as `expl3` talks about functions and variables - but I can see that perhaps we have no choice here

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,14 @@ \section{Naming functions and variables}
...
+
+When \texttt{c}-type expansion is applied to with and argument that is
+undefined, the result is a token equal to |\scan_stop:| (|\relax|). This will
+likely lead to low-level errors if applied to variables. As such, the
```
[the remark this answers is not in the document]
josephwright: I've changed (although to me they are the same)

l3kernel/doc/source3body.tex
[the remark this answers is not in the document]
josephwright: I've adjusted the text here

l3kernel/doc/source3body.tex
josephwright: I've tried to make it clear what I was getting at

josephwright:
Second version of text now available

FrankMittelbach (approved):
that is much clearer (to me), thanks

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,21 @@ \section{Naming functions and variables}
     lists (with access at both ends) and stacks.
 \end{description}
 
+\subsection{Behavior of \texttt{c}-type expansion where \texttt{N}-type token
+  resulting from expansion is undefined}
```
car222222, line 339: where . . . expansion ==> when the token resulting from the expansion . . .

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,21 @@ \section{Naming functions and variables}
...
+
+When \texttt{c}-type expansion is applied, it will produce an \texttt{N}-type
+token to be consumed by the underlying function. If the result of this process
```
car222222, line 342: "underlying" needed?

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,21 @@ \section{Naming functions and variables}
...
+is a token which is undefined, \TeX{}'s behavior is to make it equal to
+|\scan_stop:| (|\relax|).
+
+This will likely lead to low-level errors if it occurs in contexts where
+\pkg{expl3} expects a \enquote{variable}, e.g.~a \texttt{prop}, \texttt{seq},
+etc. Therefore, the programmer should ensure that \texttt{c}-type expansion is
+only applied where resulting \texttt{N}-type token will definitely exist,
```
car222222, line 349: . . . when the . . .

l3kernel/doc/source3body.tex
```diff
@@ -335,6 +335,21 @@ \section{Naming functions and variables}
...
+i.e.~is either defined prior to the application of the \texttt{c}-type
```
car222222, line 350: i.e, when it is either . . .

car222222 (approved):
Much better explanation now: thanks.  
Just a few improvements (I hope!).

l3kernel/doc/source3body.tex
josephwright: I think so - I'm trying to get at `\foo:c` => `\foo:N`

l3kernel/doc/source3body.tex
Skillmon: @car222222 `\tl_new:c` is a command that must not be used with an already existing macro, so in the no-error-case is guaranteed to let the control sequence to `\relax`, but it'll definitely be defined by it.

I know I'm now late to suggest any changes here, but I'd argue a simple "eventually" at the end of that sentence would explain it quite elegantly in a single word.

car222222: Re: \tl_new:c is a command that must not be used with an already existing macro

OK, but there is still a need to explain why this example is bring used, and what it explains.  None of this is at call ,clear at present, is it?

l3kernel/doc/source3body.tex
car222222: This did not get changed!

car222222:
There remain a few places at which 
where ==> when 
would be a marked improvement!

josephwright:
> There remain a few places at which where ==> when would be a marked improvement!

I see three "where", of which I have changed two for "when".

car222222:
Good! ? 
I shall look again at the remaking one.

josephwright:
> Good! ? I shall look again at the remaking one.

> his will likely lead to low-level errors if it occurs in contexts where `\pkg{expl3}` expects a `\enquote{variable}`,

car222222:
I found four occurrences.

But only this one is doubtful 
(line 521, I think)

given in the documentation. Where the documented behaviour of a function

Better changed to "when" or "if".

josephwright:
> I found four occurrences.
> 
> But only this one is doubtful (line 521, I think)
> 
> given in the documentation. Where the documented behaviour of a function

That's outside the scope of the PR :)

car222222:
Sure, but you are nevertheless free to change it! Or not?

PR #1669, opened 2025-01-17
Fix gh1667

Skillmon:


josephwright (approved):

l3kernel/l3keys.dtx
```diff
@@ -2858,40 +2861,47 @@
...
-    \__kernel_tl_set:Nx \l_@@_selective_clist {#2}
-    \__kernel_tl_set:Nx \l_@@_relative_tl { \exp_not:n {#3} }
   }
 \cs_new:Npn \@@_reset_bool:N #1
   {
     \exp_not:c
       { bool_set_ \bool_if:NTF #1 { true } { false } :N }
         \exp_not:N #1
   }
+\cs_new:Npn \@@_reset_var:N #1
```
muzimuzhi, line 2899: `\begin{macro}` not updated.

Also `\begin{macro}{\@@_set:nnnnnnnNn}` and an `\end{macro}` need to be removed.

l3kernel/testfiles/m3keys_gh1667.lvt
muzimuzhi, line 1: Perhaps move it into `m3keys003.lvt`, right before testing `.multichoices:nn` property?
https://github.com/latex3/latex3/blob/6573e2af9bdb85844cca60991a8d692331f4753f/l3kernel/testfiles/m3keys003.lvt#L119

l3kernel/l3keys.dtx
Skillmon: `\@@_reset_var:N` was already in a `\begin{macro}` but never actually defined. Removed the obsolete `macro` environment.

l3kernel/testfiles/m3keys_gh1667.lvt
Skillmon: Done (but at the end of that file, since that's below the `.inherit:n` property, which is necessary for this test as well).

l3kernel/l3keys.dtx
muzimuzhi: You're right. It was added by efe0c495 (Re-factor selective key setting code, 2024-05-08).

Issue #1518, opened 2024-03-25
Loading `expl3-generic.tex` in LuaMetaTeX fails with `expl3.lua:421: table index is nil` since TeX Live 2024


Witiko:
Take the following MWE document `example.tex`:

```
\input expl3-generic
```

Compiling this document using the command `context example.tex` on TeX Live 2024 fails with the following output:

```
open source     > level 1, order 3, name './example.tex'
open source     > level 2, order 4, name 'expl3-generic.tex'
Package: expl3 2024-02-20 L3 programming layer (loader)

open source     > level 3, order 5, name 'expl3-code.tex'
Package: expl3 2024-02-20 L3 programming layer (code)

lua error       > lua error on line 154 in file expl3-code.tex:

token call, execute: ...cal/texlive/2024/texmf-dist/tex/latex/l3kernel/expl3.lua:421: table index is nil
stack traceback:
	...cal/texlive/2024/texmf-dist/tex/latex/l3kernel/expl3.lua:421: in main chunk
	[C]: in upvalue 'requiem'
	...live/2024/texmf-dist/tex/context/base/mkiv/l-sandbox.lua:180: in function <...live/2024/texmf-dist/tex/context/base/mkiv/l-sandbox.lua:165>
	(...tail calls...)
	[ctxlua]:1: in main chunk
<empty file>
mtx-context     | fatal error: return code: 1
```

Here is the corresponding line 421 in `expl3.lua`:

```
    [cmd'assign_mu_glue' or cmd'register_mu_glue'] = index_not_nil,
```

Compiling the same document using the same command on TeX Live 2023 or on TeX Live 2024 but using the command `context --luatex example.tex` succeeds. This indicates that since TeX Live 2024, `expl3-generic.tex` fails to load with LuaMetaTeX, whereas it succeeds with LuaTeX.

josephwright:
I've tracked down this part: LMTX has renamed a few functions again. (Aside: we need to decide whether to keep the intermediate names or to drop them).

There's a second issue: I'm working on that too.

josephwright:
Right, second issue is that LuaMetaTeX accepts `;` as part of `\numexpr` (seems to be undocumented) - we use that a lot as a delimiter, so need to change it.

josephwright:
OK, I'll check: my feeling is this is by-design but you never know.

josephwright:
[...]

PR #1670, opened 2025-01-17
Remove `;` from the syntax of `l3fp`

Skillmon:
This PR entirely removes `;` from the syntax of `l3fp` internals. Instead a scan marker `\s__fp_sep` is used that isn't absorbed by `\numexpr` et al. (`\s__fp_sep` is the TeX primitive `\right` -- though almost any non-expandable primitive would do).


Skillmon:
I also reflowed some text to stay below 80 characters width.

josephwright (changes requested):
Nice work!

I think that we should not 'abuse' a scanmark here: they are meant not only to be equal to `\scan_stop:` but also tracked so we only create each once. So I'd prefer just `\@@_sep:` or similar here -  a unique token we can document but which doesn't abuse the 'rules'.

This should allow us to work again with LuaMetaTeX, so I think a ChangeLog entry should be added. I've reopened https://github.com/latex3/latex3/issues/1518 for that - it will fix it at our end, and we can then let Hans know that the problem is solved and he can avoid one of his many engine switch things :)

josephwright:
I'll sort the changes shortly.

car222222:
Does this new get separator defined ? 
For safety, Should it be equated to \relax 
or not?

josephwright:
Turns out a bit more work is still needed for LuaMetaTeX, but I think we are almost there - so I'll merge and make a note in the ChangeLog.

josephwright:
> Does this new get separator defined ? For safety, Should it be equated to \relax or not?

Yes, it gets defined: it needs to be a distinct token from `\relax`, so @Skillmon has gone for `\right` - with a dedicated internal name, we are much freer to pick whatever we want as a meaning.

josephwright (approved):

josephwright:
We should probably add a test for LuaMetaTeX - I'll sort that once the PR is merged.

car222222:
@Skillmon has gone for \right

Yes I recall now.

Please can someone explain why it must be a defined primitive, but \relax will not work?

josephwright:
> @Skillmon has gone for \right
> 
> Yes I recall now.
> 
> Please can someone explain why it must be a defined primitive, but \relax will not work?

@car222222  As it says in the comments in `l3fp-aux.dtx`, we can't afford for the token to be absorbed - which would be the case for `\relax` due to the syntax of `\numexpr`. So it needs to be (a) non-expandable and (b) not `\ifx`-equal to `\relax`.

car222222:
Makes sense! Thanks.

PR #1687, opened 2025-02-16
Update l3prefixes.csv

lvjr:
Rename prefix `codehigh` as `cdhh`; register prefixes `speg` and `tpeg`.


muzimuzhi:
> register prefixes `speg` and `tpeg`.

Is `tpeg` registered as a placeholder? As it seems currently prefix `tpeg` is totally replaced with `speg` in https://github.com/lvjr/pegmatch.

lvjr:
Yes, `speg` for str matching, and `tpeg` for (future) tl matching.

josephwright:
I'm confused by the removal: the code is presumably out in the wild, so shouldn't `codehigh` stay as a 'taken' prefix?

lvjr:
I want a shorter prefix to reduce line breaks in the package code.

josephwright:
@lvjr  Sure, the addition makes total sense: I mean that once a prefix is used, it can't ever really be released - old code still uses `codehigh` so another module can't take it.

lvjr:
OK, I have cancelled the removal of the old prefix.

PR #1694, opened 2025-02-28
Carelessness: An unpaired `macro` environment in the `l3galley.dtx`.

myhsia:


josephwright:
You might want to thing of slightly different commit messages -  'Add missing `\end{marco}`' is traditional :)

PR #1699, opened 2025-03-01
Add link on/off functionality to backend

josephwright:


u-fischer:
with lualatex the number is not properly terminated and e.g. `
\__pdfannot_backend_link_on: 2 x`  or `\pdfannot_link_on: 2 x`  eats the 2. 

I don't think that you need to provide the old name `\__pdf_backend_link_on:`. As long as the pdfresources needs it, it can define it itself, and if not it can go.

josephwright:
> with lualatex the number is not properly terminated and e.g. ` \__pdfannot_backend_link_on: 2 x` or `\pdfannot_link_on: 2 x` eats the 2.

I wasn't sure - I'll add something

> I don't think that you need to provide the old name `\__pdf_backend_link_on:`. As long as the pdfresources needs it, it can define it itself, and if not it can go.

Will adjust

FrankMittelbach:
> > with lualatex the number is not properly terminated and e.g. ` \__pdfannot_backend_link_on: 2 x` or `\pdfannot_link_on: 2 x` eats the 2.
> 
> I wasn't sure - I'll add something

If that is the case then perhaps the command should be `:w`?

FrankMittelbach (approved):
ignore my earlier nonsense comment

josephwright:
> ignore my earlier nonsense comment

:)

Was mainly wanting to check this is going to fit with your OR plans

FrankMittelbach:
that's for Ulrike to make sure. I don't use any internal backend commands :-)

u-fischer (approved):

PR #1703, opened 2025-03-04
Require fp indentifiers to be declared first

### LINKED ISSUE #1700: Check existence of fp word for `\fp_set_function:nnn` and `\fp_set_variable:nn`
At least when `\debug_on:n { check-declarations }` is set.

Originally reported in https://tex.stackexchange.com/q/738147, "Why do we need \fp_new_function:n?"

muzimuzhi:
Fixes #1700.


l3kernel/l3fp-symbolic.dtx
```diff
@@ -1,6 +1,6 @@
 % \iffalse
 %
-%% File l3fp-symbolic.dtx (C) Copyright 2012-2015,2017,2018,2020,2021,2023 The LaTeX Project
+%% File l3fp-symbolic.dtx (C) Copyright 2012-2025 The LaTeX Project
```
josephwright, line 3: This doesn't look right - according to the older line, there were no changes in 2016, 2022, 2024.

josephwright (changes requested):

l3kernel/l3fp-symbolic.dtx
muzimuzhi: > there were no changes in 2016, 2022, 2024.

At least there is cf9c00cb (Use N-type flags in l3fp, l3regex, l3str-convert (instead of n-type), 2024-01-12) made in 2024.

Once there was a discussion in l3build repo, and the conclusion there was to use a simple range instead of a list of ranges and years. See https://github.com/latex3/l3build/issues/325#issuecomment-1856592871.

Perhaps the conclusion can also be applied here?

PR #1706, opened 2025-03-09
Update l3tl.dtx

jlaurens:
Some editing methods do not accept `{`,  `}`, `#` in the argument.
A short indication is added to suggest the use of a `\tl_regex...` function in that case.


josephwright:
I'd merged but the overall team feeling is we should avoid recommended `regex` for most issues - they are very very slow, really mainly for 'testing' not production. So the balance is to leave as is for the present - a more general statement would likely say that low-level approaches may be used if support for these tokens is essential.

Issue #1597, opened 2024-09-29
Facilitate precompiled preamble in LuaTeX?


user202729:
Currently, there is some code in the LaTeX format that propagates things such as intarray across dumps/runs (from the format generation to the actual run).

However, if there are more than two runs, the mechanism breaks down.

```bash
lualatex '\ExplSyntaxOn\typeout{\intarray_count:N \c__codepoint_lowercase_index_intarray}\stop'
# 17408, correct

lualatex --ini --jobname=a \&lualatex '\dump'
# create a.fmt

lualatex \&a '\ExplSyntaxOn\typeout{\intarray_count:N \c__codepoint_lowercase_index_intarray}\stop'
# 0, incorrect!
```

This causes difficulty while trying to use precompiled preamble or packages such as `mylatexformat`.

Can this be fixed?

josephwright:
@zauguin is best placed to answer, but I remain wary of 'local' formats - the number of additional macros and therefore processing time tends to be low in the grand scheme of things

user202729:
I think this is not too hard to fix, the key problem is in the second run if `--ini` is passed then the stored data is wiped.

Just fix it to define `get_luadata` whenever the data is available, like this:

```diff
diff --git a/l3kernel/l3luatex.dtx b/l3kernel/l3luatex.dtx
index 9c1cabc..e352e43 100644
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -678,7 +678,6 @@ local register_luadata, get_luadata
 
 if luatexbase then
   local register = token_create'@expl@luadata@bytecode'.index
-  if status.ini_version then
 %    \end{macrocode}
 %
 % \begin{macro}[int]{register_luadata}
@@ -688,6 +687,7 @@ if luatexbase then
 % called in the \texttt{pre_dump} callback and which has to return a string that
 % evaluates to a valid Lua object to be preserved.
 %    \begin{macrocode}
+  if status.ini_version then
     local luadata, luadata_order = {}, {}
 
     function register_luadata(name, func)
@@ -713,14 +713,18 @@ if luatexbase then
         lua.bytecode[register] = assert(load(str .. "}"))
       end
     end, "ltx.luadata")
-  else
+  end
 %    \end{macrocode}
 %
 % \begin{macro}[int]{get_luadata}
 % \texttt{get_luadata} is only available if data should be restored.
 % It accept the identifier which was used when the data object was registered and
[...]

PR #1709, opened 2025-03-12
Preserve intarray across multiple dumps

user202729:
Fixes https://github.com/latex3/latex3/issues/1597 . Uses the patch there.

I don't think there's any disadvantage (for users who doesn't use precompiled preamble, the only visible difference is that in the second run `register_luadata` is also available, but if they just don't use it then there's no difference)

### DIFF
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -668,7 +668,6 @@ local register_luadata, get_luadata
 
 if luatexbase then
   local register = token_create'@expl@luadata@bytecode'.index
-  if status.ini_version then
 %    \end{macrocode}
 %
 % \begin{macro}[int]{register_luadata}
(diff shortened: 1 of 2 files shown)


josephwright (changes requested):
Something should go into the ChangeLog - I also want to see what @zauguin thinks

josephwright:
@zauguin Ping

l3kernel/l3luatex.dtx
```diff
@@ -713,14 +713,18 @@ if luatexbase then
...
 %
 % \begin{macro}[int]{get_luadata}
 % \texttt{get_luadata} is only available if data should be restored.
 % It accept the identifier which was used when the data object was registered and
 % returns the associated object. Every object can only be retrieved once.
+% Note that it is possible for both \texttt{register_luadata} and \texttt{get_luadata}
+% to be available, for example while compiling a precompiled preamble.
+% In such a case, data must be registered again to be kept in the next run.
 %    \begin{macrocode}
+  do
```
zauguin, line 727: The comment needs to be adjusted since `get_luadata` is afterward always available. Also the `do` / `end` wrapper doesn't seem useful and should be removed.

zauguin (approved):

zauguin:
IMO the change looks good except for documentation / changelog adjustment.

l3kernel/l3luatex.dtx
user202729: you mean `register_luadata` is the one always available? `get_luadata` should not be available on first run.

The `do`/`end` is because I try to preserve the semantic of the existing code where `local luadata` is a block-local variable.

PR #1710, opened 2025-03-14
Move l3graphics to l3kernel (stable)

josephwright:
Updated PR that avoids the history issues


davidcarlisle (approved):
easier to see after rebase, thanks

davidcarlisle:
```
- .\kern0.35878 (italic)
- ```
failure doesn't seem directly related to l3graphics?

josephwright:
Thanks - looking back the failure is another LuaTeX change (so my macOS setup doesn't see it yet): I'll fix then merge.

PR #1714, opened 2025-03-17
Move l3opacity to stable

josephwright:


texmf/tex/latex/latex-lab/documentmetadata-support.ltx
```diff
@@ -0,0 +1,273 @@
+%%
+%% This is file `documentmetadata-support.ltx',
+%% generated with the docstrip utility.
+%%
+%% The original source files were:
+%%
+%% documentmetadata-support.dtx  (with options: `code')
```
FrankMittelbach, line 7: This is a generated file, does that need to be checked in?

texmf/tex/latex/pdfmanagement-testphase/l3backend-testphase-dvipdfmx.def
```diff
@@ -0,0 +1,612 @@
+%%
+%% This is file `l3backend-testphase-dvipdfmx.def',
+%% generated with the docstrip utility.
+%%
+%% The original source files were:
+%%
+%% l3backend-testphase.dtx  (with options: `drivers,dvipdfmx')
```
FrankMittelbach, line 7: and so is this one (and probably more)

texmf/tex/latex/tagpdf/tagpdf-base.sty
```diff
@@ -0,0 +1,163 @@
+%%
+%% This is file `tagpdf-base.sty',
+%% generated with the docstrip utility.
+%%
+%% The original source files were:
+%%
+%% tagpdf.dtx  (with options: `base')
+%% tagpdf-mc-generic.dtx  (with options: `base')
```
FrankMittelbach, line 8: and this

FrankMittelbach (commented):
basically a question on what these generated files do in the commit

u-fischer:
I think a stubb l3opacity.sty is needed. Handling the transparent package is easy, but l3opacity is also used in latex-lab for some debugging code and it would block your PR if you have to wait until the next latex release ...

texmf/tex/latex/latex-lab/documentmetadata-support.ltx
FrankMittelbach: My gut feeling is that it is better to use search than to require us to update what is in texmf each time something changes with these files. I think there is a testfiles dir that is for checksearch=true if not we could make one and have the tests that needs this placed in there to keep the earch case at a minimum

josephwright: I've adjusted

josephwright:
> I think a stubb l3opacity.sty is needed. Handling the transparent package is easy, but l3opacity is also used in latex-lab for some debugging code and it would block your PR if you have to wait until the next latex release ...

Stub added - I've also done `l3graphics`

l3kernel/l3opacity.dtx
```diff
@@ -94,16 +91,24 @@
...
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\ProvidesExplPackage{l3opacity}{2025-03-10}{}
-  {L3 Experimental opacity support}
+%</stub>
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+%<*package>
```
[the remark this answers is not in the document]
josephwright: That's the stub bit for people who were loading `l3opacity` separately - the idea is that it's a no-change outcome.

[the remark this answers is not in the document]
FrankMittelbach: not the stub, I'm refering to the label called "package"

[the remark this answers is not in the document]
FrankMittelbach: but maybe I'm reading the diff incorrectly and it is indeed the stub, let me check

u-fischer (approved):
As a note, opacity doesn't work if one load the backend before \DocumentMetadata. Not sure if one can/should do something about it. (I fell over it as I had a line to uncompress the PDF in my test file from a test without \DocumentMetadata.).

~~~~
\ExplSyntaxOn\sys_ensure_backend:\pdf_uncompress:\ExplSyntaxOff
\DocumentMetadata{}
\documentclass[]{report}
\begin{document}
\ExplSyntaxOn
xxxx \opacity_select:n{0.5} aaaa
\ExplSyntaxOff
\end{document}
~~~~

josephwright:
> As a note, opacity doesn't work if one load the backend before \DocumentMetadata. Not sure if one can/should do something about it. (I fell over it as I had a line to uncompress the PDF in my test file from a test without \DocumentMetadata.).
> 
> ```
> \ExplSyntaxOn\sys_ensure_backend:\pdf_uncompress:\ExplSyntaxOff
> \DocumentMetadata{}
> \documentclass[]{report}
> \begin{document}
> \ExplSyntaxOn
> xxxx \opacity_select:n{0.5} aaaa
> \ExplSyntaxOff
> \end{document}
> ```

I can see why: I'll fix that separately from the PR.

PR #1716, opened 2025-03-25
Move l3benchmark to l3kernel

josephwright:


FrankMittelbach (commented):
One remark though. The fact that we move it to the kernel is perhaps something we should done as a section in ltnews41. At least for the benchmark stuff this is of interest to some people.

josephwright:
> One remark though. The fact that we move it to the kernel is perhaps something we should done as a section in ltnews41. At least for the benchmark stuff this is of interest to some people.

I'll wait for @car222222's review then I'll draft something. (I want to be sure we are OK with the function names.)

car222222 (approved):
Not much to see here, but no obvious problems with the move.

josephwright:
> Not much to see here, but no obvious problems with the move.

Thanks - you'll likely have more comment on future ones :)

car222222:
The function names have not changed, have they?
Are there any that you are particularly concerned about?

josephwright:
> The function names have not changed, have they? Are there any that you are particularly concerned about?

No, no change - I just wanted to be sure that I'd checked before moving them to stable, that’s all.

car222222:
Do _tic and _toc refer to the sounds it makes when not using _silent . . . ?? 

Presumably all this is made clear in the full text.

josephwright:
> Do _tic and _toc refer to the sounds it makes when not using _silent . . . ??
> 
> Presumably all this is made clear in the full text.

Yes - it is :)

FrankMittelbach:
> I'll wait for @car222222's review then I'll draft something. (I want to be sure we are OK with the function names.)

looking over the function names, I'm wondering if it makes sense to (also) have `\benchmark_tic:n{STRING}` and `\benchmark_toc:n{STRING}` where in "tic" the STRING is also displayed and in "toc" it is additionally compared to the "tic" we measured from. But I haven't played much with the benchmarking so I'm not too sure how useful the idea is.

FrankMittelbach:
> > Do _tic and _toc refer to the sounds it makes when not using _silent . . . ??
> > Presumably all this is made clear in the full text.
> 
> Yes - it is :)

can't really find the names explained, at least not in the user documentation you get with texdoc l3benchmark.  but to me the meaning seems pretty clear ...it means we measure the time between a tic and a toc  (as in an old fashioned clock) to find out how much time between the two events are in reality, ie is it running slow or fast

car222222:
But what causes the tick and tock??
Using the spellings I was brought up with -  and noting that, canonically, toc == table of contents.  
Not that the names matter too much, provided they are clearly explained somewhere,

mrpiggi:
> But what causes the tick and tock?? Using the spellings I was brought up with - and noting that, canonically, toc == table of contents. Not that the names matter too much, provided they are clearly explained somewhere,

At least for MATLAB are equally named functions with the same purpose: https://de.mathworks.com/help/matlab/ref/tic.html

Skillmon:
> At least for MATLAB are equally named functions with the same purpose: https://de.mathworks.com/help/matlab/ref/tic.html

Exactly, these are canonical programming-folks names for such functions. So to many programmers they sound natural and one doesn't think of table of contents when one reads `\benchmark_toc:`.

FrankMittelbach:
> > > I'll wait for @car222222's review then I'll draft something. (I want to be sure we are OK with the function names.)
> > 
> > 
> > looking over the function names, I'm wondering if it makes sense to (also) have `\benchmark_tic:n{STRING}` and `\benchmark_toc:n{STRING}` where in "tic" the STRING is also displayed and in "toc" it is additionally compared to the "tic" we measured from. But I haven't played much with the benchmarking so I'm not too sure how useful the idea is.
> 
> Are we expecting intertwined usages of `l3benchmark`? This would only make sense for such uses, but I highly doubt they're probable.

in my opinion it would already make sense if I'm adding in several tic/tocs  in different parts of the code and just don't want to confuse myself on which is which when looking at the output, but I haven't played with  the toc/tocs of the benchmark module so I don't really know how valid or useful that observation/suggestion is.

josephwright:
That's a feature request so I think the current merge is okay: one to add with someone had a moment

Skillmon:
> in my opinion it would already make sense if I'm adding in several tic/tocs in different parts of the code and just don't want to confuse myself on which is which when looking at the output, but I haven't played with the toc/tocs of the benchmark module so I don't really know how valid or useful that observation/suggestion is.

Hm, not sold, that's just a `\typeout` or similar for something that seems like a really rare use case.

PR #1742, opened 2025-05-17
Shorten/standardize tmp var names in docs

muzimuzhi:
Follow-up to f65556b1b (Shorten/standardize tmp var names, 2025-05-16). Full changes https://github.com/latex3/latex3/compare/963a29951d...6fedc63d51.

### DIFF
--- a/l3kernel/l3seq.dtx
+++ b/l3kernel/l3seq.dtx
@@ -1060,7 +1060,7 @@
 % The code as written here only works if \meta{seq~var_3} is different
 % from the other two sequence variables.  To cover all cases, items
 % should first be collected in a sequence
-% |\l__|\meta{pkg}|_internal_seq|, then \meta{seq~var_3} should be set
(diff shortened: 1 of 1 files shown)


muzimuzhi:
BTW `l3kernel/l3coffins.dtx` uses two internal variables `\l_@@_aligned_coffin` and `\l_@@_aligned_internal_coffin`. Perhaps they can be renamed to, for example `\_@@_aligned_tmp(a|b)_coffin` (plus doc clean-up as `\c_empty_coffin` is created automatically now)?

https://github.com/latex3/latex3/blob/9949aba0de3df015b1be7da8f20a0dc1fb217407/l3kernel/l3coffins.dtx#L978-L993

josephwright:
@muzimuzhi  I'd thought about that one - the thing is it is internal to the aligned code, so it's 'more internal' than normal :)

PR #1705, opened 2025-03-06
Correct typos

muzimuzhi:
All except for the 'these' -> `that` one were caught by [`typos`](https://github.com/crate-ci/typos), command [`typos l3kernel/*.dtx`](https://github.com/crate-ci/typos).

The `typos` cli provides a GitHub Action, which can be used as `uses: crate-ci/typos@v1`. A whitelist is needed if we'd like to add it to CI.

PR #1745, opened 2025-05-24
Add spelling check to CI

muzimuzhi:
Follow-up to https://github.com/latex3/latex3/pull/1705#issuecomment-2709289275.

### DIFF
--- a/typos.toml
+++ b/typos.toml
@@ -6,6 +6,7 @@ extend-exclude = [
   "*.eps",
   "*.lvt",
   "*.tlg",
+  "l3trial/l3htoks/tokens-use-count.txt",
   "support/texlive.sh",
   "support/UShyphen.tex",
   "texmf/**",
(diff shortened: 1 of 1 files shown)


typos.toml
```diff
@@ -0,0 +1,43 @@
+# https://github.com/crate-ci/typos/blob/master/docs/reference.md
+
+[files]
+extend-exclude = [
+  "*.csv",
+  "*.eps",
+  "*.lvt",
```
muzimuzhi, line 7: Maybe a bit brute to ignore all the lvt files. But otherwise we end with a not-too-short list of lvt file paths to ignore.

josephwright:
Looks good but we need to skip the historical dirs, in particular `xpackages`.

typos.toml
```diff
@@ -0,0 +1,43 @@
...
+  "IEEE Comput\\.",
+  "juxt(_int|:N)", # special csnames used in l3fp-parse.dtx
+  "LSO",
+  "mis-",
+  "nd",
+  "Noo",
+  "openin",
+  "OT1",
+  "parm",
+  "pre-empted",
```
josephwright, line 40: I guess should go for US-style spelling?

FrankMittelbach:
nice. I wonder what something like this would do to the 2e sources

typos.toml
FrankMittelbach: yes you should I guess, sorry :-)

typos.toml
```diff
@@ -0,0 +1,43 @@
...
+extend-ignore-re = [
+  ":[NncVvoxefTFpwD]+",
+  "[aA]lph",
+  "\\.ist",
+  "\\(mis\\)use",
+  "\\{ (fo|(03)?BA|Nd) \\}",
+  "\\{inteval\\}",
+  "ABD",
+  "fith",
+  "for emptyness",
```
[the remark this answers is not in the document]
muzimuzhi: Fixed now. I got "pre-empted" and "emptyness" mixed up :(

typos.toml
muzimuzhi: Appending config `locale = "en-us"` would result a looong list of corrections, thus I prefer doing it in a separate PR.

To ask in advance, what's your opinion on renaming `\sys_finalise:` to `\sys_finalize:` (ise -> ize)?

PR #1748, opened 2025-05-26
Macros for typesetting the key list

myhsia:
Added macros:

- `\keys [<default key>] {<keys list>}`: typeset the keys list with | separated, and emphasize the default key in the keys list if necessary.
- `\keysarg [<default key>] {<keys list>}`: typeset as `\keys` but wrap it with angle brackets `<...>`.
- `\bTF [default Boolean]`: Set `<keys list>` in the 2nd argument of `\keys` as `true, false`.
- `\bTFarg [default Boolean]`: typeset as `\bTF` but wrap it with angle brackets `<...>`.


myhsia:
> @myhsia
> 
> Please note, we do not normally accept pull requests.
> 
> If you feel there are missing features please first open an issue describing the problem, preferably with a test file. If it is decided that code changes are needed then in some cases we may except Pull Requests.
> 
> However stability considerations of `l3doc` means that we are unlikely to accept extensions for l3doc.

Mentioned in Issue #1749.

davidcarlisle:
@myhsia  I'll close the PR for now, if you open an issue describing any problem you are having with l3doc then it could be re-opened, although as already stated, we are unlikely to change l3doc and you may wish to produce a modified class rather than extend l3doc in the core release.

PR #1622, opened 2024-11-26
Add wrapper functions for regex concepts

josephwright:
The naming here keeps the `regex...` part together. I've not included all of the `TF` versions from the regex module as they don't directly match up with how things are done in the target modules. I've covered what seem like the most useful variants, including global versions as they match the nearest-equivalents in the target modules.

PR #1756, opened 2025-06-09
Move l3str-format to stable

josephwright:


l3kernel/l3str-format.dtx
```diff
@@ -90,7 +87,7 @@
 %
 % \section{Formatting various data-types}
 %
-% \begin{function}[EXP]{\tl_format:Nn, \tl_format:cn, \tl_format:nn}
+% \begin{function}[EXP, added = 2025-06-09]{\tl_format:Nn, \tl_format:cn, \tl_format:nn}
```
muzimuzhi: Perhaps the docs for `\<type>_format:..` functions can be moved to the corresponding `l3<type>` module, like we did in
- #1622

For example, the doc for `\tl_format:(N|n)n` can be moved to `l3tl` module, under the subsection "Using token lists" (or under a new subsection titled "Formatting token lists").

The implementation looks centric, thus can be left in `l3str-format.dtx` and this module would be another implementation-only module.

josephwright: I've kept the general bit about formats: this is therefore similar to `l3sort`.

FrankMittelbach (approved):

l3kernel/l3fp.dtx
```diff
@@ -392,6 +392,32 @@
 %   This function is identical to \cs{fp_to_decimal:N}.
 % \end{function}
 %
+% \section{Formatting floating points}
+%
+% \begin{function}[EXP, added = 2025-06-09]{\fp_format:nn}
+%   \begin{syntax}
+%     \cs{fp_format:nn} \Arg{fp expr} \Arg{format specification}
+%   \end{syntax}
+%   Evaluates the \meta{floating point expression} and converts the
```
muzimuzhi: Perhaps
```suggestion
%   Evaluates the \meta{fp expr} and converts the
```

l3kernel/l3int.dtx
```diff
@@ -832,6 +832,19 @@
 %   expands to |mmmmmmmmccxlix|.
 % \end{function}
 %
+% \begin{function}[EXP, added = 2025-06-09]{\int_format:nn}
+%   \begin{syntax}
+%     \cs{int_format:nn} \Arg{intexpr} \Arg{format specification}
```
muzimuzhi:
```suggestion
%     \cs{int_format:nn} \Arg{int expr} \Arg{format specification}
```

l3kernel/l3int.dtx
```diff
@@ -832,6 +832,19 @@
...
+%   \end{syntax}
+%   Evaluates the \meta{integer expression} and converts the result to a
```
muzimuzhi: Perhaps
```suggestion
%   Evaluates the \meta{int expr} and converts the result to a
```

l3kernel/l3seq.dtx
```diff
@@ -917,6 +917,16 @@
 %   \end{texnote}
 % \end{function}
 %
+% \begin{function}[EXP, added = 2025-06-09]{\seq_format:Nn, \seq_format:cn}
+%   \begin{syntax}
+%     \cs{seq_format:Nn} \Arg{sequence} \Arg{format specification}
```
muzimuzhi:
```suggestion
%     \cs{seq_format:Nn} \meta{seq~var} \Arg{format specification}
```

l3kernel/l3seq.dtx
```diff
@@ -917,6 +917,16 @@
...
+%   \end{syntax}
+%   Converts each item in the \meta{sequence} to a string according to
+%   the \meta{format specification}, and concatenates the results.
```
muzimuzhi: I added "as a token list", so readers know which `style`s are supported. May need a better wording.

Also may need to explain the behavior of "concatenates the results" in more detail (like `\seq_use:Nn <seq var> {}`?).

BTW is the behavior of "concatenates the results" really useful?

```suggestion
%   Converts each item in the \meta{seq~var} as a token list to a string according to
%   the \meta{format specification}, and concatenates the results.
```

Each item is formatted as a token list according to the
%   specification.

l3kernel/l3tl.dtx
```diff
@@ -1148,6 +1148,21 @@
 %   \end{texnote}
 % \end{function}
 %
+% \subsection{Formatting token lists}
+%
+% \begin{function}[EXP, added = 2025-06-09]{\tl_format:Nn, \tl_format:cn, \tl_format:nn}
+%   \begin{syntax}
+%     \cs{tl_format:nn} \Arg{token list} \Arg{format specification}
```
muzimuzhi, line 1156: Add syntax for `\tl_format:Nn`.

```suggestion
%     \cs{tl_format:Nn} \meta{tl~var} \Arg{format specification}
%     \cs{tl_format:nn} \Arg{token list} \Arg{format specification}
```

l3kernel/l3tl.dtx
```diff
@@ -1148,6 +1148,21 @@
...
+%   \end{syntax}
+%   Converts the \meta{token list} to a string according to the
```
muzimuzhi:
```suggestion
%   Converts the \meta{tl~var} or \meta{token list} to a string according to the
```

l3kernel/l3fp.dtx
```diff
@@ -392,6 +392,32 @@
 %   This function is identical to \cs{fp_to_decimal:N}.
 % \end{function}
 %
+% \section{Formatting floating points}
+%
+% \begin{function}[EXP, added = 2025-06-09]{\fp_format:nn}
+%   \begin{syntax}
+%     \cs{fp_format:nn} \Arg{fp expr} \Arg{format specification}
```
muzimuzhi, line 399: _Do not accept it. This suggestion was wrong!_

```suggestion
%     \cs{fp_format:nn} \meta{fp~expr} \Arg{format specification}
```

muzimuzhi (changes requested):
Improve consistency in argument names. Extend doc for `\str_format:nn`.

Started a discussion about the "concatenates the results" behavior of `\str_format:nn`.

On more `N`-type functions:

Since this adds both `\tl_format:Nn` and `\tl_format:nn`, do we also want to add `\fp_format:Nn` and `\int_format:Nn`? And, how about `\str_format:(N|n)n`?

l3kernel/l3fp.dtx
mbertucci47: This one should still be `\Arg` no?

muzimuzhi: You're right. The last-minute suggestion is usually unreliable :).

josephwright:
> > On more `N`-type functions:
> > Since this adds both `\tl_format:Nn` and `\tl_format:nn`, do we also want to add `\fp_format:Nn` and `\int_format:Nn`? 

Usually we don't has you can always use an `int`/`fp` inside `\<thing>_<fn>:n`

> And, how about `\str_format:(N|n)n`?

Maybe I guess - outside the PR

PR #1757, opened 2025-06-13
 Add "set" into function names for adjusting drawing state

josephwright:


l3experimental/CHANGELOG.md
```diff
@@ -151,7 +155,7 @@ this project uses date-based 'snapshot' version identifiers.
 ## [2021-08-04]
 
 ### Added
-- `\draw_baseline:n`
+- `\draw_set_baseline:n`
```
muzimuzhi: Revert this one? What added in 2021-08-04 release was indeed `\draw_baseline:n`, without `set`.

```suggestion
- `\draw_baseline:n`
```

l3experimental/l3draw/l3draw-state.dtx
```diff
@@ -85,23 +85,23 @@
...
-% \begin{macro}{\draw_linewidth:n}
+% \begin{macro}{\draw_set_linewidth:n}
 %   Set the linewidth: we need a wrapper as this has to pass to the driver
 %   layer.
 %    \begin{macrocode}
-\cs_new_protected:Npn \draw_linewidth:n #1
+\cs_new_protected:Npn \draw_set_linewidth:n #1
   {
     \dim_gset:Nn \g_@@_linewidth_dim { \fp_to_dim:n {#1} }
     \@@_backend_linewidth:n \g_@@_linewidth_dim
```
muzimuzhi, line 95: For consistency, maybe the corresponding backend functions (if exist) need a `set` in their names too?

l3experimental/l3draw/l3draw-state.dtx
```diff
@@ -120,30 +120,30 @@
...
+\cs_new_protected:Npn \draw_set_miterlimit:n #1
   { \exp_args:Ne \@@_backend_miterlimit:n { \fp_eval:n {#1} } }
 %    \end{macrocode}
 % \end{macro}
 %
 % \begin{macro}
 %   {
-%     \draw_cap_butt:, \draw_cap_rectangle:, \draw_cap_round:,
+%     \draw_set_cap_butt:, \draw_set_cap_rectangle:, \draw_set_cap_round:,
 %     \draw_evenodd_rule:, \draw_nonzero_rule:,
```
muzimuzhi: Why these two dont't need `set`: `\draw_(evenodd|nonezero)_rule:` vs `\draw_set_(evenodd|nonezero)_rule:`.

l3experimental/l3draw/l3draw.dtx
muzimuzhi, line 1: Two omissions

```
l3draw/l3draw.dtx:282:%     \cs{draw_dash_pattern:nn} \Arg{pattern} \Arg{phase}
l3draw/l3draw.dtx:355:%     \cs{draw_miterlimit:n} \Arg{factor}
```

found by
```
rg --no-heading 'draw_(base|cap|dash|join|line|miter|\()'`
# or the stricter and fancier
rg --no-heading --pcre2 '(?<!_)draw_(base|cap|dash|join|line|miter|\()' -g '!CHANGELOG.md'
```

muzimuzhi (changes requested):

l3experimental/CHANGELOG.md
josephwright: Ah, yes

l3experimental/l3draw/l3draw-state.dtx
FrankMittelbach: they are drawing not setting something, I think that that's ok (aka correct)

muzimuzhi: They are _setting_ the interior rule. The rule is then used by following `\draw_path_use:n { fill }`.
image

Update: According to the PDF Reference, in PDF layer the two rules are distinguished by different operators (`f` for nonzero and `f*` for even-odd), thus the current active rule is not part of the graphics state. Probably it's `pgf` that factorizes it as a state first.

FrankMittelbach: in that case you are right and it should be "set"

l3experimental/l3draw/l3draw-state.dtx
josephwright: Yes eventually - but that means changes in `l3backend`, which I then have to work out how to time - once the code moves to `l3kernel` this is easier. So 'for later'.

muzimuzhi (approved):

muzimuzhi:
Minor addition: As now "set" is added into all function names documented under section "Graphics state", maybe it's a chance to also rename that section to "Setting graphics state".

https://github.com/latex3/latex3/blob/585dbdb390ae4766770bfa51aae200b8ffeb0481/l3experimental/l3draw/l3draw.dtx#L261-L265

Also, in "some are rest at the start of each drawing", is "rest" a typo for "reset"?

PR #1761, opened 2025-06-16
Add \tl_retokenize:n

josephwright:


josephwright:
Thoughts on the name welcome: I'm aiming for different-from-rescan as this is 'dynamic', but wasn't sure that was the best term. (Context: I need this for `ltx-talk`, really want it for documenting `l3draw` and more generally it 'goes with` the new `c`-type grabber in `ltcmd`.)

muzimuzhi:
`\tl_rescan_dynamic:n` or `\tl_dynamic_rescan:n`?

l3kernel/l3tl.dtx
```diff
@@ -2213,6 +2229,20 @@
...
 %
+% \begin{macro}{\tl_retokenize:n, \tl_retokenize:V}
+%   A thin wrapper from \tn{scantokens}.
+%    \begin{macrocode}
+\cs_gset_protected:Npn \tl_retokenize:n #1
+  {
+    \group_begin:
+      \int_set:Nn \tex_newlinechar:D { 13 }
+    \exp_after:wN \group_end:
+    \tex_scantokens:D \exp_after:wN { \tl_to_str:n {#1} }
```
Skillmon: What happened to the `\scantextokens` idea? Should we try to add something that will (most likely) remove the trailing space? If not, the fact that if this is used in horizontal mode it'll have a trailing space in output should be documented!

josephwright: I couldn't find a way to do the same as `\scantextokens` reliably in pdfTeX - ideas welcome

Skillmon: Exactly the same behaviour won't be possible (the grabbing additional arguments for instance). The space-suppression is possible, but only sort of unreliable. Provided the argument doesn't contain a NUL char (or any other arbitrary char), and doesn't change the category code of that character (the latter will be impossible to check, so we have to hope for the best here), you can put `\char_set_catcode_comment:n { 0 }` in the setup code and put a catcode 12 NUL char at the end of the `\tex_scantokens:D` argument.

Skillmon: Or persuade someone (not me, I've given up on grasping WEB) to back port it...

muzimuzhi:
Is `\scantokens` supported by LuaMetaTeX?

The LuaMetaTeX version of `\tl_(g)set_rescan:Nnn` [says](https://github.com/latex3/latex3/blob/ad0b8a30b0284321a4434b4f8ef71771650c3dcc/l3kernel/l3tl.dtx#L2134-L2136)

> in LuaMetaTeX there is no \scantokens so something different is needed.

but running the following ConTeXt example shows that `\scantokens` still exists in LuaMetaTeX 2.11.07.

```tex
\starttext
\show\scantokens
\stoptext
```

```
> \scantokens=global primitive \scantokens
...
mkxl run stats  > used engine: luametatex version: 2.11.07, functionality level: 20250609, format id: 720, compiler: clang
mkxl run stats  > tex properties: 938216 hash slots used of 2097152, 52562 control sequences, approximate memory usage: 113 MB
mkxl run stats  > lua properties: engine: lua 5.5, format 7, used memory: 57 MB, ctx: 54 MB, max: 81 MB, symbol mask: utf (τεχ)
```

l3kernel/l3tl.dtx
josephwright: I've looked at a back-port - quite a bit of work and not a high priority.

josephwright:
I've updated to allow for `\scantextokens`.

l3kernel/l3tl.dtx
```diff
@@ -2213,6 +2229,29 @@
...
 % \end{macro}
 %
+% \begin{macro}{\tl_retokenize:n, \tl_retokenize:V}
+%   A thin wrapper from \tn{scantokens} or \tn{scantextokens}: the
+%   trailing |^^J| ensures we get the \emph{same} behavior from both.
+%    \begin{macrocode}
+\cs_new_protected:Npe \tl_retokenize:n #1
+  {
+    \group_begin:
+      \exp_not:N \int_set:Nn \exp_not:N \tex_newlinechar:D { 13 }
```
muzimuzhi: `\int_set:Nn` is engine-protected, so these two `\exp_not:N` are both optional.
```suggestion
      \int_set:Nn \tex_newlinechar:D { 13 }
```

josephwright: There's a timing isn't - `\int_set:Nn` isn't defined yet - and for variables we routinely add `\exp_not:N` for consistency (as we need for `tl`, `fp`, _etc._).

josephwright:
I guess I could ask Hans about `\scantokens` in LuaMetaTeX ...

josephwright:
Screenshot 2025-06-18 at 07 55 43

l3kernel/l3tl.dtx
muzimuzhi: > There's a timing isn't - `\int_set:Nn` isn't defined yet

Oh, I see!

l3kernel/l3tl.dtx
```diff
@@ -2213,6 +2229,23 @@
 % \end{macro}
 % \end{macro}
 %
+% \begin{macro}{\tl_rescan_dynamic:n, \tl_rescan_dynamic:V}
+%   A thin wrapper from \tn{scantokens}. At present, we are told by
+%   Hans that LuaMeta\TeX{} will retain this primitive, so we should be
+%   OK for all engines.
+%    \begin{macrocode}
+\cs_gset_protected:Npn \tl_rescan_dynamic:n #1
```
muzimuzhi: Should be `new` then.
```suggestion
\cs_set_protected:Npn \tl_rescan_dynamic:n #1
```

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,22 @@
...
 %
+% \begin{function}[added = 2025-06-16]{\tl_rescan_dynamic:n, \tl_rescan_dynamic:V}
+%   \begin{syntax}
+%     \cs{tl_rescan_dynamic:n} \Arg{tokens}
+%   \end{syntax}
+%   Retokenizes the \meta{tokens} by applying the current category code
+%   régime. In contrast to \cs{tl_rescan:nn}, this function retokenizes
+%   the input \enquote{dynamically}: category code changes which occur
+%   whilst processing the retokenized material \emph{will} apply to later
+%   content. The \emph{tokens} are always treated as ending with a space.
```
car222222: Is it clear what is meant by this sentence?

l3kernel/l3tl.dtx
```diff
@@ -2213,6 +2229,23 @@
 % \end{macro}
 % \end{macro}
 %
+% \begin{macro}{\tl_rescan_dynamic:n, \tl_rescan_dynamic:V}
+%   A thin wrapper from \tn{scantokens}. At present, we are told by
```
car222222, line 2236: from  ==>  for

car222222 (commented):
Some wording queries.

l3kernel/l3tl.dtx
josephwright: This is deliberate - as we are wanting a global def (yes, i know we should be at top-level but if you look at `new` it works globally).

josephwright: This is deliberate - as we are wanting a global def (yes, i know we should be at top-level but if you look at `new` it works globally).

josephwright:
> As `\tl_rescan_dynamic:n` is protected, do we also need `\tl_(g)set_rescan_dynamic:Nn`?

No, the whole issue with `\scantokens` is that isn't really workable - a 'dynamic' interface which is only usable in typesetting contexts is _necessary_ if a bit awkward. (It's this or I have to stick to a classical `beamer`-like write-and-read-back approach).

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,22 @@
 %   \end{texnote}
 % \end{function}
 %
+% \begin{function}[added = 2025-06-16]{\tl_rescan_dynamic:n, \tl_rescan_dynamic:V}
+%   \begin{syntax}
+%     \cs{tl_rescan_dynamic:n} \Arg{tokens}
+%   \end{syntax}
```
FrankMittelbach: Reading the command name I have no idea what it could mean.

I would have some idea if it would say `_dynamically`
or mabye `_gradationally` or mabye just `_stepwise`.

but perhaps that's only me

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,22 @@
...
+%   Retokenizes the \meta{tokens} by applying the current category code
+%   régime. In contrast to \cs{tl_rescan:nn}, this function retokenizes
+%   the input \enquote{dynamically}: category code changes which occur
+%   whilst processing the retokenized material \emph{will} apply to later
```
FrankMittelbach: I don't see how catcode changes can happen while processing. The code below shows everything is first turned into a string, ie catcode 12 throughout and then \scantokens is applied to that, so where is this "dynamic stuff" happening?

davidcarlisle: it acts like `\scantokens`  where ``\def\x{\catcode`X=7 \show X} \scantokens\expandafter{\x}``  shows the X has catcode 7, in contrast to `\tl_rescan:nn` which (for whatever reason)  rescans but process the result as a macro argument before executing it so the catcode change wouldn't affect the catcode of the rescanned X (but "dynamic" isn't a good description for this)

FrankMittelbach: ah, so not enough context visiblefor me, the stuff is not assigned to a tl but placed into the input stream (should have realized this from the signature). Anyway, that still makes me feel that the names are wrong (but also the original "rescan") The new one is, as you suggested really a retokenize and the one called _rescan is really _recatcode because nothing is really scanned anew there but instead all the catcodes are changes and *then* it is processed while with the new one we have retokenization while (re)processing it

josephwright:
The aim here is to find some interface that's workable to the 'raw' behaviour of `\scantokens`. What we have for `\tl_set_rescan:nn` and `\tl_rescan:nn` is set up to deal with the fact that `\scantokens` is 'tricky', but unless we want to say 'tough, you always have to go via a file', then some form of 'raw' interface is needed - and people do that. So to avoid `\tex_scantokens:D`, I'd like to find _some_ name we can go with.

FrankMittelbach:
> The aim here is to find some interface that's workable to the 'raw' behaviour of `\scantokens`. What we have for `\tl_set_rescan:nn` and `\tl_rescan:nn` is set up to deal with the fact that `\scantokens` is 'tricky', but unless we want to say 'tough, you always have to go via a file', then some form of 'raw' interface is needed - and people do that. So to avoid `\tex_scantokens:D`, I'd like to find _some_ name we can go with.

Well it is retokenization  while you process the input and more than just changing to preset catcodes (and you initially suggested that name), David prefers it too and I too find it much more understandable than `rescan_dynamic`. I'm not sure we need to have "rescan" in the name but if we want that then perhaps `_stepwise` or `_fully` but I think `\tl_retokenize:nn`is just fine.

However, I think the documentation of `\tl_set_rescan:Nnn` and `\tl_rescan:nn`should explicitly mention that catcode changes within the tokens of argument do not apply because they aren't set (or seen) during the rescanning. And then you can nicely explain what the `\lt_retokenize:nn` does differently.

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,22 @@
 %   \end{texnote}
 % \end{function}
 %
+% \begin{function}[added = 2025-06-16]{\tl_rescan_dynamic:n, \tl_rescan_dynamic:V}
+%   \begin{syntax}
+%     \cs{tl_rescan_dynamic:n} \Arg{tokens}
+%   \end{syntax}
+%   Retokenizes the \meta{tokens} by applying the current category code
+%   régime. In contrast to \cs{tl_rescan:nn}, this function retokenizes
+%   the input \enquote{dynamically}: category code changes which occur
```
car222222: "dynamically" would be OK here, but best to follow it with:
. . . in the sense that category code . . .

QUESTION: are all other assignments, definitions, etc. also done "dynamically":
If they all are, then need to explain this too, surely? 
If not, then this certainly needs a lot more explantation. 

BUT it does not follow that "dynamic" is also optimal as part of the name.

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,22 @@
...
+%   whilst processing the retokenized material \emph{will} apply to later
+%   content. The \emph{tokens} are always treated as ending with a space.
+%   \begin{texnote}
+%     This is a thin wrapper around the \tn{scantokens} primitive which
+%     resets the value of \tn{newlinechar} to~$13$ and detokenizes the
```
car222222: This "which" here seems totally ambiguous: need to be clear what is the difference between these two and which does what. 

Maybe also need here a full explanation of what \scantokens does? Or at least a reference to where it is fully documented. 
Having read the whole of this, I think it is essential to explain how \scantokns works.
And why here it is necessary to set (only) the "newline character".

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,22 @@
...
+%     input, but which otherwise follows the primitive behavior.
```
car222222: I think you mean:

. . . the behavio(u)r of the primitive. 

Assuming it is clear what "the primitive" refers to here. 

"follows" may also be problematic for some readers?

josephwright:
Yeah but
```latex
\long\def\myscan#1{\scantokens\expandafter{\detokenize{#1}\noexpand}}
\catcode`/=0
/catcode`\\=12
/edef/x{/myscan{abc /tenbf x}}/show/x
```

car222222 (commented):
Some queries and suggestions on the text.

car222222:
Since it is so very close to \scantokens, maybe the name should reflect this:
Something like:

car222222:
Are these catcodes the only things that can be "dynamically" changed within a use of  \scantokens?

car222222:
For example: could there be a primitive \def in there, that dynamically (re)defines a csname?

davidcarlisle:
> Yeah but
> 
> ```
> \long\def\myscan#1{\scantokens\expandafter{\detokenize{#1}\noexpand}}
> \catcode`/=0
> /catcode`\\=12
> /edef/x{/myscan{abc /tenbf x}}/show/x
> ```

that looks depressingly familiar:-)  `\scantextokens` has some appeal....

l3kernel/l3tl.dtx
josephwright: I've set the newline char to 13 as this matches the normal endline char for e.g. the `ltcmd` `c`-type argument - this seems by far the most useful value. If that is not 'popular', we could make this a second argument, or of curse could provide an interface to set `\endlinechar` and`\newlinechar`.

josephwright:
@car222222  I've updated the docs based on your feedback - hope this is now clearer.

davidcarlisle:
> On the `\newlinechar` business, perhaps we want a second arg to cover that as it's a bit odd?
> 
> ```
> \tl_retokenize:nn { <line-end-char> } { <tokens> }
> ```
> 
> used e.g.
> 
> ```
> \tl_retokenize:nn { `\^^M } { .... }
> ```
> 
> perhaps?

13 is a bit odd though in the normal case when it's _not_ verbatim, if there were such an argument (or with the current code that sets it to 13 internally) perhaps the newlinechar should be locally catcode 12? 

In the current code its 13 but by default that's catcode 5 so `\tl_whatever:n{abc^^Mxyz}`  comments out the rest of the line so it's not clear how the `newline` setting  of 13 is intended to be used. The usual latex setting of 10 ^^J is perhaps easier to understand,

davidcarlisle:
> @davidcarlisle The key I guess is a predictable value - happy to adjust - pick a value :)

depends on the use cases but plain TeX's default of -1 has some benefits here, or the usual latex value of 10 would be my choices, I think.  -1 has some appeal unless you have a specific use case in mind where you need a newlinechar here. The fact that newlinechar has any effect at all seems to be a side effect of the internal implementation as a write followed by a read, so not something I'd really highlight.

josephwright:
> > @davidcarlisle The key I guess is a predictable value - happy to adjust - pick a value :)
> 
> depends on the use cases but plain TeX's default of -1 has some benefits here, or the usual latex value of 10 would be my choices, I think. -1 has some appeal unless you have a specific use case in mind where you need a newlinechar here. The fact that newlinechar has any effect at all seems to be a side effect of the internal implementation as a write followed by a read, so not something I'd really highlight.

If it's set to -1, then you can't rescan into lines - that would be an issue for verbatim (unless I miss something) - try fiddiling about in `ltx-talk` I guess :)

josephwright:
OK, let's see if people are happy now ...

l3kernel/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ this project uses date-based 'snapshot' version identifiers.
 ## [Unreleased]
 
 ### Added
+ `\tl_retokenize:n`
```
muzimuzhi:
```suggestion
- `\tl_retokenize:n`
```

muzimuzhi (commented):
Need some tests? (Sorry if it's already discussed before. I didn't read through the discussion.)

Currently `\tl_(g)set_rescan:Nnn` is tested in both `m3tl008.lvt` and `m3tl009.lvt` (for `\halign` robustness)`

davidcarlisle:
> Need some tests? (Sorry if it's already discussed before. I didn't read through the discussion.)
> 
> Currently `\tl_(g)set_rescan:Nnn` is tested in both `m3tl008.lvt` and `m3tl009.lvt` (for `\halign` robustness)`

It will need some more tests, but probably more tweaks to the implementation first, probably we can't get away with disabling newlinechar with -1 and need to go back to 10 (`^^J`)

some experiments with how it interacts with `c` argument grabbing....

```
\documentclass{article}

\ExplSyntaxOn
\group_begin:
  \cs_gset_protected:Npn \tl_retokenize:n #1 %
  { %
    %\def\x{#1}\show\x
      \group_begin: %
        \int_set:Nn \tex_newlinechar:D { 10 } %
        \int_set:Nn \tex_endlinechar:D { 13 } %
      \exp_after:wN \group_end: %
      \tex_scantokens:D \exp_after:wN %
        { \tl_to_str:n {#1} } %
    } %
\group_end: %
\cs_generate_variant:Nn \tl_retokenize:n { V }
\cs_set_eq:NN \xxretokenize \tl_retokenize:n
\ExplSyntaxOff
\begin{document}

\makeatletter
\NewDocumentEnvironment{foo}{c}{\ttfamily
  \let\do\@makeother\dospecials\obeylines
   \edef\obeyedline{\UseName{char_generate:nn}{10}{12}}%
  \ExpandArgs{e}\xxretokenize{#1}}{}%
\NewDocumentEnvironment{foo2}{c}{\ttfamily
  % \let\do\@makeother\dospecials\obeylines
  \edef\obeyedline{\UseName{char_generate:nn}{10}{12}}%
  \edef\tmpa{#1}%
  \def\obeyedline{\par}%
  \ExpandArgs{e}\xxretokenize{\tmpa}}{}%
\makeatother
\begin{foo}
#wo{o
bong
%4&\bong
\end{foo}


\bigskip


\begin{foo2}
\begin{verbatim}
#wo{o
bong

%4&\bong
\end{verbatim}

stuff

more stuff
\end{foo2}


\end{document}
```

l3kernel/l3tl.dtx
```diff
@@ -1419,6 +1419,25 @@
 %   \end{texnote}
 % \end{function}
 %
+% \begin{function}[added = 2025-07-08]{\tl_retokenize:n, \tl_retokenize:V}
+%   \begin{syntax}
+%     \cs{tl_retokenize:n} \Arg{tokens}
+%   \end{syntax}
+%   Retokenizes the \meta{tokens} by applying the current category code
+%   régime. In contrast to \cs{tl_rescan:nn}, this function executes the
+%   \meta{tokens} as the updated tokens are generated. As such, any
```
Skillmon:
```suggestion
%   \meta{tokens} as the category codes are reassigned to the tokens. As such, any
```

PR #1762, opened 2025-06-18
Update `\tl_trim_spaces:n` comment

muzimuzhi:
To sync with its new, speed-up implementation, see 0fdf1714 (speed up trim spaces (#<span></span>1679), 2025-02-04).


Skillmon:
The code is documentation enough! (sorry, I obviously missed those comments back then)

PR #1765, opened 2025-06-24
Add mech. to exclude words from case changing

josephwright:


l3kernel/l3text-case.dtx
```diff
@@ -909,6 +958,27 @@
 % \end{macro}
 % \end{macro}
 %
+% \begin{macro}
+%   {
+%     \text_declare_lowercase_exclusion:nn ,
+%     \text_declare_titlecase_exclusion:nn ,
+%     \text_declare_uppercase_exclusion:nn
```
muzimuzhi:
```suggestion
%     \text_declare_lowercase_exclusion:n ,
%     \text_declare_titlecase_exclusion:n ,
%     \text_declare_uppercase_exclusion:n
```

muzimuzhi (changes requested):
Only reviewed the non-code part.

josephwright:
Currently I'm still using a two-pass approach: find words to protect from case changing, mark them up, then loop over all the input to case change. One issue there is that at present something like
```latex
\text_words_map_function:nN { Some~text~\cite { WithFun } } \tl_show:n
```
splits `\cite` from its argument - because it was not really obvious how to handle commands in word breaking. I could I guess alter to do something like treat as one 'word' if there is a directly-following brace group - thoughts?

josephwright:
(I could then look at using a more integrated approach to the word boundaries for things like the final sigma rule, titlecasing only the first word, etc.)

l3kernel/l3text-case.dtx
```diff
@@ -395,10 +426,19 @@
   {
     \@@_if_q_recursion_tail_stop_do:Nn #5
       { \@@_change_case_break: }
+    \str_if_eq:nnTF {#5} { \@@_change_case_exclude_word:n }
```
Skillmon, line 429: Do we have a possibility to use a faster test than `\str_if_eq:nnTF`? Can the marker be a quark? If so an `\ifx`-based test would be preferable, if I skimmed over its usages correctly it should never actually be expanded, so its meaning can be arbitrary.

Skillmon (approved):
Changelog entry is missing. Otherwise code looks fine.

No idea, yet, of a good optimisation worth being implemented.

l3kernel/l3text-case.dtx
josephwright: I'd thought about this a bit - whilst I see your point, we are not in a place where performance at this level of detail is going to matter (I think) - and we know that quarks have a tendency to escape even though 'they will never be expanded' :)

PR #1769, opened 2025-07-02
Add support for key value expansion

josephwright:
This is part one of addressing https://github.com/latex3/latex2e/issues/1801 - I also need to do the template code in the 2e repo.

I've chosen to go for the 'higher level' `l3keys` rather than `\keyval_parse:nnn`, as the latter is meant to be entirely generic, so to me feels like expansion control is down the programmer there.

I expect to write something for `ltnews` but will do that as part of the PR for template code; I may also look at adding something to `usrguide` at that point (but that's more tricky as this doesn't affect _all_ keyvals provided by the base distro. most obviously the classical `keyval` pkg).

I've gone for performance over allowing `:` in key names - I hope this minor restriction is not too much of an issue as a braking change.

### DIFF
--- a/l3kernel/l3keys.dtx
+++ b/l3kernel/l3keys.dtx
@@ -106,7 +106,7 @@
 %     }
 % \end{verbatim}
 %
-% Key names may contain any tokens, as they are handled internally
+% Key names may contain any tokens except |:|, as they are handled internally
 % using \cs{tl_to_str:n}. As discussed in
 % section~\ref{sec:l3keys:subdivision}, it is suggested that the character
 % |/| is reserved for sub-division of keys into different subsets.
@@ -773,6 +773,23 @@
 %   empty if no value was given. This is stored in \cs{l_keys_value_tl}, and
 %   is not processed in any way by \cs{keys_set:nn}.
 % \end{variable}
+%
+% \subsection{Expanding the values of keys}
+%
+% To allow the user to apply value expansion at point-of-use, key names can be
+% followed by an expansion specifier. This is given by appending |:| and a
+% single letter specifier to the key name. The latter are the normal argument
+% specifiers for \pkg{expl3}, thus may be one of \texttt{n} (redundant but
+% supported), \texttt{o}, \texttt{V}, \texttt{v} or \texttt{e}. Thus for example
+% \begin{verbatim}
+%   \tl_set:Nn \l_mymodule_value_tl
+%   \keys_set:nn { mymodule } { key = value }
+%   \keys_set:nn { mymodule } { key : o = \l_mymodule_value_tl }
+%   \keys_set:nn { mymodule } { key : V = \l_mymodule_value_tl }
+%   \keys_set:nn { mymodule } { key : v = { l_mymodule_value_tl } }
+%   \keys_set:nn { mymodule } { key : e = \exp_not:V \l_mymodule_value_tl }
+% \end{verbatim}
+% all pass \texttt{value} to the key handler.
 %
 % \section{Handling of unknown keys}
 % \label{sec:l3keys:unknown}
@@ -1603,6 +1620,13 @@
 %    \begin{macrocode}
 \int_new:N \l_keys_choice_int
 \tl_new:N \l_keys_choice_tl
+%    \end{macrocode}
+% \end{variable}
+%
+% \begin{variable}{\l_@@_exp_str}
+%   The expansion postfix of a key name.
+%    \begin{macrocode}
+\str_new:N \l_@@_exp_str
 %    \end{macrocode}
 % \end{variable}
 %
@@ -3104,7 +3128,8 @@
 %     \@@_find_key_module_auxi:Nw   ,
 %     \@@_find_key_module_auxii:Nw  ,
 %     \@@_find_key_module_auxiii:Nn ,
-%     \@@_find_key_module_auxiv:Nw
+%     \@@_find_key_module_auxiv:NNw ,
+%     \@@_find_key_module_auxiv:NNw
 %   }
 % \begin{macro}{\@@_set_selective:}
 %   A shared system once again. First, set the current path and add a
@@ -3141,6 +3166,12 @@
     \str_clear:N \l_@@_inherit_str
     \exp_after:wN \@@_find_key_module:wNN \l_keys_path_str \s_@@_stop
       \l_@@_module_str \l_keys_key_str
+    \str_set:Ne \l_keys_path_str
+      {
+        \l_@@_module_str 
+        \str_if_empty:NF \l_@@_module_str { / }
+        \l_keys_key_str
+      }
     \tl_set_eq:NN \l_keys_key_tl \l_keys_key_str
     \@@_value_or_default:n {#3}
     \bool_if:NTF \l_@@_selective_bool
@@ -3152,9 +3183,10 @@
 %    \end{macrocode}
 %   This function uses \cs{cs_set_nopar:Npe} internally for performance reasons,
 %   the argument |#1| is already a string in every usage, so turning it into a
-%   string again seems unnecessary.
+%   string again seems unnecessary. The expansion part of a key is always stored
+%   in the same place, so that is not passed along.
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_find_key_module:wNN #1 \s_@@_stop #2 #3
+\cs_new_protected:Npn \@@_find_key_module:wNN #1 \s_@@_stop #2#3
   {
     \@@_find_key_module_auxi:Nw #2 #1 \s_@@_nil \@@_find_key_module_auxii:Nw
       / \s_@@_nil \@@_find_key_module_auxiv:Nw #3
@@ -3174,11 +3206,21 @@
     \cs_set_nopar:Npe #1 { \tl_if_empty:NF #1 { #1 / } #2 }
     \@@_find_key_module_auxi:Nw #1
   }
-\cs_new_protected:Npn \@@_find_key_module_auxiv:Nw
+\cs_new_protected:Npe \@@_find_key_module_auxiv:Nw
     #1 #2 \s_@@_nil #3 \s_@@_mark
     \s_@@_nil \@@_find_key_module_auxiv:Nw #4
   {
-    \cs_set_nopar:Npn #4 { #2 }
+    \exp_not:N \@@_find_key_module_auxv:Nw #4
+    #2 \token_to_str:N : n \token_to_str:N : \s_@@_mark
+  }
+\use:e
+  {
+    \cs_new_protected:Npn \exp_not:N \@@_find_key_module_auxv:Nw
+      #1 #2 \token_to_str:N : #3 \token_to_str:N : #4 \s_@@_mark
+  }
+  {
+    \cs_set_nopar:Npn #1 {#2}
+    \cs_set_nopar:Npn \l_@@_exp_str {#3}
   }
 %    \end{macrocode}
 %  If selective setting is active, there are a number of possible sub-cases
@@ -3263,7 +3305,14 @@
               { \@@_default_inherit: }
           }
       }
-      { \tl_set:Nn \l_keys_value_tl {#1} }
+      {
+        \cs_if_exist_use:cF { tl_set:N \l_@@_exp_str }
+          {
+            \msg_error:nnV { keys } { unknown-expansion } \l_@@_exp_str
+            \use_none:nn
+          }
+          \l_keys_value_tl {#1}
+      }
   }
 \cs_new_protected:Npn \@@_default_inherit:
   {
@@ -3725,6 +3774,12 @@
   {
     The~module~'#2'~does~not~have~a~key~called~'#1'.\\
     Check~that~you~have~spelled~the~key~name~correctly.
+  }
+\msg_new:nnnn { keys } { unknown-expansion }
+  { The~value~expansion~'#1'~is~unknown. }
+  {
+    Key~values~can~only~be~expanded~using~one~of~the~pre-defined~methods:~
+    n,~o,~V,~v~or~e.
   }
 \msg_new:nnnn { keys } { nested-choice-key }
   { Attempt~to~define~'#1'~as~a~nested~choice~key. }
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Added
+- Support for key value expansion in `l3keys`
+
 ## [2025-06-30]
 
 ### Added
--- a/l3kernel/testfiles/m3keys001.lvt
+++ b/l3kernel/testfiles/m3keys001.lvt
@@ -267,6 +267,22 @@
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+\TEST { Expanding~values }
+  {
+    \keys_define:nn { module }
+      {
+        key-one .code:n = { \TYPE { "#1" } }
+      }
+    \tl_set:Nn \l_tmpa_tl { value }
+    \keys_set:nn { module } { key-one = value }
+    \keys_set:nn { module } { key-one : o = \l_tmpa_tl }
+    \keys_set:nn { module } { key-one : V = \l_tmpa_tl }
+    \keys_set:nn { module } { key-one : v = { l_tmpa_tl } }
+    \keys_set:nn { module } { key-one : e = \l_tmpa_tl \l_tmpa_tl \l_tmpa_tl }
+  }
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
 \TEST { Unknown~keys }
   {
     \keys_define:nn { module }
@@ -306,6 +322,16 @@
         key-two .code:n ,
         key-three
       }
+  }
+
+\TEST { Unknown~expansion }
+  {
+    \keys_define:nn { module }
+      {
+        key-one .code:n = { \TYPE { "#1" } }
+      }
+    \keys_set:nn { module } { key-one : q = \l_tmpa_tl }
+    \keys_set:nn { module } { key-one : nn = \l_tmpa_tl }
   }
 
 \TEST { Known~property~but~missing~argument }
--- a/l3kernel/testfiles/m3keys001.tlg
+++ b/l3kernel/testfiles/m3keys001.tlg
@@ -107,7 +107,17 @@ Check that you have spelled the key name correctly.
 "value"
 ============================================================
 ============================================================
-TEST 9: Unknown keys
+TEST 9: Expanding values
+============================================================
+Defining key module/key-one on line ...
+"value"
+"value"
+"value"
+"value"
+"valuevaluevalue"
+============================================================
+============================================================
+TEST 10: Unknown keys
 ============================================================
 Defining key module/key-one on line ...
 "value"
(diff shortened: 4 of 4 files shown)


l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,23 @@
 %   is not processed in any way by \cs{keys_set:nn}.
 % \end{variable}
 %
+% \subsection{Expanding the values of keys}
+%
+% To allow the user to apply value expansion at point-of-use, key names can be
```
FrankMittelbach: already at the time the key is set (and not later when it is used), % or something

I find point-of-use difficult to understand as the "usage" seems to happen later not then

l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,23 @@
...
+% followed by an expansion specifier. This is given by appending |:| and a
+% single letter specifier to the key name. The latter are the normal argument
+% specifiers for \pkg{expl3}, thus may be one of \texttt{n} (redundant but
+% supported), \texttt{o}, \texttt{V}, \texttt{v} or \texttt{e}. Thus for example
+% \begin{verbatim}
+%   \tl_set:Nn \l_mymodule_value_tl
```
FrankMittelbach: where is the "n" arg ?

l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,23 @@
...
+%   \keys_set:nn { mymodule } { key = value }
+%   \keys_set:nn { mymodule } { key : o = \l_mymodule_value_tl }
+%   \keys_set:nn { mymodule } { key : V = \l_mymodule_value_tl }
+%   \keys_set:nn { mymodule } { key : v = { l_mymodule_value_tl } }
+%   \keys_set:nn { mymodule } { key : e = \exp_not:V \l_mymodule_value_tl }
```
FrankMittelbach: I would suggest to use different (more meaningful) examples, e.g., some where V is actually doing something or e containing \dim_eval:n or the like

l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,23 @@
...
+% \end{verbatim}
+% all pass \texttt{value} to the key handler.
```
FrankMittelbach: that would then need some adjustment of course

l3kernel/l3keys.dtx
```diff
@@ -3726,6 +3775,12 @@
     The~module~'#2'~does~not~have~a~key~called~'#1'.\\
     Check~that~you~have~spelled~the~key~name~correctly.
   }
+\msg_new:nnnn { keys } { unknown-expansion }
+  { The~value~expansion~'#1'~is~unknown. }
+  {
+    Key~values~can~only~be~expanded~using~one~of~the~pre-defined~methods:~
+    n,~o,~V,~v~or~e.
```
FrankMittelbach: c might be useful here too? (given that we support v)

l3kernel/l3keys.dtx
josephwright: I'm trying to contrast with a key where the expansion is 'built-in', so e.g. `\tl_set_e:N` or similar - suggestions on wording welcome.

l3kernel/l3keys.dtx
josephwright: Key values are treated as `n`-type as standard - multiple tokens with no expansion. So for example with `key .tl_set:N = \l_tmpa_tl`, setting `key = value` does the same as `\tl_set:Nn \l_tmpa_tl { value }`.

l3kernel/l3keys.dtx
josephwright: OK, these were just my test set as they show the outcomes working :)

l3kernel/l3keys.dtx
josephwright: Doesn't really make sense - we are expanding `n`-type tokens - as I've said, we are in the end storing a token list in the first instance, so the expansion behaviours available do not include `c`. Also, really would be odd to do `key:c = value` rather than just `key = \value`.

l3kernel/l3keys.dtx
FrankMittelbach: not my point I think. You need to say `\tl_set:Nn \l_mymodule_tl {value}` in your example (as you do in your test file)

l3kernel/l3keys.dtx
FrankMittelbach: but less useful to understand for a user why one or the other is useful

l3kernel/l3keys.dtx
FrankMittelbach: that depends on what "\value" is, it may contain chars that you can't just write behind a backslash but you need c expansio nto construct it. For example a use who want to have \@secntformat as a value. Not insisting but it is similar to supporting v in addition to V (same reason)

l3kernel/l3keys.dtx
FrankMittelbach: normally key values aren't expanded at time of setting and all this expansion stuff is there precisely when no expansion happens. I thought I gave a suggestion. Not good?

l3kernel/l3keys.dtx
josephwright: Ah, right

l3kernel/l3keys.dtx
josephwright: I see your point re. `v`, but my take is to omit that :)

l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,36 @@
...
+% followed by an expansion specifier. This is given by appending |:| and a
+% single letter specifier to the key name. The latter are the normal argument
+% specifiers for \pkg{expl3}, thus may be one of \texttt{n} (redundant but
+% supported), \texttt{o}, \texttt{V} or \texttt{e}. Thus for example
+% \begin{verbatim}
+%   \keys_define:n { mymodule } { key .code:n = \tl_show:n { "#1" } }
+%   \tl_set:Nn \l_mymodule_value_tl { value }
+%   \dime_set:N \l__mymodule_value_dim { 123pt }
+%   \keys_set:nn { mymodule } { key = value }
+%   \keys_set:nn { mymodule } { key : o = \l_mymodule_value_tl }
```
davidcarlisle: I think here it would be better to show no space around the `:` spaces work here but just because of L3 syntax, you could equally spell the key name as `k e y` with spaces.  Unlike defining the keys, setting them is often, perhaps usually, lifted up to a document level command and there you need `key:e` with no space (unless we want to take the cost of removing spaces by hand.)

```
\documentclass{article}

\begin{document}

\ExplSyntaxOn
\keys_define:nn {me} {
  k .tl_set:N  = \l_me_foo_tl
}

\def\myset{\keys_set:nn {me}}
\def\myshow{\show\l_me_foo_tl}
\ExplSyntaxOff

\def\qq{88}

%\myset{k :e=\qq} %expansion ' e' unknown
%\myset{k :e=\qq} %key 'me/k' unknown
%\myset{k : e=\qq} %expansion ' e' unknown
\myset{k:e=\qq} % works
\myshow

\end{document}
```

josephwright: I could do that, or I could zap the spaces ... thoughts?

davidcarlisle: I think the analogy with csname suffixes is stronger if you use `:e` with no space so I'd be happy to document that form, but in all the other kv parsing I suppose an extra pass to lose space here is another possibility if people want to write `key : e `  (or could you lose the spaces in the same pass, so ignoring spaces was "free"

josephwright: I've changed the docs; we do have 'trim spaces only one side' functions, so if we wish to revisit later, the cost of space trimming isn't a disaster.

FrankMittelbach: I prefer not show spaces between around the  : . In my opinion the syntax should be key:char regardless of the fact that within ExplSyntaxOn you can in fact always write `m yk ey : e` or worse to mean mykey:e

l3kernel/l3keys.dtx
```diff
@@ -3104,7 +3141,8 @@
 %     \@@_find_key_module_auxi:Nw   ,
 %     \@@_find_key_module_auxii:Nw  ,
 %     \@@_find_key_module_auxiii:Nn ,
-%     \@@_find_key_module_auxiv:Nw
+%     \@@_find_key_module_auxiv:NNw ,
+%     \@@_find_key_module_auxiv:NNw
```
muzimuzhi:
```suggestion
%     \@@_find_key_module_auxiv:Nw ,
%     \@@_find_key_module_auxv:Nw
```

l3kernel/l3keys.dtx
```diff
@@ -3726,6 +3788,12 @@
     The~module~'#2'~does~not~have~a~key~called~'#1'.\\
     Check~that~you~have~spelled~the~key~name~correctly.
   }
+\msg_new:nnnn { keys } { unknown-expansion }
+  { The~value~expansion~'#1'~is~unknown. }
+  {
+    Key~values~can~only~be~expanded~using~one~of~the~pre-defined~methods:~
+    n,~o,~V~or~e.
```
muzimuzhi: Lowercase-`v` is missing.

l3kernel/l3keys.dtx
```diff
@@ -3263,7 +3318,14 @@
               { \@@_default_inherit: }
           }
       }
-      { \tl_set:Nn \l_keys_value_tl {#1} }
+      {
+        \cs_if_exist_use:cF { tl_set:N \l_@@_exp_str }
+          {
+            \msg_error:nnV { keys } { unknown-expansion } \l_@@_exp_str
+            \use_none:nn
+          }
```
muzimuzhi: `f` or `x` also passes the check `\cs_if_exist_use:cF { tl_set:N \l_@@_exp_str }`, making it more loose than the `unknown-expansion` message reads.

FrankMittelbach (approved):
Still think that c an v are useful, but that can wait until you get user requests :-)

l3kernel/l3keys.dtx
```diff
@@ -3174,11 +3219,21 @@
...
+  }
+\use:e
+  {
+    \cs_new_protected:Npn \exp_not:N \@@_find_key_module_auxv:Nw
+      #1 #2 \token_to_str:N : #3 \token_to_str:N : #4 \s_@@_mark
+  }
+  {
+    \cs_set_nopar:Npn #1 {#2}
+    \cs_set_nopar:Npn \l_@@_exp_str {#3}
   }
```
muzimuzhi, line 3237: `\token_to_str:N :` can be replaced with `\c_colon_str`.

Or, these two function definitions can be wrapped in `\@@_tmp:w`, then actually defined by `\exp_args:No \@@_tmp:w { \c_colon_str }`.

muzimuzhi (changes requested):
Typos and a bit more.

l3kernel/l3keys.dtx
josephwright: By-design - see discussion with @FrankMittelbach about the logic of what seems reasonable to pas

l3kernel/l3keys.dtx
josephwright: Yes, again by-design - these work but I don't want to suggest them to people as almost always `e`-type would be better

l3kernel/l3keys.dtx
josephwright: Yeah, but I don't know this is any clearer

l3kernel/l3keys.dtx
davidcarlisle: @josephwright  I'm not sure why you say `c` wouldn't make sense? It would make sense in the same way that c generally makes sense wouldn't it? supporting all the L3 letters seems simpler than subsetting here.

josephwright: Key values are strings, so are  (more-or-less) `n`-type - you can therefore `o`-, `V`-, `v`-, `e`-, `f`- or `x`-type expand them, but not `c`-type. `v`-type feels rather odd to do, and we don't really want to encourage new `f`- or `x`-type usage, hence I've omitted those from the message by design.

muzimuzhi: > Key values are strings, [...]

Not l3str string, but `n`-type parameter.

josephwright: Yeah, I do mean here the general programming idea of a string - so for us a tl.

Skillmon: First: Key values are *not* strings, they are token lists in their most abstract form! (as already established)

Second: Subsetting is more work than simply allowing all variants (make sure it's a single token, stringify and do `\cs_if_exist_use:cTF { exp_args:N #1 }`.

Third: There might be valid keys that only take a single token in user land, which might be enforced by additional checks in the user code. For instance a key allowing to define the control sequence in which to store the results of some operation. Here `:c` would be helpful.

l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,36 @@
 %   is not processed in any way by \cs{keys_set:nn}.
 % \end{variable}
 %
+% \subsection{Expanding the values of keys}
+%
+% To allow the user to apply expansion of values when the key is set, rather
+% than when they are used, key names can be
```
Skillmon: I still find the distinction with "use" strange. What does "use" mean? The *key* isn't used when the (internal) variable, which was set to the value given, is used, the *variable* is used then. So what is key usage?

l3kernel/l3keys.dtx
josephwright: > First: Key values are _not_ strings, they are token lists in their most abstract form! (as already established)

Yeah, sorry - like I say, I meant in general programming terms - for us, yes, the _key_ values are token lists even if they contain something used in another way (control sequence, dim expr, ...).

> Second: Subsetting is more work than simply allowing all variants (make sure it's a single token, stringify and do `\cs_if_exist_use:cTF { exp_args:N #1 }`.

The subset is in the message (user guidance) not in the code, which does indeed just do a `\cs_if_exist:c`, but for `{ tl_set:N #1 }` (hence no `c`-type: `v`, `f` and `x` are a case of what goes in the docs).

> Third: There might be valid keys that only take a single token in user land, which might be enforced by additional checks in the user code. For instance a key allowing to define the control sequence in which to store the results of some operation. Here `:c` would be helpful.

I see that, but the _key_ is still a token list so isn't `N` type!

davidcarlisle: I don't think the distinction you are making really holds here. `c` expansion would have a clear unambiguous and useful behaviour here. `key:c=wibbl@`  would be `key=\wibbl@` without needing to surround the outer call in `\makeatletter` the documentation would be simpler if we just said all exp_arg letters worked.

FrankMittelbach: precisely. that is the use case for c (or for v) where you want to have a csname as the value but you need to get it constructed because it contains unusual catcodes. And yes one could put `\makeatletter`or `\ExplSyntaxOn`on the outside but that would then apply to all key/values  not to just a single value of a key value list.

l3kernel/l3keys.dtx
josephwright: Suggestion on how to phrase this?

car222222: Needs more context to work out the correct phrasing here.

FrankMittelbach: To allow the user to apply expansion of values when the key is set, rather
 than when the value is later  used, ...

I agree with @Skillmon that it is not the key that is used. It is the value that is used at a later stage.

josephwright:
@Skillmon , @car222222  How does this look now?

l3kernel/l3keys.dtx
Skillmon: And even if `\makeatletter` or `\ExplSyntaxOn` were acceptable, it might very well be that the cs name is constructed programmatically.

l3kernel/l3keys.dtx
Skillmon: I'd just drop the "later used" part. For instance "To allow expansion control of values inside key=value lists" (or something along the lines, the first part of @FrankMittelbach's suggestion would be fine).

l3kernel/l3keys.dtx
```diff
@@ -774,6 +774,36 @@
...
+% To allow the user to apply expansion of values when the key is set, rather
+% than when the value is later used, key names can be
+% followed by an expansion specifier. This is given by appending |:| and a
+% single letter specifier to the key name. The latter are the normal argument
+% specifiers for \pkg{expl3}, thus may be one of \texttt{n} (redundant but
+% supported), \texttt{o}, \texttt{V} or \texttt{e}. Thus for example
+% \begin{verbatim}
+%   \keys_define:n { mymodule } { key .code:n = \tl_show:n { "#1" } }
+%   \tl_set:Nn \l_mymodule_value_tl { value }
+%   \dime_set:N \l__mymodule_value_dim { 123pt }
```
car222222, line 788: Should this be \dim_set:N ? (currency free version:-).

l3kernel/l3keys.dtx
```diff
@@ -3174,11 +3219,21 @@
...
-\cs_new_protected:Npn \@@_find_key_module_auxiv:Nw
+\cs_new_protected:Npe \@@_find_key_module_auxiv:Nw
     #1 #2 \s_@@_nil #3 \s_@@_mark
     \s_@@_nil \@@_find_key_module_auxiv:Nw #4
   {
-    \cs_set_nopar:Npn #4 { #2 }
+    \exp_not:N \@@_find_key_module_auxv:Nw #4
+    #2 \token_to_str:N : n \token_to_str:N : \s_@@_mark
+  }
+\use:e
```
car222222, line 3229: Maybe needs some indentation?

l3kernel/l3keys.dtx
```diff
@@ -3243,6 +3298,12 @@
 %
 % \begin{macro}{\@@_value_or_default:n}
 % \begin{macro}{\@@_default_inherit:}
+% \begin{macro}
+%   {
+%     \@@_value_set:Nn, \@@_value_set:No, \@@_value_set:NV,
+%     \@@_value_set:Nv, \@@_value_set:Ne,
+%   }
+% \begin{macro}{\@@_value_set:NN, \@@_value_set:Nc}
 %   If a value is given, return it as |#1|, otherwise send a default if
```
car222222, line 3307: "send" ?? 
Maybe  ==>  set
Or     ==>  return

l3kernel/l3keys.dtx
```diff
@@ -3263,7 +3324,14 @@
...
           }
       }
-      { \tl_set:Nn \l_keys_value_tl {#1} }
+      {
+        \cs_if_exist_use:cF { @@_value_set:N \l_@@_exp_str }
+          {
+            \msg_error:nnV { keys } { unknown-expansion } \l_@@_exp_str
+            \use_none:nn
+          }
+          \l_keys_value_tl {#1}
```
car222222, line 3333: Indentation OK?

car222222 (commented):
A few suggestions and queries.

l3kernel/l3keys.dtx
josephwright: It's basically `\tl_set:Nn`, hence the naming

l3kernel/l3keys.dtx
josephwright: Adjusted

l3kernel/l3keys.dtx
josephwright: Fixed :)

l3kernel/l3keys.dtx
FrankMittelbach: Chris talks about line 3307

josephwright: Yeah, but this is still an indirection for `\tl_set:Nn`, and it's still about what the key value is set to.

l3kernel/l3keys.dtx
muzimuzhi: Aha there are two minor typos in the name `\dime_set:N`: `dime` should be `dim` and `:N` should be `:Nn`.

https://github.com/latex3/latex3/commit/b9fc76a353d9550bf90df6d5bd693e7a0d4c2689 fixed the second one, while the first one survives until @dbitouze opened PR #1787.

PR #1770, opened 2025-07-03
Support LuaMetaTeX's def. of \meaning

josephwright:
This is done using an internal function as the Lua emulation doesn't 100% match TeX82's `\meaning`: we need to use `\meaning` for various things, so this is important.

### DIFF
--- a/l3kernel/l3basics.dtx
+++ b/l3kernel/l3basics.dtx
@@ -3370,10 +3370,15 @@
 %
 % \subsection{Decomposing a macro definition}
 %
+%    \begin{macrocode}
+%<@@=cs>
+%    \end{macrocode}
+%
 % \begin{macro}{\cs_prefix_spec:N}
 % \begin{macro}{\cs_parameter_spec:N}
 % \begin{macro}{\cs_replacement_spec:N}
 % \begin{macro}{\@@_prefix_arg_replacement:wN}
+% \begin{macro}{\@@_meaning:N}
 %   We sometimes want to test if a control sequence can be expanded to
 %   reveal a hidden value. However, we cannot just expand the macro
 %   blindly as it may have arguments and none might be
@@ -3394,7 +3399,7 @@
     \token_if_macro:NTF #1
       {
         \exp_after:wN \@@_prefix_arg_replacement:wN
-          \token_to_meaning:N #1 \s_@@_stop \use_i:nnn
+          \@@_meaning:N #1 \s_@@_stop \use_i:nnn
       }
       { \scan_stop: }
   }
@@ -3403,7 +3408,7 @@
     \token_if_macro:NTF #1
       {
         \exp_after:wN \@@_prefix_arg_replacement:wN
-          \token_to_meaning:N #1 \s_@@_stop \use_ii:nnn
+          \@@_meaning:N #1 \s_@@_stop \use_ii:nnn
       }
       { \scan_stop: }
   }
@@ -3412,11 +3417,53 @@
     \token_if_macro:NTF #1
       {
         \exp_after:wN \@@_prefix_arg_replacement:wN
-          \token_to_meaning:N #1 \s_@@_stop \use_iii:nnn
+          \@@_meaning:N #1 \s_@@_stop \use_iii:nnn
       }
       { \scan_stop: }
   }
 %    \end{macrocode}
+%   LuaMeta\TeX{}'s version of \tn{meaning} doesn't give the same output
+%   as the classical one. So we abstract out the idea at a low level, as
+%   we need the details to be correct here.
+%    \begin{macrocode}
+\cs_if_exist:NF \@@_meaning:N
+  { \cs_new_eq:NN \@@_meaning:N \cs_meaning:N }
+%</code>
+%    \end{macrocode}
+%   This is a modified version of code suggested by Max Chernoff
+%   (\url{https://chat.stackexchange.com/transcript/message/67947633#67947633}).
+%   We aim to provide the classical \TeX{}82 appearance but covering \eTeX{}:
+%   due to changes in the engine, the only useful prefix here is |\protected|.
+%   This does not cover tokens other than control sequences, so we only use
+%   it for the internal needed here.
+%    \begin{macrocode}
+%<*lua>
+if status.luatex_engine == "luametatex" then
+  luacmd('@@_meaning:N', function()
+    local csname = scan_csname(true)
+    local parameters = get_macro(csname, false, true)
+    local body = get_macro(csname)
+    if parameters == body then
+      -- If parameters and body are the same, we assume that there
+      -- are no parameters. (Probably a LuaMetaTeX bug...)
+      parameters = ""
+    end
+    local cstoken = token_create(csname)
+    local prefixes = {}
+    if token["getprotected"](cstoken) then
+      prefixes[#prefixes + 1] = "\\protected"
+    end
+    local prefix = table_concat(prefixes, " ")
+    if prefix ~= "" then
+      prefix = prefix .. " "
+    end
+    sprint(-2, string.formatters["%smacro:%s->%s"](prefix, parameters, body))
+  end, 'global')
+end
+%</lua>
+%<*code>
+%    \end{macrocode}
+% \end{macro}
 % \end{macro}
 % \end{macro}
 % \end{macro}
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -563,11 +563,11 @@ ltxutils.filesize = filesize
 %
 % \begin{macro}[int]{luacmd}
 % An internal function for defining control sequences form Lua which behave
-% like primitives. This acts as a wrapper around |token.set_lua| which accepts
+% like primitives. This acts as a wrapper around |token.setlua| which accepts
 % a function instead of an index into the functions table.
 %    \begin{macrocode}
 local luacmd do
-  local set_lua = token.set_lua
+  local set_lua = token.setlua or token.set_lua
   local undefined_cs = command_id'undefined_cs'
 
   if not context and not luatexbase then require'ltluatex' end
@@ -591,7 +591,7 @@ local luacmd do
     function luacmd(name, func, ...)
       local tok = token_create(name)
       if tok.command == undefined_cs then
-        token.set_lua(name, register(func), ...)
+        set_lua(name, register(func), ...)
       else
         functions[tok.index or tok.mode] = func
       end
--- a/l3kernel/l3luatex.dtx
+++ b/l3kernel/l3luatex.dtx
@@ -336,6 +336,7 @@ local string   = string
 local tex      = tex
 local texio    = texio
 local tonumber = tonumber
+local token    = token
 %    \end{macrocode}
 %
 %   Local copies of standard functions.
@@ -358,11 +359,13 @@ local package_loaded    = package.loaded
 local package_searchers = package.searchers
 local table_concat      = table.concat
 
+local scan_csname  = token.scancsname
 local scan_int     = token.scan_int or token.scan_integer
 local scan_string  = token.scan_string
 local scan_keyword = token.scan_keyword
 local put_next     = token.put_next
 local token_create = token.create
+local get_macro    = token.getmacro
 local token_new    = token.new
 local set_macro    = token.set_macro
 %    \end{macrocode}
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -7,6 +7,9 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+- Track LuaMetaTeX changes in `\meaning` output
+
 ## [2025-06-30]
 
 ### Added
--- a/l3kernel/l3.ins
+++ b/l3kernel/l3.ins
@@ -177,6 +177,7 @@ and all files in that bundle must be distributed together.
 \generate{\file{expl3.lua}{
   \from{l3luatex.dtx}{lua}
   \from{l3names.dtx}{lua}
+  \from{l3basics.dtx}{lua}
   \from{l3sys.dtx}{lua}
   \from{l3token.dtx}{lua}
   \from{l3intarray.dtx}{lua}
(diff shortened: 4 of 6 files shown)


josephwright:
The ConTeXt test is currently disabled - I'll switch back on once `l3build` updates in TL.

l3kernel/l3luatex.dtx
```diff
@@ -358,11 +359,13 @@ local package_loaded    = package.loaded
 local package_searchers = package.searchers
 local table_concat      = table.concat
 
+local scan_csname  = token.scancsname
 local scan_int     = token.scan_int or token.scan_integer
 local scan_string  = token.scan_string
 local scan_keyword = token.scan_keyword
 local put_next     = token.put_next
 local token_create = token.create
+local get_macro    = token.getmacro
```
gucci-on-fleek: LuaTeX only provides `token.scan_csname` and `token.get_macro` (with the underscores), while the LuaMetaTeX engine only provides `token.scancsname` and `token.getmacro` (without the underscores), but both ConTeXt and LuaMetaLaTeX provide compatibility aliases for the underscored versions (`token.scan_csname` and `token.get_macro`). 

So I'd recommend either just using the underscored versions, or putting these new `local`s inside the `if status.luatex_engine == "luametatex" then` block, because these current definitions of `scan_csname` and `get_macro` won't work with LuaTeX.

l3kernel/l3basics.dtx
```diff
@@ -3412,11 +3417,53 @@
...
+    local parameters = get_macro(csname, false, true)
+    local body = get_macro(csname)
+    if parameters == body then
+      -- If parameters and body are the same, we assume that there
+      -- are no parameters. (Probably a LuaMetaTeX bug...)
+      parameters = ""
+    end
+    local cstoken = token_create(csname)
+    local prefixes = {}
+    if token["getprotected"](cstoken) then
```
gucci-on-fleek:
```suggestion
    if token.getprotected(cstoken) then
```

is better here since `a.["b"] == a.b` in Lua. I was only using the `["…"]` syntax in my earlier code since the index variable was generated by a loop.

l3kernel/l3basics.dtx
```diff
@@ -3412,11 +3417,53 @@
...
+      prefixes[#prefixes + 1] = "\\protected"
+    end
+    local prefix = table_concat(prefixes, " ")
+    if prefix ~= "" then
+      prefix = prefix .. " "
+    end
+    sprint(-2, string.formatters["%smacro:%s->%s"](prefix, parameters, body))
```
gucci-on-fleek: The only point to using `string.formatters` is that you can separate the format string definition from its use; if you're not going to store it in a separate variable, it's better to do

```suggestion
    sprint(-2, string.format("%smacro:%s->%s", prefix, parameters, body))
```

or

```suggestion
    sprint(-2, ("%smacro:%s->%s"):format(prefix, parameters, body))
```

l3kernel/l3basics.dtx
```diff
@@ -3412,11 +3417,53 @@
...
+    end
+    local cstoken = token_create(csname)
+    local prefixes = {}
+    if token["getprotected"](cstoken) then
+      prefixes[#prefixes + 1] = "\\protected"
+    end
+    local prefix = table_concat(prefixes, " ")
+    if prefix ~= "" then
+      prefix = prefix .. " "
+    end
```
gucci-on-fleek: Since there can only be 1 possible prefix, you can skip the table:

```suggestion
    local protected = ""
    if token["getprotected"](cstoken) then
      protected = "\\protected "
    end
```

l3kernel/l3basics.dtx
```diff
@@ -3412,11 +3417,53 @@
...
+%   This is a modified version of code suggested by Max Chernoff
+%   (\url{https://chat.stackexchange.com/transcript/message/67947633#67947633}).
+%   We aim to provide the classical \TeX{}82 appearance but covering \eTeX{}:
+%   due to changes in the engine, the only useful prefix here is |\protected|.
+%   This does not cover tokens other than control sequences, so we only use
+%   it for the internal needed here.
+%    \begin{macrocode}
+%<*lua>
+if status.luatex_engine == "luametatex" then
+  luacmd('@@_meaning:N', function()
```
gucci-on-fleek: I'd recommend using either all `'single quoted strings'` or all `"double quoted strings"`, but right now you have a mix of both.

gucci-on-fleek (commented):
Functionality all looks good to me (although I haven't tested this exact version). Just some minor stylistic comments; feel free to ignore :).

l3kernel/l3luatex.dtx
josephwright: I'll got for `name = LuaMetaTeX-name or LuaTeX-name`

l3kernel/l3basics.dtx
josephwright: I'm in two minds about whether to cover all prefixes or only the classical one - as I can't 'sort' `\long` anyway, it's not clear what's best.

l3kernel/l3basics.dtx
josephwright: I find the entire 'fill in the blanks' approach weird, so I just stuck to what you wrote :) I'll use your suggestion.

l3kernel/l3basics.dtx
josephwright: Again, first I think we need to decide if we need all the prefixes or not.

l3kernel/l3basics.dtx
josephwright: I have a feeling we are a bit inconsistent here - I'll tidy up the PR, and after it's merged look to pick one and stick to it generally!

josephwright:
@gucci-on-fleek OK, I think for the present we go with just `\protected`, but with a note that this may need to be revisited later. I'll adjust the code.

l3kernel/l3basics.dtx
```diff
@@ -3412,11 +3417,50 @@
...
+%   at present. (This may need to be revisited once the engine is more stable.)
+%   This does not cover tokens other than control sequences, so we only use
+%   it for the internal needed here.
+%    \begin{macrocode}
+%<*lua>
+if status.luatex_engine == "luametatex" then
+  luacmd("@@_meaning:N", function()
+    local csname = scan_csname(true)
+    local parameters = get_macro(csname, false, true)
+    local body = get_macro(csname)
```
zauguin: This breaks for active characters.

josephwright:
> I would suggest to reimplement \cs_prefix_spec:N and friends directly in Lua for LuaMetaTeX, otherwise we use fragile constructions to build up a string only to split it up into the parts we started with immediately afterwards.

OK, I can sort that - how do want us to grab the argument?

l3kernel/l3basics.dtx
gucci-on-fleek: Right, we should be able to use `token.scannext()` (aka `token.get_next()`) instead though.

l3kernel/l3luatex.dtx
```diff
@@ -358,13 +359,16 @@ local package_loaded    = package.loaded
...
+local scan_csname   = token.scancsname or token.scan_csname
+local scan_int      = token.scan_int or token.scan_integer
+local scan_string   = token.scan_string
+local scan_keyword  = token.scan_keyword
+local put_next      = token.put_next
+local token_create  = token.create
+local get_macro     = token.getmacro or token.get_macro
+local get_protected = token.getprotected or token.get_protected
+local token_new     = token.new
+local set_macro     = token.set_macro
```
gucci-on-fleek: Might as well add the other non-underscored names while you're editing this part of the file.

```suggestion
local scan_string   = token.scanstring or token.scan_string
local scan_keyword  = token.scankeyword or token.scan_keyword
local put_next      = token.putnext or token.put_next
local token_create  = token.create
local get_macro     = token.getmacro or token.get_macro
local get_protected = token.getprotected or token.get_protected
local token_new     = token.new
local set_macro     = token.setmacro or token.set_macro
```

gucci-on-fleek (approved):
LGTM as an implementation of `\__cs_meaning:N`; but @zauguin's suggestion of instead implementing `\cs_prefix_spec:N` also seems like a reasonable idea to me.

gucci-on-fleek:
@josephwright 

> > I would suggest to reimplement \cs_prefix_spec:N and friends directly in Lua for LuaMetaTeX, otherwise we use fragile constructions to build up a string only to split it up into the parts we started with immediately afterwards.
> 
> OK, I can sort that - how do want us to grab the argument?

`token.scannext()` will grab the next token literally. Don't use `token.scantoken()`, that will expand the next token before grabbing it.

josephwright:
@zauguin  Thanks - I'm happy to rework the code but likely you'll be faster!

gucci-on-fleek:
@zauguin 

> > > I would suggest to reimplement \cs_prefix_spec:N and friends directly in Lua for LuaMetaTeX, otherwise we use fragile constructions to build up a string only to split it up into the parts we started with immediately afterwards.
> > 
> > 
> > OK, I can sort that - how do want us to grab the argument?
> 
> Basically token.scannext, then there sadly don't seem to be nice direct accessors for the parameters so you probably have to differentiate between active charaters and regular commands to finally call token.getmacro.

You can use the U+FFFF trick. Compiling this standalone test document

```tex
\startluacode
    local index = context.functions.register(function()
        local argument = token.scannext()
        local csname = argument.csname
        if argument.active then
            csname = "\string\u{FFFF}" .. argument.csname
        end

        local parameters = token.getmacro(csname, false, true)
        local body = token.getmacro(csname)
        if parameters == body then
             parameters = ""
        end

        local protected = ""
        if argument.protected then
            protected = "\\protected "
        end
        tex.sprint(-2, ("%smacro:%s->%s"):format(protected, parameters, body))
    end)
    token.setlua("__cs_meaning:N", index)
\stopluacode

\usemodule[expl3-generic]

\ExplSyntaxOn
    \cs_gset_eq:NN \cs_meaning:N \__cs_meaning:N

    \cs_new_nopar:Npn \example_cs: { }

    \char_set_catcode_active:N \@
    \cs_set:Npn \@ { Macro }
    \cs_set:Npn @ { Active }

    \cs_generate_variant:Nn \tl_analysis_show:n { e }
    \tl_analysis_show:e { \cs_meaning:N \example_cs: }
    \tl_analysis_show:e { \cs_meaning:N \@ }
    \tl_analysis_show:e { \cs_meaning:N @ }
\ExplSyntaxOff
\normalend
```

gives the following results:

```text
just a message  > The token list contains the tokens:
>  m (the character U+006D m)
>  a (the character U+0061 a)
>  c (the character U+0063 c)
>  r (the character U+0072 r)
>  o (the character U+006F o)
>  : (the character U+003A 'colon')
>  - (the character U+002D 'hyphen minus')
>  > (the character U+003E 'more than sign')
just a message  > The token list contains the tokens:
>  m (the character U+006D m)
>  a (the character U+0061 a)
>  c (the character U+0063 c)
>  r (the character U+0072 r)
>  o (the character U+006F o)
>  : (the character U+003A 'colon')
>  - (the character U+002D 'hyphen minus')
>  > (the character U+003E 'more than sign')
>  M (the character U+004D M)
>  a (the character U+0061 a)
>  c (the character U+0063 c)
>  r (the character U+0072 r)
>  o (the character U+006F o)
just a message  > The token list contains the tokens:
>  m (the character U+006D m)
>  a (the character U+0061 a)
>  c (the character U+0063 c)
>  r (the character U+0072 r)
>  o (the character U+006F o)
>  : (the character U+003A 'colon')
>  - (the character U+002D 'hyphen minus')
>  > (the character U+003E 'more than sign')
>  A (the character U+0041 A)
>  c (the character U+0063 c)
>  t (the character U+0074 t)
>  i (the character U+0069 i)
>  v (the character U+0076 v)
>  e (the character U+0065 e)
```

josephwright:
But first I have an `siunitx` bug to fix ...

l3kernel/l3basics.dtx
```diff
@@ -3382,45 +3389,121 @@
...
+      }
+  }
+%</code>
+%    \end{macrocode}
+%   This is a modified version of code suggested by Max Chernoff
+%   (\url{https://chat.stackexchange.com/transcript/message/67947633#67947633}).
+%   We aim to provide the classical \TeX{}82 appearance but covering \eTeX{}:
+%   due to changes in the engine, the only useful prefix here is |\protected|
+%   at present. (This may need to be revisited once the engine is more stable.)
+%   This does not cover tokens other than control sequences, so we only use
```
gucci-on-fleek, line 3471:
```suggestion
%   This does not cover tokens other than control sequences and active characters, so we only use
```

l3kernel/l3basics.dtx
```diff
@@ -3382,45 +3389,121 @@
...
+if status.luatex_engine == 'luametatex' then
+  local function scan_full_csname()
+    local t = get_next()
+    local csname = get_csname(t)
+    return t.active and active_prefix .. csname or csname, t
+  end
+  luacmd('@@_macro_prefix_spec:N', function()
+    local token = get_next()
+    if get_protected(token) then
+      sprint(-2, "\\protected ")
```
gucci-on-fleek, line 3484: The other strings in this PR use `'single quotes'`, so maybe this one should too?

```suggestion
      sprint(-2, '\\protected ')
```

gucci-on-fleek, line 3484: Maybe `-2` should be moved into a constant like `thetoks_catcodes`?

gucci-on-fleek (approved):
I haven't tested this, but the functionality all looks good to me. I've added a few stylistic comments, but feel free to ignore those.

l3kernel/l3basics.dtx
zauguin: Generally I tend to use single quotes for keyword like/internal strings (like `'global'`) and double quotes for user strings (like error messages). The `\protected` here is somewhat in between since it's a keyword in TeX, but from the Lua perspective mostly arbitrary user output. So I would tend towards keeping the double quotes, but also wouldn't mind changing it if you think that single quotes are a better fit.

josephwright:
I think we are good to go here - we can tidy up quotes, etc., later - I'll merge.

l3kernel/l3basics.dtx
gucci-on-fleek: Ah, that makes sense then (and seems reasonable to me too).

PR #1774, opened 2025-07-11
Skip math mode in text word/grapheme mapping

josephwright:


l3kernel/l3text-map.dtx
```diff
@@ -127,6 +133,69 @@
         \@@_map_output:nn {#1} {#3}
         \text_map_break:
       }
+    \exp_args:NV \@@_map_math_search:nnnnN
+      \l_text_math_delims_tl {#1} {#2} {#3} #4
+  }
+%    \end{macrocode}
+%   We need to exclude math mode here as quite apart from the conceptual
+%   questions, it could contain implicit tokens. Those would case all sorts
```
FrankMittelbach: would case?

FrankMittelbach (commented):
can't pretend to understand the code (just from the diff) so only one comment/question

PR #1776, opened 2025-07-17
Wordbreak actives as csnames

josephwright:


josephwright:
Addresses https://github.com/latex3/latex2e/issues/1824

l3kernel/testfiles/m3text006.lvt
```diff
@@ -195,4 +195,11 @@
     \test:n { Oops~not~close~token~in~$y~=~mx~+~c }
   }
 
+\cs_gset_protected:Npn \cs_tmp:w #1 { XXX #1 }
+\char_set_active_eq:NN * \cs_tmp:w
+\TEST { Active~chars }
+  {
+    \test:n { Some~text~*A~bbb }
+  }
```
muzimuzhi: Lacking a `\char_set_catcode_active:N *` which indeed sets `*` to active. Thus the current version of this test works even without the fix.

```suggestion
\BEGINTEST { Active~chars }
  \cs_gset:Npn \cs_tmp:w #1 { XXX #1 }
  \char_set_active_eq:NN * \cs_tmp:w
  \char_set_catcode_active:N *

  \test:n { Some~text~*A~bbb }
\ENDTEST
```

muzimuzhi (changes requested):

josephwright:
As it stands, the PR will sort the immediate issue in `babel` but does leave open how to handle macros or actives that take arguments more generally - it's not obvious how to break 'words' in this case.

davidcarlisle (approved):

PR #1791, opened 2025-08-30
Use Lua parser for Unicode data files

zauguin:
Currently breaking Context compatibility and depends on a not yet released lua-uni-algos version.

Replaces parser for UnicodeData.txt, GraphemeBreakProperty.txt and WordBreakCompatibility.txt with Lua implementations.

The l3unicode.dtx changes are best reviewed with whitespace changes ignored since a lot of the code got moved into conditionals and therefore the indentation changed.


texmf/tex/luatex/lua-uni-algos/lua-uni-case.lua
```diff
@@ -0,0 +1,67 @@
+-- lua-uni-graphemes.lua
```
josephwright: Name seems to be wrong!

car222222: Why wrong? What does it contain? 

But maybe it got changed by now?

josephwright: @car222222 Yup

josephwright:
@zauguin Anything outstanding or am I OK to merge?

zauguin:
Nothing outstanding, can be merged.

zauguin:
We should probably alert TeX Live to add a dependency to lua-uni-algos once this is released.

josephwright:
> We should probably alert TeX Live to add a dependency to lua-uni-algos once this is released.

Cool, I'll let Karl know now

zauguin:
> It's not quite what I expected, but it will work for 99% of users. I suspect people using `expl3` in generic mode will either have LaTeX installed anyway or will be using the ConTeXt stand-alone system. The latter won't be affected by a TL dependency anyway, so it's not vital to adjust.

`latex-bin` already depended on lua-uni-algos indirectly anyway through luaotfload. So if it's only added there we don't need to bother with the dependency. But otherwise I agree, unlikely to have a lot of practical impact.

PR #1799, opened 2025-09-26
Fix a bug in `\GetDocumentCommandArgSpec`.

sasozivanovic:
Commands with argument specification `+m` were incorrectly reported to have argument specification `m`.  In fact, more generally, commands with an arbitrary number of `+m` arguments were reported to have that number of `m` arguments.

Analysis: all these commands use `\cmd_start_optimized`. `\cmd_get_arg_spec:NTF` (called by `\GetDocumentCommandArgSpec` and friends) explicitly checked for the situation, but then simply returned the appropriate number of `m`s without checking the `\long` status of the command.

Fix: I have added the code prefixing the `m` with a `+` whenever the command is `\long`.

Implementation detail 1: I have used `\str_case:` to check whether `\long` appears in the return value of `\cs_prefix_spec:` because `\str_if_in:` is not expandable.  The method is feasible because `\cs_prefix_spec:` has only three possible return values: `\long `, `\protected\long ` and empty.

Implementation detail 2: the return value of `\cs_prefix_spec:` contains a space at the end, but not between `\protected` and `\long`.  Thus `\string` in front of `\protected`.

### DIFF
--- a/l3packages/xparse/xparse.dtx
+++ b/l3packages/xparse/xparse.dtx
@@ -1429,6 +1429,8 @@
 %   command is optimized: in the latter case, we can reconstruct the
 %   spec.
 %    \begin{macrocode}
+\cs_generate_variant:Nn \cs_prefix_spec:N { c }
+\cs_generate_variant:Nn \str_case:nn { ee }
 \cs_gset_protected:Npn \@@_get_arg_spec:NTF #1#2#3
   {
     \__kernel_cmd_if_xparse:NTF #1
@@ -1445,7 +1447,15 @@
                           { \cs_to_str:N #1 \c_space_tl code }
                       } / 2
                   }
-                  { m }
+                  {
+                    \str_case:ee
+                      { \cs_prefix_spec:c { \cs_to_str:N #1 \c_space_tl code } }
+                      {
+                        { \long } { + }
+                        { \string\protected\long } { + }
+                      }
+                    m
+                  }
               }
               { \tl_item:Nn #1 { 2 } }
           }


l3packages/xparse/xparse.dtx
```diff
@@ -1445,7 +1447,15 @@
...
                   }
-                  { m }
+                  {
+                    \str_case:ee
+                      { \cs_prefix_spec:c { \cs_to_str:N #1 \c_space_tl code } }
+                      {
+                        { \long } { + }
+                        { \string\protected\long } { + }
+                      }
+                    m
```
muzimuzhi: Use `\token_to_string:N` instead of `\string` to be more expl3-ic?

Or, use builtins `\token_if_long_macro:NTF` and `\token_if_protected_long_macro:NTF`.

muzimuzhi (commented):
Can you add some tests to `xparse009.lvt`? Can be as simple as a new test case `\DeclareDocumentCommand \fooc { +m +m +m } { }`.

https://github.com/latex3/latex3/blob/4c017aa7117e2e5bc024f2e7b52bda3b513bf757/l3packages/xparse/testfiles/xparse009.lvt#L90-L98

To update the test result, you can run
```
$ cd /path/to/latex3
$ cd l3packages/xparse
$ l3build save xparse009
```

muzimuzhi:
> I'm wondering if we really want to change what is now historical code which is more-or-less 'frozen'.

It's mostly about the add-on utilities that do not exist in the LaTeX2e `ltcmd` module, but exist in `xparse` package.

josephwright:
@muzimuzhi Yes but they are here as we decided specifically against including them in the kernel - retained for stability only.

josephwright:
We've discussed the PR internally and will go ahead with a merge. At the same time, `xparse` is _all_ regarded as 'only use for historical code' status. As such, what we will do is look to re-add the functionality here to the kernel - fixing some issues while we are at it. We want to be clear to the user that this is quite a specialist function, so plan to go with a code-level interface (`\cmd_...` - naming to be finalised - and I may go with something expandable in parallel to a 'get' function). There will be a PR shortly in the LaTeX2e repo to implement that.

We would like to include the rationale for the function in the sources, not just on GitHub. I'll aim to do that for the 2e PR.

sasozivanovic:
@josephwright This is great news, thank you all very much!

I added the changelog entry. If there's anything else I can do, please let me know. (Should I draft the rationale for the inclusion in the sources?)

sasozivanovic:
@u-fischer None :smile: The retrieved argument specification is *interpreted* by `\CollectArguments` of package CollArgs, an auxiliary package to Memoize, whose job is to replicate the invocation of the command, so that Memoize can associate the result of the externalization with the code that produced it.

sasozivanovic:
@josephwright Should I write from a general perspective, or mention the usage case in Memoize as well?

l3packages/xparse/xparse.dtx
```diff
@@ -1445,7 +1445,17 @@
...
+                    \exp_args:Nc \token_if_protected_long_macro:NTF
+                      { \cs_to_str:N #1 \c_space_tl code }
+                      { + }
+                      {
+                        \exp_args:Nc \token_if_long_macro:NT
+                        { \cs_to_str:N #1 \c_space_tl code }
+                        { + }
+                      }
+                    m
+                  }
```
FrankMittelbach, line 1460: I think it would be great if you copy your rationale given in the PR about that code piece into the code itself (between macrocode). Helps others to understand it in the future.

josephwright:
As part of moving to the 2e repo I'm also going to review some of the supporting structures: I'll aim to do also align the ones in `xparse` so we have a consistent position - may therefore be a few days before I can release once merged. But I will aim to have something in the kernel for 2025-11-01.

l3packages/xparse/xparse.dtx
sasozivanovic: I hope that I understand correctly that this suggestion is unrelated to my discussion with @josephwright above (about the rationale for why the code-level interface counterpart of `\GetDocumentCommandArgSpec` is (will be) included in the kernel), and that you suggest to comment the implementation of `\cmd_get_arg_spec:NTF` (currently in `xparse.dtx`). I pushed a commit containing the comment, but I added the text to the macro description, i.e. outside the `macrocode`, because as far as I can see, the macro definitions in the LaTeX source are normally not interrupted by comments.  Sorry if I misunderstood you (or Joseph).

josephwright:
@sasozivanovic  You've used Markdown comments in the LaTeX :-)

l3packages/xparse/xparse.dtx
josephwright: Looks good - I will likely revise a bit later but that will be as part of more significant steps to integrate (back) into the kernel.

FrankMittelbach: I meant to list in some way what you wrote above in the PR as part of the code, eg

>Implementation detail 1: I have used \str_case: to check whether \long appears in the return value of \cs_prefix_spec: because \str_if_in: is not expandable. The method is feasible because \cs_prefix_spec: has only three possible return values: \long , \protected\long and empty.

>Implementation detail 2: the return value of \cs_prefix_spec: contains a space at the end, but not between \protected and \long. Thus \string in front of \protected.

sasozivanovic: I see. However, these details are now obsolete, as I followed @muzimuzhi's suggestion and replaced that mess with `\token_if_long_macro:NTF` and `\token_if_protected_long_macro:NTF`.

sasozivanovic:
Many thanks, @josephwright!

`\bool_lazy_or:nnT` ... good to know :smile:

PR #1806, opened 2025-10-14
Use/test \NoValue rather than "-NoValue-"

josephwright:
Assuming we go with this, we will want to update ltcmd - probably that means checking the expl3 version and requiring it to be at least 2025-10-14.

### DIFF
--- a/l3kernel/l3tl.dtx
+++ b/l3kernel/l3tl.dtx
@@ -334,10 +334,6 @@
 %   \cs{NoValue}. This function is intended to allow
 %   construction of flexible document interface structures in which missing
 %   optional arguments are detected.
-%   \begin{texnote}
-%     The test will work for both the current \cs{NoValue} marker token and
-%     and the previous marker (|-NoValue-|).
-%   \end{texnote}
 % \end{function}
 %
 % \begin{function}[added = 2025-10-14]{\NoValue}


l3kernel/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+- Revise definition of 'no value' to allow use of `\NoValue` as a marker at the
+  design level
+
+### Deprecated
+- `\c_novalue_tl` in favor of `\NoValue`
```
FrankMittelbach, line 15: ... without the need to "o" expand  ?

l3kernel/l3tl.dtx
```diff
@@ -325,20 +325,19 @@
...
-%   contain the same list of tokens, both in respect of character codes and
-%   category codes. This means that
-%   \cs{exp_args:No} \cs{tl_if_novalue:nTF} \texttt\{ \cs{c_novalue_tl} \texttt\} is 
-%   logically \texttt{true} but \cs{tl_if_novalue:nTF} \texttt\{ \cs{c_novalue_tl} \texttt\}
-%   is logically \texttt{false}.
-%   This function is intended to allow construction
-%   of flexible document interface structures in which missing optional
-%   arguments are detected.
+%   Tests if the \meta{token list} matches the special marker token
+%   \cs{NoValue}. This function is intended to allow construction of
```
FrankMittelbach: ... (or its exansion).

I know you say that implicitly below but still.

l3kernel/doc/l3obsolete.txt
```diff
@@ -14,6 +14,7 @@ These functions are deprecated and should be removed from packages.
 ---------------------------------------------------
 Function                            Date deprecated
 ---------------------------------------------------
+\c_novalue_tl                            2025-10-14
```
FrankMittelbach, line 17: just wondering: why don't we suggest what should be done instead of the deprecated command?

l3kernel/CHANGELOG.md
josephwright: I'm not sure if that makes things clearer - perhaps '(mechanism adjusted)' or something like that

l3kernel/l3tl.dtx
josephwright: Well the idea is that the expansion is now fallback behaviour: you'll notice I've made `\NoValue` protected

l3kernel/doc/l3obsolete.txt
josephwright: We normally put something in the deprecation source, but for `tl` it's always hard as you can't really issue a warning then continue. So this one has to be covered somewhere in the docs if we want to - but I think that's better in `ltcmd`/`usrguide`/`clsguide`.

l3kernel/l3tl.dtx
FrankMittelbach: how much does that help against "o" expansion or \expandafter?

l3kernel/doc/l3obsolete.txt
FrankMittelbach: it could have been on line 17 in a third column (but if that is never done in that file it isn't useful, I guess).

l3kernel/CHANGELOG.md
FrankMittelbach: whatever. I look at it from the point of some developer who had used \c_novalue_tl and always explicitly passed on its expansion into the stream (that always the case in all code pieces I saw) and that developer will understand what is meant by the "without any need to "o"-expand). But it is a small world they will all learn :-)

josephwright: I was not 100% sure how to handle `\c_novalue_tl` at all: we could just leave it alone and all would work. I was thinking that if you are going to have `\NoValue` anyway, then using it at one remove via `\c_novalue_tl` feels a bit odd.

l3kernel/l3tl.dtx
josephwright: Well it doesn't, but in general I'd not expect people to be `o`-type expanding arbitrary input - you can't know up-front if the grabbed argument has any particular form.

l3kernel/doc/l3obsolete.txt
josephwright: I'll find somewhere to note this

l3kernel/l3tl.dtx
FrankMittelbach: not on arbitrary input agreed. But a lot of the usage of \c_novalue_tl right now is something like `angle .initial:V         = \c_novalue_tl,` and similar. So of people replace that with \NoValue (without also removing the `:V` or `:o` currently used ) expansion happens

l3kernel/l3tl.dtx
```diff
@@ -2827,33 +2837,49 @@
...
+%   \texttt{-NoValue-} marker. The two are done in the same way, with an
+%   auxiliary needed to avoid \TeX{} \tn{if} matching issues. In both
+%   cases, the first argument of \cs[no-index]{@@_if_novalue_aux$n$:w}
+%   is empty if and only if |#1| starts with the marker, while the second
+%   argument is empty if |#1| is exactly the marker or if it has a question
+%   mark just following the marker. In this second case, however, the material
+%   after the  first |?!| remains and makes the emptiness test return
 %   \texttt{false}.
 %    \begin{macrocode}
+\prg_new_conditional:Npnn \tl_if_novalue:n #1
```
Skillmon, line 2843: Slightly faster implementation (with the drawback of slightly weirder code):

```latex
\cs_set_protected:Npn \@@_tmp:w #1
  {
    \prg_new_conditional:Npnn \tl_if_novalue:n ##1
      { p , T ,  F , TF }
      {
        \__tl_if_empty_if:o
          { \@@_if_novalue_auxi:w {} ##1 {} ? ! \NoValue ? ? ! }
          \use_none:nnn
        \else:
          \@@_if_novalue_auxii:n {##1}
          \prg_return_true:
        \else:
          \prg_return_false:
        \fi:
      }
    \cs_new:Npn \@@_if_novalue_auxii:n ##1
      {
        \fi:
        \__tl_if_empty_if:o
          { \@@_if_novalue_auxiii:w {} ##1 {} ? ! #1 ? ? ! }
      }
    \cs_new:Npn \@@_if_novalue_auxiii:w ##1 #1 ##2 ? ##3 ? ! { ##1 ##2 }
  }
\exp_args:No \@@_tmp:w { \NoValue }
\cs_new:Npn \@@_if_novalue_auxi:w #1 \NoValue #2 ? #3 ? ! { #1 #2 }
```

Skillmon: Alternatively, this code is slower than the one posted above *and* the one from this PR for the case that `#1` is `\NoValue`, but faster in every other case compared to both of them, and is cleaner than both, imho.

```latex
\cs_set_protected:Npn \@@_tmp:w #1
  {
    \prg_new_conditional:Npnn \tl_if_novalue:n ##1
      { p , T ,  F , TF }
      {
        \__tl_if_empty_if:o
          { \@@_if_novalue_auxi:w {} ##1 {} ? ! \NoValue ? ? ! }
          \use_none:nnn
        \fi:
        \__tl_if_empty_if:o
          { \@@_if_novalue_auxiii:w {} ##1 {} ? ! #1 ? ? ! }
          \prg_return_true:
        \else:
          \prg_return_false:
        \fi:
      }
    \cs_new:Npn \@@_if_novalue_auxii:n ##1
      {
        \__tl_if_empty_if:o
          { \@@_if_novalue_auxiii:w {} ##1 {} ? ! #1 ? ? ! }
      }
    \cs_new:Npn \@@_if_novalue_auxiii:w ##1 #1 ##2 ? ##3 ? ! { ##1 ##2 }
  }
\exp_args:No \@@_tmp:w { \NoValue }
\cs_new:Npn \@@_if_novalue_auxi:w #1 \NoValue #2 ? #3 ? ! { #1 #2 }
```.

File: l3kernel/CHANGELOG.md (hunk 1 above)
Comment (Skillmon) [reply l3kernel/CHANGELOG.md:15]: Maybe: "Deprecate `\c_novalue_tl` in favor of `\NoValue` or its expansion"

**Comment by Skillmon** [comment]
I thought the general idea was to use something as marker that can't be input without taking extra steps. `\NoValue` is rather easy to input, and defeats this purpose. Yes, its name is indication enough to not input this light heartedly, and the fact that this can be input without jumping through hoops is intriguing in itself (to purposefully do something like `\otherPkgCommand[\NoValue]` to emphasise that we *really* don't want to use that value), but still I wonder whether this is a good idea.

File: l3kernel/l3tl.dtx (hunk 4 above)
Comment (josephwright) [reply l3kernel/l3tl.dtx:2843]: Thinking again, I wonder if we want to support the expansion of `\NoValue`. My original thinking was that for the transition we need to cover both `\NoValue` and `-NoValue-`. However, due to the way `ltcmd` sets things up, we don't need to set up to support both old and new markers. So the only question is whether we are worried about `\exp_after:wN \some-fn \NoValue`. Really any well-behaved code should not be doing that - the programmer should be testing using `\tl_if_novalue:n(TF)`/`\IfNoValue(TF)` up-front. So we could just test against `\NoValue` I think: this would mean we have no performance impact. Thoughts?

**Comment by josephwright** [comment]
> I thought the general idea was to use something as marker that can't be input without taking extra steps. `\NoValue` is rather easy to input, and defeats this purpose. Yes, its name is indication enough to not input this light heartedly, and the fact that this can be input without jumping through hoops is intriguing in itself (to purposefully do something like `\otherPkgCommand[\NoValue]` to emphasise that we _really_ don't want to use that value), but still I wonder whether this is a good idea.

You might guess we've been through this :-)  I've made that case but use cases have come up where it is useful to have some way of passing 'no value' without needing to use the expansion of `\c_novalue_tl` (which is awkward at the design level). So I'm simply looking for the best implementation of the decision.

(BTW, if you look back, we _used_ to have `\NoValue` in the version of `xparse` in `xpackages` in days of yore.)

File: l3kernel/l3tl.dtx (hunk 4 above)
Comment (davidcarlisle) [reply l3kernel/l3tl.dtx:2843]: >  Really any well-behaved code should not be doing that - the programmer should be testing using `\tl_if_novalue:n(TF)/\IfNoValue(TF) up-front.

But isn't that the point?, in the template usages it's more natural to pass in the marker directly, exactly as you show by (currently) expanding `\c_novalue_tl` once before the call. Using `\NoValue` avoids the need for expansion, so you can use `\some-fn \NoValue`. 

So we don't need to support `\exp_after:wN \some-fn \NoValue` but we do need to support  `\some-fn \NoValue` and for compatibility `\exp_after:wN \some-fn \c_no_value_tl`

Comment (FrankMittelbach) [reply l3kernel/l3tl.dtx:2843]: I guess Joseph's point is that if `\c_novalue_tl` expands to `\NoValue` then the old behavior is still supported, so people can keep
```
  foo-key:o = \c_novalue_tl    % what they have now
```
or replace it by
```
  foo-key = \NoValue
```
Explicilty expanding \NoValue would then simply not work and pass the expansion as a value. If you want to catch that mistake, you need to test for both \NoValue and -NoValue-.  The advantage of not testing is that it is faster, but since when is Joseph and not you or I worried about speed? :-)

I tend to think we can document that \NoValue should be passed without expansion and be done with it because the old usage of `\c_novalue_tl` is still supported (I'm always for fast as you know)

Comment (davidcarlisle) [reply l3kernel/l3tl.dtx:2843]: I did suggest that at one point:-) (possibly not here) I thought the concern was if someone had used full expansion but if all (or even most) known existing cases just expand c_novalue_tl once then yes I agree this is simpler. actually as NoValue is \protected even e expansion would be OK.

Comment (FrankMittelbach) [reply l3kernel/l3tl.dtx:2843]: the list of all usages of \c_novalue_tl across TeXLive should be checked once more but if I remember correctly it is used only via "V" or "o" expansion and "e" expansion as you say would work too due to the \protected

Comment (josephwright) [reply l3kernel/l3tl.dtx:2843]: @FrankMittelbach Believe it or not, I do worry about performance sometimes :-)

Comment (josephwright) [reply l3kernel/l3tl.dtx:2843]: > the list of all usages of \c_novalue_tl across TeXLive should be checked once more but if I remember correctly it is used only via "V" or "o" expansion and "e" expansion as you say would work too due to the \protected

Looking over TL, I agree: the uses are `V`/`o`/`x` so are all 'safe' with the proposed change without needing to check for the older version too.

Note that originally I was thinking we needed to test both for a transition, but I now realise that `\c_novalue_tl` is always defined in `l3kernel` so `ltcmd` will automatically inherit things. The only place it might show is where someone manually defines `\c_novalue_tl`, but the only case I find of that is where there's an `\@ifundefined` test first, so that would still be safe too.

Comment (FrankMittelbach) [reply l3kernel/l3tl.dtx:2843]: > @FrankMittelbach Believe it or not, I do worry about performance sometimes :-)

you saw the smiley I hope? But I guess it is true to say that David and I are burnt by the machine speed in the early nineties

Comment (josephwright) [reply l3kernel/l3tl.dtx:2843]: > > @FrankMittelbach Believe it or not, I do worry about performance sometimes :-)
> 
> you saw the smiley I hope? But I guess it is true to say that David and I are burnt by the machine speed in the early nineties

Oh sure, I do get it - there's a lot we do that if you want max performance you wouldn't - but then you'd end up with a newer format more akin to OpTeX ('you need to know what you are doing everywhere'). I'm quite happy to look at performance particularly when we move things to more stable status.

File: l3kernel/l3tl.dtx [hunk 5]
```diff
@@ -325,20 +325,39 @@
...
-%   contain the same list of tokens, both in respect of character codes and
-%   category codes. This means that
-%   \cs{exp_args:No} \cs{tl_if_novalue:nTF} \texttt\{ \cs{c_novalue_tl} \texttt\} is 
-%   logically \texttt{true} but \cs{tl_if_novalue:nTF} \texttt\{ \cs{c_novalue_tl} \texttt\}
-%   is logically \texttt{false}.
-%   This function is intended to allow construction
-%   of flexible document interface structures in which missing optional
-%   arguments are detected.
+%   Tests if the \meta{token list} matches the special marker token
+%   \cs{NoValue} (or its expansion). This function is intended to allow
```
Skillmon:
```suggestion
%   \cs{NoValue}. This function is intended to allow
```
That's no longer the case.

l3kernel/l3tl.dtx
```diff
@@ -325,20 +325,39 @@
...
+%   construction of flexible document interface structures in which missing
+%   optional arguments are detected.
+%   \begin{texnote}
+%     The test will work for both the current \cs{NoValue} marker token and
+%     and the previous marker (|-NoValue-|).
+%   \end{texnote}
```
Skillmon:
```suggestion
```

this here as well.

l3kernel/l3tl.dtx
```diff
@@ -325,20 +325,39 @@
...
+%   \begin{verbatim}
+%     \tl_if_eq:NnTF \NoValue { -NoValue- }
+%   \end{verbatim}
+%   is logically \texttt{false}. The \cs{NoValue} marker is intended for
+%   use in creating document-level interfaces, where it serves as an indicator
+%   that an (optional) argument was omitted.
+%   \begin{texnote}
+%     This \emph{function} replaces the previously \emph{tl~var}
+%     \cs{c_novalue_tl}. In contrast to \cs{c_novalue_tl}, it is \emph{not}
+%     necessary to \texttt{o}-type expand \cs{NoValue} for it to be used
```
Skillmon, line 354:
```suggestion
%     \cs{c_novalue_tl}. In contrast to \cs{c_novalue_tl}, it is \emph{not}
%     allowed to \texttt{o}-type expand \cs{NoValue} for it to be used
```

PR #1816, opened 2025-11-06
Cleanup l3prefixes.tex

muzimuzhi:
- Fix use of local ior variable.
- Avoid using expl3 scratch variables.
- Drop loading of expl3 package.


josephwright:
I'll merge but in general for truly 'one off' usage, the general scratch vars should be fine.

PR #1827, opened 2025-12-24
`\tl_set_rescan:Nnn` marker

Skillmon:
This PR uses the `\c_@@_rescan_marker_tl` for the delimiter in the `single`-case with `\scantextokens`, resulting in the function accepting arbitrary input.


josephwright:
ChangeLog entry please

PR #1875, opened 2026-03-27
Enhance docs for `V`/`v` arg specifiers

muzimuzhi:
Before (only for changes in `source3body.tex`)

image

After

image

------

Related:
- https://github.com/Witiko/expltools/issues/189


Witiko:
I appreciate that _explcheck_ can also serve as a kind of survey of how people actually use expl3 in practice, hopefully leading to improvements in both the language and its documentation.

PR #1881, opened 2026-04-14
Support Japanese input for string conversion in upTeX

### LINKED ISSUE #1880: Support for native input in upTeX
As originally raised in https://github.com/latex3/pdfresources/issues/18 and more recently in https://tex.stackexchange.com/questions/761866/str-set-convertnnnn-trouble-in-lualatex, the handling of native input in upTeX is an issue. It uses bytes for the UTF-8 range (so `utf8`) _except_ for Japanese, where single tokens are produced. To fix, we need details of exactly which codepoints are 'Japanese' from the POV of upTeX: see https://github.com/texjporg/ptex-manual/issues/12?reload=1.

josephwright:


josephwright:
Not sure if this goes in the docs or not!

u-fischer (approved):
I do not claim to have tested everything but it clearly improves the situation. The example in the pdfmanagement (updated to the current state) compiles with it.

josephwright:
> I do not claim to have tested everything but it clearly improves the situation. The example in the pdfmanagement (updated to the current state) compiles with it.

Yes, I suspect we may have more to do here to get 'everything' right, but I am minded to move forward in steps :-)

FrankMittelbach:
> Not sure if this goes in the docs or not!

if you mean by this should it have an ltnews entry, I think it perhaps should, stating something like we made first steps to support uptex ...

josephwright:
> > Not sure if this goes in the docs or not!
> 
> if you mean by this should it have an ltnews entry, I think it perhaps should, stating something like we made first steps to support uptex ...

Sure, I'll write that once we agree this is being merged

josephwright:
(I actually meant in the docs for `expl3` to say something like '`utf8` for upTeX is a bit odd', more or less.)

FrankMittelbach:
that too, I'm always for documenting code :-)

l3kernel/l3str-convert.dtx
```diff
@@ -2066,6 +2069,41 @@
 % \end{macro}
 % \end{macro}
 % \end{macro}
+% \begin{macro}
+%   {
+%     \@@_decode_utf_viii_start:N ,
+%     \@@_decode_utf_viii_start_auxi:N
+%     \@@_decode_utf_viii_start_auxii:N
+%   }
+%   For up\TeX{}, the input approach is a more complex as characters in
```
niruvt: a bit* more complex?

josephwright: I think we are OK: it's not a simple 8-bit encoding, so that's definitely more complex

niruvt: I was pointing at the grammatical error. "the input approach **is a more** complex" doesn't sound okay. Doesn't it need a noun there?

niruvt: Or the article "a" could be dropped.

PR #1896, opened 2026-05-14
Apply transparency group via an Xform object

josephwright:
See https://github.com/latex3/graphics-def/issues/47


josephwright:
Ping ...

davidcarlisle (approved):
The tex code looks good (ignoring merge conflict in change log) although I must admit edge case behaviours of transparency with xforms are not something I know in detail.

josephwright:
OK, I'm going to risk this and see what happens

Issue #1913, opened 2026-05-22
`l3draw`: transformation errors with `\draw_transform_shift_invert:`


peaR-red:
The shift component of a general affine transformation is defined as

$$S = \begin{pmatrix} s_x \\\ s_y \end{pmatrix}$$

with the full transform $P' = M \cdot P + S$. The inverse shift is $S' = -M^{-1} \cdot S$, but `\draw_transform_shift_invert:` does $-S$, which is only valid if $M$ is the identity matrix. As a result, shifts are not correctly inverted when rotation and scaling is encoded in the coordinate transformation matrix.
```latex
\documentclass{minimal}
\usepackage{l3draw}
\begin{document}
\ExplSyntaxOn
\draw_begin:
  \draw_transform_rotate:n { 30 }
  \draw_transform_shift:n { 2cm , 0cm }
  \draw_path_moveto:n { 0cm , 0cm }
  \draw_path_lineto:n { 1cm , 0cm }
  \color_select:n { red }
  \draw_path_use_clear:n { stroke }
  \draw_transform_shift_invert:
  \draw_path_moveto:n { 0cm , 0cm }
  \draw_path_lineto:n { 1cm , 0cm }
  \color_select:n { blue }
  \draw_path_use_clear:n { stroke }
\draw_end:
\ExplSyntaxOff
\end{document}
```
I would expect the lines to be colinear, just like in this PGF example:
```latex
\documentclass{minimal}
\usepackage{tikz}

\begin{document}
\begin{tikzpicture}
  \pgftransformrotate{30}
  \pgftransformshift{\pgfpoint{2cm}{0cm}}
  \pgfpathmoveto{\pgfpoint{0cm}{0cm}}
  \color{red}
  \pgfpathlineto{\pgfpoint{1cm}{0cm}}
  \pgfusepath{stroke}
  \pgftransformshift{\pgfpoint{-2cm}{0cm}}
  \pgfpathmoveto{\pgfpoint{0cm}{0cm}}
  \color{blue}
  \pgfpathlineto{\pgfpoint{1cm}{0cm}}
  \pgfusepath{stroke}
\end{tikzpicture}
\end{document}
```
On a related note, the original `\__draw_transform_invert:n` used raw `\fp_set:Nn`, which mutated the matrix element by element, corrupting the subsequent math.

PR #1914, opened 2026-05-22
Correct transformation math in `\draw_transform_shift_invert:`

peaR-red:
Addresses #1913.


josephwright:
Test case?

peaR-red:
> Test case?

Does it need one? I suppose we can change `m3draw002.lvt`.

josephwright:
Code changes normally either show up in an existing `.lvt` or get a new test added: that's a long-standing 2e policy that I try to stick to here and in my own code.

peaR-red:
> Code changes normally either show up in an existing `.lvt` or get a new test added: that's a long-standing 2e policy that I try to stick to here and in my own code.

Done.

PR #1920, opened 2026-05-26
Correct transformation math in `\draw_point_vec_polar:nn(n)`

peaR-red:
## Rationale
To map a polar coordinate $(r, \theta)$ onto a custom vector space defined by basis vectors $\vec{x}$ and $\vec{y}$, the true Cartesian formula is $P = r \cos(\theta)\vec{x} + r \sin(\theta)\vec{y}$. The components should distribute across both vectors. However, `\@@_draw_vector_polar:nnn` entirely drops the cross-components, resulting in wildly corrupted math for any rotated or non-orthogonal basis vectors.

Observe with the following MWE:
```latex
\documentclass{article}
\usepackage{l3draw}
\begin{document}
\ExplSyntaxOn
\draw_xvec:n { 1cm, 1cm }
\draw_yvec:n { -1cm, 1cm }

\tl_show:e { \draw_point_vec_polar:nn { 1 } { 0 } }
\ExplSyntaxOff
\end{document}
```
This should yield (1, 1), but it yields (1, 0) instead.

On the naming side, the internal function is erroneously named `\@@_draw_vec_polar:nnn`, but the intended name seems to be `\@@_point_vec_polar:nnn`.


josephwright:
Should show up in a test file ...

josephwright:
Probably I copied the code from the basic polar version

josephwright:
I'm wondering if the `pgf` result is right ...

l3experimental/l3draw/l3draw-points.dtx
```diff
@@ -1058,16 +1058,19 @@
...
     \draw_point:n
       {
-        cosd(#1) * (#2) * \l_@@_xvec_x_dim ,
-        sind(#1) * (#3) * \l_@@_yvec_y_dim
+            cosd(#1) * (#2) * \l_@@_xvec_x_dim
+          + sind(#1) * (#3) * \l_@@_yvec_x_dim
+        ,
+            cosd(#1) * (#2) * \l_@@_xvec_y_dim
+          + sind(#1) * (#3) * \l_@@_yvec_y_dim
       }
```
josephwright, line 1071: Should this use a second auxiliary to avoid doing the sine/cosine twice?

FrankMittelbach: those operations are probably costly, so probably yes

josephwright: I'll sort post-merge.

PR #1922, opened 2026-05-27
Support comma list input for `\text_declare_<case>_exclusion:n`.

myhsia:
Now one can input the excluded list like `\text_declare_titlecase_exclusion:n { and, for, to, of, }` :-)

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -10,7 +10,7 @@ this project uses date-based 'snapshot' version identifiers.
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


peaR-red:
Testfiles should probably be updated as well.

myhsia:
> Testfiles should probably be updated as well.

Added

josephwright:
We've discussed this internally, and feel that it's not really the right approach: at a code level, you could always use a simple `clist` mapping to achieve this, so the existing interface is already working.

PR #1935, opened 2026-06-27
Add \draw_path_replace_bb:n/rework \draw_path_replace_bb:

### LINKED ISSUE #1933: `l3draw`: `\draw_path_replace_bb:` adds line width
Currently, `\draw_path_replace_bb:` adds some line width to the bounding box. Maybe I am overlooking something, but I don't think that it should do that. 

I understand this command as doing something similar to the Ti*k*Z key [`use as bounding box`](https://tikz.dev/tikz-actions#pgf./tikz/use:as:bounding:box) which, according to the doc, "has the same effect as clipping all subsequent drawing against the current path – without actually doing the clipping, only making PGF treat everything as if it were clipped" and thus, not adding any line width (since `clip` doesn't do that either).

The current definition is:
```tex
\cs_new_protected:Npn \draw_path_replace_bb:
  {
    \@@_path_replace_bb:NnN x { max } +
    \@@_path_replace_bb:NnN y { max } +
    \@@_path_replace_bb:NnN x { min } -
    \@@_path_replace_bb:NnN y { min } -
    \@@_softpath_clear:
    \@@_path_reset_limits:
  }
\cs_new_protected:Npn \@@_path_replace_bb:NnN #1#2#3
  {
    \dim_gset:cn { g_draw_bb_ #1#2 _dim }
      {
          \dim_use:c { g_@@_path_ #1#2 _dim }
        #3 0.5 \g_@@_linewidth_dim
      }
  }
```

I'd suggest to simply change it to:
```tex
\cs_new_protected:Npn \draw_path_replace_bb:
  {
    \@@_path_replace_bb:Nn x { max }
    \@@_path_replace_bb:Nn y { max }
    \@@_path_replace_bb:Nn x { min }
    \@@_path_replace_bb:Nn y { min }
    \@@_softpath_clear:
    \@@_path_reset_limits:
  }
\cs_new_protected:Npn \@@_path_replace_bb:Nn #1#2
  {
    \dim_gset:cn { g_draw_bb_ #1#2 _dim }
      {
          \dim_use:c { g_@@_path_ #1#2 _dim }
      }
  }
```

josephwright:
Closes #1933


josephwright:
I've been reflecting: I think for the present we'll just have the 'no line' version - update coming up.

jasperhabicht:
There is also still the open question how to handle bounding boxes of scopes (PR #1939). These two are really very close I have to admit.

PR #1939, opened 2026-07-01
Add public interface to retrieve scope bbox

jasperhabicht:
Currently, it is only possible for the user to retrieve the bbox of the complete picture. Even if inside a scope, it is only possible to retrieve the bbox of all items that were drawn until this point. It is not possible to retrieve the bbox only for a specific scope.


jasperhabicht:
Failed checks due to newly introduced token list variables which are not yet in the test file.

l3experimental/l3draw/l3draw-scopes.dtx
```diff
@@ -299,6 +331,7 @@
   {
     \@@_backend_scope_begin:
     \group_begin:
+      \__draw_scope_reset_bb:
```
josephwright: `\@@_`...

josephwright:
Conceptual question - as set up, this will track exactly the size of the bb of items directly in the current scope, but not nested ones or those in box constructs inside the current scope. Is that what you intend? If so, this should be documented.

josephwright:
Review stuff: we need a ChangeLog entry, and at least one test.

jasperhabicht:
> Conceptual question - as set up, this will track exactly the size of the bb of items directly in the current scope, but not nested ones or those in box constructs inside the current scope. Is that what you intend? If so, this should be documented.

Wouldn't boxes be tracked if `use`d due to `\@@_path_update_limits:nn` ?

Otherwise: good point! I was thinking of additionally making `\__draw_scope_reset_bb:` public, so the user can decide where exactly they want to reset. Maybe, the user also would need commands to start and stop tracking the local bbox next to the command to start and stop tracking the global bbox, because they might want to exclude some items only from the local bbox, but not from the global one. 

We might need to more or less duplicate the whole bbox construct which makes me wonder whether there is a better or more efficient solution to this.

jasperhabicht:
It would be good to be able to change the local bbox by manupulating the values directly and push this back to the global bbox.

Use case: When applying a pattern (as PDF color) to a stroke, it might be necessary to know the size of the bounding box of the relevant path including the line width before actually saying `\draw_path_use:n { stroke }` (see https://github.com/jasperhabicht/hawkdraw/blob/6c27708b4d0dca45e2d57bcbe980a565c5c27701/hawkdraw-shadings.code.tex#L89 for a reference). 

Maybe a command like `\draw_path_phantom:n` could be an interesting solution. It could receive `fill` or `stroke` as argument. Or maybe just make the relevant internal function public?

PR #1940, opened 2026-07-02
Add "raw" color export mode

### LINKED ISSUE #1934: request for public interface to access the backend color space of a spot color
Continued from https://github.com/lualatex/luamplib/issues/196.

As a maintainer of luamplib package which emits a bunch of pdf/dvi literals including color-related pdf objects, I encountered the need for a public command which allows access to the backend color space of a spot color. 

It would be much appreciated if, in the following example, we can get `/color1` (in pdf mode) or `@pdf.obj22` (in dvi mode) with the public command. Using the existing command `\color_export:nnN`, we cannot get backend color space name, but can only get TeX-side name. 
```latex
\ifnum\outputmode = 0
  \DocumentMetadata{ uncompress, xmp=false, backend=dvipdfmx }
\else
  \DocumentMetadata{ uncompress, xmp=false }
\fi
\documentclass{article}
\usepackage{luamplib}
\ExplSyntaxOn
\color_model_new:nnn { pantone1215 }
  { Separation }
  {
    name = PANTONE~1215~U ,
    alternative-model = cmyk ,
    alternative-values = {0, 0.15, 0.51, 0}
  }
\color_set:nnn{spotA}{pantone1215}{1}
\color_export:nnN{spotA}{backend}\l_tmpa_tl \show\l_tmpa_tl
\ExplSyntaxOff
\begin{document}
\mpfig
fill fullcircle scaled 50 withcolor mplibcolor "spotA";
\endmpfig
\end{document}
```

josephwright:
Also re-arranges how \current@color is set to use
a common mechanism.


josephwright:
Main review question really is the naming: do we keep `backend` as is and add `raw`, or do we make a breaking change, use `backend` here and call the existing `backend` something else (`model+space-sep-values`, perhaps)

l3kernel/testfiles-backend/m3color005.xetex.tlg
```diff
@@ -30,3 +30,22 @@ l. ...  }
...
+> \l_tmpa_tl=0.5 0 0.5 rg 0.5 0 0.5 RG.
+<recently read> }
+l. ...  }
+> \l_tmpa_tl=1.0 0.0 0.0 0.0 k 1.0 0.0 0.0 0.0 K.
+<recently read> }
+l. ...  }
+> \l_tmpa_tl=0.6 g 0.6 G.
+<recently read> }
+l. ...  }
+> \l_tmpa_tl=/color1 cs 1.0 scn /color1 CS 1.0 SCN.
```
dohyunkim, line 48: I'm not sure whether `/color1` here makes sense. `/color1` can be a PDF id name. But imho it cannot have meaning in DVI file if `/color1` is not an entry in page resources. (Of course, as I can handle the string, current result is not so bad for me.)

josephwright: You are looking at the XeTeX output :-) In the dvips test you see
```latex
> \l_tmpa_tl=gray 0.6.
<recently read> }
l. ...  }
> \l_tmpa_tl=separation color1 1.0.
```
which is why the docs say the format is not predictable

l3kernel/testfiles-backend/m3color005.etex-dvisvgm.tlg
```diff
@@ -26,7 +26,26 @@ l. ...  }
...
+> \l_tmpa_tl=rgb 0.5 0 0.5.
+<recently read> }
+l. ...  }
+> \l_tmpa_tl=cmyk 1.0 0.0 0.0 0.0.
+<recently read> }
+l. ...  }
+> \l_tmpa_tl=gray 0.6.
+<recently read> }
+l. ...  }
+> \l_tmpa_tl=.
```
davidcarlisle, line 48: `.` here seems wrong?

josephwright: It's entirely empty - I'm really not sure what the correct result is here

davidcarlisle: so it is sorry I misread the log.

svg will need a css color at that point so empty isn't so good, you could use a fallback rgb color or perhaps better just pass on the name something like `var(--color1)`  then the color is a css variable (custom property)  and anything using the generated css could set that to be whatever they want the color to be.

In theory you could use the svg named color  profile

https://www.w3.org/TR/SVGColorPrimer12/#named

but I'm not sue that actually works in browsers

josephwright: I think this is independent of the current PR, as there's currently no support for spot colours in CSS in the release version: best to log an issue for an enhancement, then I can work out what actually works and extend. Is that OK?

davidcarlisle: yes don't block the PR over it but since the driver is expecting dvips style specials with `<model> argument` syntax (that ends up as `argument set<model>color` in PostScript) then an empty string looks like a syntax error rather than "no color" I should test what dvisvgm actually does I suppose.

davidcarlisle (approved):
as discussed some peculiarities in dvisvgm backend but decide not to block this PR (as they apply generally not just to the new raw output form)

josephwright:
One other thing to consider here I think is what the right results are for backends other than pdfTeX/LuaTeX (where PDF literals are clear). I think for XeTeX and `dvipdfmx` we want PDF literals, but for `dvips` and `dvisvgm` have currently gone for the `dvips` format rather than raw PostScript/SVG. Does that seem reasonable?

josephwright:
> One other thing to consider here I think is what the right results are for backends other than pdfTeX/LuaTeX (where PDF literals are clear). I think for XeTeX and `dvipdfmx` we want PDF literals, but for `dvips` and `dvisvgm` have currently gone for the `dvips` format rather than raw PostScript/SVG. Does that seem reasonable?

I wonder if the 'raw' results should be PostScript for `dvips` and SVG/CSS for `dvisvgm`.

josephwright:
@dohyunkim I'm guessing you are not that fussed what happens in DVI mode :-)

josephwright:
> > One other thing to consider here I think is what the right results are for backends other than pdfTeX/LuaTeX (where PDF literals are clear). I think for XeTeX and `dvipdfmx` we want PDF literals, but for `dvips` and `dvisvgm` have currently gone for the `dvips` format rather than raw PostScript/SVG. Does that seem reasonable?
> 
> I wonder if the 'raw' results should be PostScript for `dvips` and SVG/CSS for `dvisvgm`.

I've thought about this a bit more, and I suspect that `raw` should indeed provide PostScript for `dvips` and CSS for `dvisvgm` - unless there is an objection, I'll change.

dohyunkim:
@josephwright No problem. Thanks for your concern.

Issue #1595, opened 2024-09-26
Feature request: \sys_if_engine_unicode: 


daleif:
Would be nice to have a check for unicode engines, aka can I load `fontspec` or not. Similar to `\iftutex` from `iftex`, though `\sys_if_engine_unicode:` would perhaps be more general?

One can use 
```
\bool_if:nTF { \sys_if_engine_luatex_p: || \sys_if_engine_xetex_p:  } ...
```
 but that seems a bit clunky

josephwright:
A few obvious questions

- What is the feeling about this vs the existing 'lazy' approach?
- If we want to go for it, is the underlying test just 'LuaTeX/XeTeX' or is it something else?
- Do we like the proposed name?

muzimuzhi:
According to the l3kernel README,

> The existence of the primitive `\Umathcode` is used as the marker for Unicode support.

https://github.com/latex3/latex3/blob/e591f0850706acbf06f619b0566b3f9c2333da1e/l3kernel/README.md?plain=1#L66-L75

muzimuzhi:
Can be defined statically (just a draft):
```diff
diff --git a/l3kernel/l3sys.dtx b/l3kernel/l3sys.dtx
index 0fcfa9f9..1144abb9 100644
--- a/l3kernel/l3sys.dtx
+++ b/l3kernel/l3sys.dtx
@@ -510,6 +510,11 @@
     \@@_const:nn { sys_if_engine_ #1 }
       { \str_if_eq_p:Vn \c_sys_engine_str {#1} }
   }
+\@@_const:nn { sys_if_engine_unicode }
+  {
+    \bool_lazy_or:nnTF
+      { \sys_if_engine_luatex_p: } { \sys_if_engine_xetex_p: }
+  }
 %    \end{macrocode}
 % \end{variable}
 % \end{macro}
```

> According to the l3kernel README,
> 
> > The existence of the primitive `\Umathcode` is used as the marker for Unicode support.

BTW It seems it's the LaTeX2e which uses `\Umathcode` as the marker, but the l3kernel always uses
```tex
\bool_lazy_or:nnTF { \sys_if_engine_luatex_p: } { \sys_if_engine_xetex_p: }
```
as the marker. For example in `l3kernel/l3unicode.dtx`:
https://github.com/latex3/latex3/blob/e591f0850706acbf06f619b0566b3f9c2333da1e/l3kernel/l3unicode.dtx#L165-L167

josephwright:
[...]

PR #1944, opened 2026-07-20
Add (basic) HiTeX support

josephwright:
No backend at the moment (looks 'interesting'), but working toward at least getting a buildable format.

### DIFF
--- a/l3kernel/l3names.dtx
+++ b/l3kernel/l3names.dtx
@@ -1261,6 +1261,12 @@
   \@@_primitive:NN \omathcode             \tex_omathcode:D
   \@@_primitive:NN \oradical              \tex_oradical:D
 %    \end{macrocode}
+% Prote primitives (for Hi\TeX{}).
+%    \begin{macrocode}
+  \@@_primitive:NN \expanddepth           \tex_expanddepth:D
+  \@@_primitive:NN \Proterevision         \tex_Proterevision:D
+  \@@_primitive:NN \Proteversion          \tex_Proteversion:D
+%    \end{macrocode}
 % Hi\TeX{} primitives.
 %    \begin{macrocode}
   \@@_primitive:NN \HINTafter             \tex_HINTafter:D
--- a/l3kernel/l3names.dtx
+++ b/l3kernel/l3names.dtx
@@ -1261,6 +1261,26 @@
   \@@_primitive:NN \omathcode             \tex_omathcode:D
   \@@_primitive:NN \oradical              \tex_oradical:D
 %    \end{macrocode}
+% Hi\TeX{} primitives.
+%    \begin{macrocode}
+  \@@_primitive:NN \HINTafter             \tex_HINTafter:D
+  \@@_primitive:NN \HINTbefore            \tex_HINTbefore:D
+  \@@_primitive:NN \HINTcolor             \tex_HINTcolor:D
+  \@@_primitive:NN \HINTdefaultcolor      \tex_HINTdefaultcolor:D
+  \@@_primitive:NN \HINTdefaultlinkcolor  \tex_HINTdefaultlinkcolor:D
+  \@@_primitive:NN \HINTdest              \tex_HINTdest:D
+  \@@_primitive:NN \HINTendcolor          \tex_HINTendcolor:D
+  \@@_primitive:NN \HINTendlink           \tex_HINTendlink:D
+  \@@_primitive:NN \HINTimage             \tex_HINTimage:D
+  \@@_primitive:NN \HINTlinkcolor         \tex_HINTlinkcolor:D
+  \@@_primitive:NN \HINTminorversion      \tex_HINTminorversion:D
+  \@@_primitive:NN \HINToutline           \tex_HINToutline:D
+  \@@_primitive:NN \HINTsetpage           \tex_HINTsetpage:D
+  \@@_primitive:NN \HINTsetstream         \tex_HINTsetstream:D
+  \@@_primitive:NN \HINTstartlink         \tex_HINTstartlink:D
+  \@@_primitive:NN \HINTstream            \tex_HINTstream:D
+  \@@_primitive:NN \HINTversion           \tex_HINTversion:D
+%    \end{macrocode}
 % Newer cross-engine primitives.
 %    \begin{macrocode}
   \@@_primitive:NN \ignoreprimitiveerror  \tex_ignoreprimitiveerror:D
--- a/l3kernel/l3sys.dtx
+++ b/l3kernel/l3sys.dtx
@@ -87,8 +87,9 @@
 %
 % \section{Engine}
 %
-% \begin{function}[added = 2015-09-07, EXP, pTF]
+% \begin{function}[added = 2015-09-07, updated = 2026-07-20, EXP, pTF]
 %   {
+%     \sys_if_engine_hitex: ,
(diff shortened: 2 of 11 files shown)


l3kernel/l3names.dtx
```diff
@@ -1261,6 +1261,24 @@
...
+  \@@_primitive:NN \HINTdefaultcolor       \tex_HINTdefaultcolor:D
+  \@@_primitive:NN \HINTdefaultlinkcolor   \tex_HINTdefaultlinkcolor:D
+  \@@_primitive:NN \HINTdest               \tex_HINTdest:D
+  \@@_primitive:NN \HINTendcolor           \tex_HINTendcolor:D
+  \@@_primitive:NN \HINTendlink            \tex_HINTendlink:D
+  \@@_primitive:NN \HINTimage              \tex_HINTimage:D
+  \@@_primitive:NN \HINTlinkcolor          \tex_HINTlinkcolor:D
+  \@@_primitive:NN \HINTminorversion       \tex_HINTminorversion:D
+  \@@_primitive:NN \HINToutline            \tex_HINToutline:D
+  \@@_primitive:NN \HINTsertpage           \tex_HINTsetpage:D
```
muzimuzhi:
```suggestion
  \@@_primitive:NN \HINTsetpage           \tex_HINTsetpage:D
```

muzimuzhi (changes requested):
Checking `\HINTxxx` commands mentioned in `hitexman.pdf`, are `\HINTstream` and `\HINTsetstream` primitives missing here?

muzimuzhi (approved):

davidcarlisle:
also  `\Proteversion`, `\Proterevision`

l3kernel/testfiles/m3expl001.luatex.tlg
```diff
@@ -10,6 +10,10 @@ Defining \sys_if_output_dvi:T on line ...
 Defining \sys_if_output_dvi:F on line ...
 Defining \sys_if_output_dvi:TF on line ...
 Defining \sys_if_output_dvi_p: on line ...
+Defining \sys_if_output_hnt:T on line ...
+Defining \sys_if_output_hnt:F on line ...
+Defining \sys_if_output_hnt:TF on line ...
+Defining \sys_if_output_hnt_p: on line ...
```
muzimuzhi: Unrelated: as now l3kernel is bundled into the format, if these logging lines are considered side-effect of using `\RequirePackage[log-functions]{expl3}`, then perhaps they can be suppressed by loading backend earlier using `\sys_ensure_backend:`.

For example, `m3expl001.lvt` can be changed to something like

```tex
\documentclass{minimal}
\input{regression-test}

\ExplSyntaxOn
\sys_ensure_backend:
\ExplSyntaxOff

% all this tests is that the package and its sub-components load properly as
% sometimes changes might introduce errors not caught at all.

\START
\RequirePackage[log-functions]{expl3}

\ExplSyntaxOn
\cs_new:Npn \test: {}
\ExplSyntaxOff
\END
```

muzimuzhi:
> also `\Proteversion`, `\Proterevision`

Oh. I also found `\expanddepth`, in `hitexman.pdf`, sec. 3.2 LaTeX and ProTE, where `\Proteversion` and `\Proterevision` are listed.

And, need to update `m3names001.lvt`.

josephwright:
> also `\Proteversion`, `\Proterevision`

Those are no HiTeX-specific. Rather, they are part of the Prote code used to add the pdfTeX primitives without using GPL code: that comes from KerTeX, which doesn't use pdfTeX for license reasons. We could add them, but not really for HiTeX-specifc paths.

davidcarlisle:
true but I don't think l3names is directly for engine specific paths (although of course the names get used in tests later)  it's more just a formal list of the available primitives.

josephwright:
> true but I don't think l3names is directly for engine specific paths (although of course the names get used in tests later) it's more just a formal list of the available primitives.

In the main we've only listed up primitives for engines we support in some way. But I see this one is a bit odd - the primitives are there in HiTeX. So I will add them.

l3kernel/testfiles/m3expl001.luatex.tlg
josephwright: The tests don't really do anything any more, as pre-loading the main payload means they don't catch what they were intended for. So removing them entirely seems the easiest way forward :-)

josephwright:
@davidcarlisle OK now?

josephwright:
In particular for review

- Do we like the OpenType/Unicode split
- Are we happy with the test for HINT output as `hnt`, and related changes to the `dvi`/`pdf` tests?

Assuming we merge here, I will do a second PR for HiTeX at the backend level (which may be a bit sketchy ATM).

davidcarlisle (approved):
looks good now.

I  do wonder if we should ask if there should be  or could be a HiTeXVersion primitive currently there is a ProteVersion (which isn't really useful for any top level tests) and a HINTVersion but that's the version of HINT used for output which is (presumably) not necessarily the same as the version of the executable.

Currently it announces itself as

```
This is HiTeX 2.0, TeX version 3.141592653 (TeX Live 2026)
```

but neither the version string (nor the whole banner) are accessible from within the system

The 2.0 is as, far as I can see, just used there and again in the "created by" output, but not exposed otherwise.

```
@d HITEX_VERSION "2.0"

@d banner "This is HiTeX " HITEX_VERSION ", TeX version 3.141592653 "@|
```


But engine changes are out of scope for this PR, so approving here

josephwright:
> looks good now.

Thanks
 
> I do wonder if we should ask if there should be or could be a HiTeXVersion primitive currently there is a ProteVersion (which isn't really useful for any top level tests) and a HINTVersion but that's the version of HINT used for output which is (presumably) not necessarily the same as the version of the executable.

Indeed I was not that happy about having to use `\HINTversion`, but that's all there is ATM.

l3kernel/l3sys.dtx
```diff
@@ -532,6 +556,18 @@
...
 %
+% \begin{function}[pTF]{\sys_if_engine_unicode:}
+%    \begin{macrocode}
+\@@_const:nn
+  { sys_if_engine_unicode }
+  {
+    \bool_lazy_and_p:nn
+      { \cs_if_exist_p:N \tex_Uchar:D }
+      { ! \cs_if_exist_p:N \tex_kanjiskip:D }
+  }
```
muzimuzhi, line 567: I reviewed https://github.com/latex3/latex3/issues/1595 which requested `\sys_if_engine_unicode:(TF)` but resolved by adding  `\sys_..._opentype:(TF)`. It reminds me that the l3kernel README says

> The existence of the primitive `\Umathcode` is used as the marker for Unicode support.

But here `\Uchar` is used as the marker. Need to unify them.

muzimuzhi (commented):
1. A comment on the choice of `\sys_if_engine_unicode:(TF)` marker: `\Uchar` or `\Umathcode`. See comment below.

2. The `\msg_fatal:nnnnnn` doc may need an update to cover the HINT case. Currently it says

  > No PDF file will be produced in this case (DVI mode runs may produce a truncated DVI file).

  (I searched for `dvi` in `interface3.pdf` and this seems to be the only missing case.)

l3kernel/l3sys.dtx
josephwright: I've updated here

PR #1948, opened 2026-08-03
Merge l3backend into l3kernel

josephwright:
We now almost always update the two together:
the split is simply asking for issues.

I have kept l3backend.ins as this is quite long and clear as a separate file from l3.ins.

There were a fed older entries that were cleaned up in build.lua here.


l3kernel/CHANGELOG-backend-pre-08-2026.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
-All notable changes to the `l3backend` bundle will be documented in this file.
+All notable changes to the `l3backend` bundle were documented in this file;
```
FrankMittelbach, line 3: I always think it is better to name files so that they are next to related files, e.g., I would have called it CHANGELOG-backend-pre....md

just a thought

FrankMittelbach (approved):
assuming it compiles ... :-)

one naming suggestion

davidcarlisle:
I think I'd add `\batchinput{l3backend.ins}` to `l3.ins` that would match what we do elsewhere in in latex2e/base `unpack2e.ins`  inputs the more specific ones

josephwright:
> I think I'd add `\batchinput{l3backend.ins}` to `l3.ins` that would match what we do elsewhere in in latex2e/base `unpack2e.ins` inputs the more specific ones

We adjust the prefix, which means things get 'interesting' - I worry it might lead to fragile file orderijng.

davidcarlisle:
> > I think I'd add `\batchinput{l3backend.ins}` to `l3.ins` that would match what we do elsewhere in in latex2e/base `unpack2e.ins` inputs the more specific ones
> 
> We adjust the prefix, which means things get 'interesting' - I worry it might lead to fragile file orderijng.

can't we just input it at the end so everything from there on has the l3backend prefix?

josephwright:
> > > I think I'd add `\batchinput{l3backend.ins}` to `l3.ins` that would match what we do elsewhere in in latex2e/base `unpack2e.ins` inputs the more specific ones
> > 
> > 
> > We adjust the prefix, which means things get 'interesting' - I worry it might lead to fragile file orderijng.
> 
> can't we just input it at the end so everything from there on has the l3backend prefix?

Both `l3.ins` and `l3backend.ins` set the prefix, and that's not reset by `\BatchInput`. So we can certainly look to rearrange the `.ins` setup, but it's not as easy as adding one line. We'd likely want something like `l3kernel.ins`, `l3kernel-lua.ins` and `l3.ins`, for example.

davidcarlisle:
if  you add `\batchinput{l3backend.ins}` before `\endbatchfile` in `l3.ins` then the backend files are generated with no changes as long as you put `\let\MetaPrefix\DoubleperCent`  either just before that or (possibly better) as the first line of `l3backend.ins`

josephwright:
> if you add `\batchinput{l3backend.ins}` before `\endbatchfile` in `l3.ins` then the backend files are generated with no changes as long as you put `\let\MetaPrefix\DoubleperCent` either just before that or (possibly better) as the first line of `l3backend.ins`

I guess what I'm missing here is how this is clearer than just having two `.ins` files: we end up with a relatively fragile setup. Overall, if we want to avoid running multiple `.ins` files, it's likely better to split what we have into a few and put all of the Lua-extraction in one place.

davidcarlisle:
I think all (most?) of our distributions have a single top level `.ins` and having two makes it more likely things get missed. Not if using `l3build` as that just does whatever is needed  but we don't distribute the build scripts to ctan
so having to install both .ins when in all our other distributions processing multiple .ins is sub-optimal as you'll extract everything twice complicates things I think.  It's not the most important thing but as it's only a couple of lines (or one line if you place it before the Lua block changes the prefix) it's worth doing I think.

FrankMittelbach (approved):
looks okay to me and address David's concerns (and mine). Assuming that it installs and keeps the test suite happy afterwards, I guerss this is fine.

Issue #1943, opened 2026-07-17
Using exponential search in \flag_height:N


tfenise:
`\flag_height:N` could use an [exponential search](https://en.wikipedia.org/wiki/Exponential_search) to accelerate. This would make its time complexity roughly O(log n) instead of O(n).

Skillmon:
I'm on a vacation the next two weeks. I can pick this up afterwards and create a PR if you want me to.

PR #1949, opened 2026-08-05
Implement an exponential search for flag_height:N

Skillmon:
This PR uses an exponential search algorithm as suggested in #1943.

The performance is better for a height of up to 2, turns worse than the original implementation and is then consistently faster than that starting at a height of 23. The following two plots show the benchmark results (linear is the previous implementation, exponential the one introduced in this PR):

performance_1-30

performance_full_loglog

### DIFF
--- a/l3kernel/CHANGELOG.md
+++ b/l3kernel/CHANGELOG.md
@@ -8,6 +8,10 @@ this project uses date-based 'snapshot' version identifiers.
 
 ## [Unreleased]
 
+### Changed
+
+- Use an exponential search for `\flag_height:N` (see issue \#1943)
+
 ## [2026-08-02]
 
 ### Added
--- a/l3kernel/l3flag.dtx
+++ b/l3kernel/l3flag.dtx
@@ -71,9 +71,9 @@
 % error message describing incorrect inputs that were encountered.
 %
 % Flags should not be used without carefully considering the fact that
-% raising a flag takes a time and memory proportional to its height and
-% that the memory cannot be reclaimed even if the flag is cleared.
-% Flags should not be used unless it is unavoidable.
+% raising a flag takes a time logarithmically proportional to its height and
+% memory proportional to its height and that the memory cannot be reclaimed even
+% if the flag is cleared. Flags should not be used unless it is unavoidable.
 %
 % In earlier versions, flags were referenced by an \texttt{n}-type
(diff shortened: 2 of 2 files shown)


Skillmon:
Given that flags aren't meant to be used on huge values I'd argue that the better performance for big numbers isn't worth the complexity and actually advise against merging this. One could get better performance for small numbers by unrolling more of the linear search (currently done up to the value of 2 -- but this just shifts the range in which the old implementation is faster, and the speed up in the lower numbers is bought with considerably longer (though not really more complex) code).

Skillmon:
If the only criteria is speed and code complexity isn't an issue at all I can provide code which is faster at all heights (except maybe 0 and 1).

josephwright:
I'm wondering what to do here: the comments suggest we should just skip this, but it clearly works for heavy usage :-)

Skillmon:
@josephwright the suggested usage of flags is to simply store yes/no information. The documentation is pretty clear that it's not a good idea to use raise them into "heavy usage".

I could implement a hybrid approach:

- check for height 0
- check for height 1
- check for height ≥*n*
    - if true use exponential search
    - if false use current linear implementation

Should give identical performance for 0, minusculy better performance for 1, exponentially better performance for ≥*n*, roughly identical if not very slightly worse performance for all other cases. I can then benchmark both paths and find the optimum *n*.

Skillmon:
I'd have a new implementation ready. Takes roughly 100 lines of undocumented code (plus the binary search implementation already contained in this PR). New benchmarking results:

benchmark_linear_hybrid_0-24

benchmark_linear_hybrid

benchmark_linear_hybrid_logarithmic

Shall I push this?

Skillmon:
The hybrid implementation uses a linear implementation up to a height of 158, and an exponential search up to the height 1073741823 (and starting there just does a binary search between that and `\c_max_int` -- that range is just theoretical though, with normal settings one can't raise a flag beyond roughly 100000 until the hash size, string number or pool size limits kick in).

The linear search with two step sizes (8 for rough steps, then 1 in that range -- that's the saw-like part of the curve) proves to be faster than an (almost) pure exponential search up to a height of roughly 190, so in the entire practical range of flags this is faster. As a result my suggestion would be to simply implement a linear search with two different step sizes and drop the exponential search in favour for a simpler implementation.

Issue #1495, opened 2024-03-02
`l3fp`: Add function to extract items from a tuple


jasperhabicht:
In reference to #1491 I would appreciate a function that allows to extract single items from a tuple. Currently such a function does not exist or at least I don't know of one after having consulted LaTeX3 Interfaces.

One could maybe (ab)use `\clist_item:en`, but then the parentheses around the tuple would be extracted as well and it is surely not the best way to do this. 

I could think of a syntax either as a function, such as `select( ⟨tuple⟩ , ⟨item⟩ )` (probably the favorable approach), or as command, such as `\fp_select_from_tuple:nn`.

jasperhabicht:
I propse the following implementation:

```tex
\scan_new:N \s__fp_tuple_stop

\cs_new:Npn \fp_tuple_item:nn #1#2
  {
    \exp_args:Nee \__fp_tuple_item:nn
      { \int_eval:n {#2} }
      {
        \exp_last_unbraced:Ne \__fp_tuple_item_aux:w
          {
            \__fp_parse:n {#1}
          }
      }
  }

\cs_new:Npn \__fp_tuple_item_aux:w \s__fp_tuple \__fp_tuple_chk:w #1 \__fp_sep:
  {
    #1 \s__fp_tuple_stop
  }

\cs_new:Npn \__fp_tuple_item:nn #1#2
  {
    \int_compare:nNnTF {#1} < 0
      {
        \exp_args:Ne \__fp_tuple_item:nn
          {
            \int_eval:n
              {
                \__fp_array_count:n {#2} + 1 + #1
              }
          }
          {#2}
      }
      {
        \__fp_tuple_item_auxi:nw {#1} #2
      }
  }

\cs_new:Npn \__fp_tuple_item_auxi:nw #1#2
  {
    \int_compare:nNnTF {#1} = 1
      { \__fp_tuple_item_auxii:w #2 }
      { \__fp_tuple_item_auxiii:nw #1#2 }
  }

\cs_new:Npn \__fp_tuple_item_auxii:w #1 \__fp_sep: #2 \s__fp_tuple_stop
  {
    #1 \__fp_sep:
  }

\cs_new:Npn \__fp_tuple_item_auxiii:nw #1#2 \__fp_sep: #3
  {
    \exp_args:Ne \__fp_tuple_item_auxi:nw
      { \int_eval:n { #1 - 1 } } #3
  }
```

I can sumbit a PR if this is okay.

[...]

PR #1951, opened 2026-08-09
Add function to extract item from tuple

### LINKED ISSUE #1491: New l3draw functions to extract x/y/z-coordinate of a point
Discussed in https://github.com/latex3/latex3/issues/1488#issuecomment-1974746683.

`pgf` provides `\pgfextractx` and `\pgfextracty` with syntax `\pgfextractx{<dimension>}{<point>}`, see https://tikz.dev/base-points#sec-101.6.

One possibility: `\draw_point_extract_x:Nn <dim var> { <point> }`.

If needed `\draw_point_extract_last_x:N` which extracts the x-coordinate of last updated point can also be provided. `pgf` provides `\pgfgetlastxy{<macro for x>}{<macro for y>}`.

jasperhabicht:
Intends to fix #1491 and #1495


jasperhabicht:
Unsure whether it is safe to use `\s__fp_stop` here (I think it is) or it would be required to add a new scan mark. Can somebody check this?

jasperhabicht:
This should probably rather be implemented as a fp function such as `item((...),i)` or `select(i,<fp expr>)` (which is more similar to the relevant Lua expression) where the first or last argument is the tuple and the other the index. Ideas for a better syntax are welcome!

If this should be implemented as a function, however, somebody needs to help me understand how the `l3fp` module does this exactly, as I don't fully understand the code of the existing functions. 

This might also need to be put into some other sub-module file. I am unsure where this would fit best.

josephwright:
I suspect that this is just `\fp_item:nn`, as this matches e.g. `\clist_item:nn` (you shouldn't need a tuple to be able to apply the function).

josephwright:
I am wondering it we want `\fp_item:nn` (fast but rather low-level), or some `item` or `select` function inside the `fp` (slower but more general, as it can be used at the document level), or perhaps both. [It seems that this request is going to keep popping up until implemented :-)]

josephwright:
At a technical level for the PR:

- ChangeLog entry
- Tests
- Documentation

jasperhabicht:
@josephwright Thanks for the reply! Yes, I can try to create a `select()` function, but I guess I need some time to understand how to implement such a fp function in general. The code is a bit opaque to me.

Naming the function `\fp_item:nn` is not an issue of course, but other such functions don't leave an internal fp representation in the stream unlike the implementation above.

The above code works like in the following example:

```tex
\fp_eval:n { \fp_tuple_item:nn { 2 * ( 1 , 2 ) } { 2 } }
```

l3kernel/l3fp-convert.dtx
```diff
@@ -599,7 +598,7 @@
   }
 \cs_new:Npn \@@_tuple_item_aux:w \s_@@_tuple \@@_tuple_chk:w #1 \@@_sep:
   {
-    #1 \s_@@_tuple_stop
+    #1 \s_@@_stop
```
jasperhabicht, line 601: `\s__fp_stop` needs to be placed later (namely after the `#2` in line 614), otherwise, negative indices won't work.

jasperhabicht:
I tried the following basic implementation of a `select` operator:

```tex
\makeatletter
\cs_set:Npn \__fp_parse_word_select:N
  { \__fp_parse_function:NNN \__fp_select_o:Nw ? }

\cs_set:Npn \__fp_select_o:Nw ? #1 \__fp_sep: #2 @
  {
    \fp_tuple_item:nn {#2} { #1 \__fp_sep: }
    @ 
  }
\makeatother
```

But this eventually expands to something like:

```tex
\__fp_parse_after:ww \s__fp ... \__fp_sep: @ \exp_end_continue_f:w \__fp_parse_infix_end:N \c__fp_prec_end_int \s__fp_expr_stop \exp_end:
```

Which does not meet the argument definition of `\__fp_parse_after:ww` which expects `#1 @ \__fp_parse_infix_end:N \s__fp_expr_stop #2`. I am unsure where to fix this.

jasperhabicht:
I got a working solution.

Todo: Include checks to ensure that the first item is an integer and the second a tuple and throw an error if not.

---

Checks added:
- Checks that number of arguments is exactly 2
- Checks that first argument is a tuple
- Check that last argument is an integer

l3kernel/l3fp-logic.dtx
```diff
@@ -110,6 +167,24 @@
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}[pTF]{\fp_if_tuple:n}
+%   Check if the result is a tuple.
+%    \begin{macrocode}
+\prg_new_conditional:Npnn \fp_if_tuple:n #1 { TF , T , F , p }
+  {
+    \exp_args:Ne \tl_if_eq:nnTF
```
jasperhabicht: `\tl_if_eq:nn(TF)` is not expandable. Better use `\str_if_eq:nn(TF)` here.

jasperhabicht:
> At a technical level for the PR:
> 
> 
> 
> - ChangeLog entry
> 
> - Tests
> 
> - Documentation

Done.

l3kernel/l3fp-convert.dtx
```diff
@@ -582,6 +582,14 @@
...
 % \end{macro}
 %
+% \begin{macro}[EXP]{\fp_item:nn}
+%   Extracts an item from a tuple, the second argument being the intex.
+%    \begin{macrocode}
+\cs_new:Npn \fp_item:nn #1#2
+  { \fp_to_decimal:n { item ( \@@_parse:n {#1} , \@@_parse:n {#2} ) } }
+%    \end{macrocode}
+% \end{macro}
+%
```
josephwright: I'm wondering what the advantage is here as it uses the fp parser: if it was all internal and avoiding needing to parse `item`, I could see an argument for performance.

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,117 @@
...
+    \exp:w \exp_end_continue_f:w
+    \@@_item:w #1
+  }
+\cs_new:Npn \@@_item:w #1 \@@_sep: #2 @
+  {
+    \tl_if_empty:nTF {#2}
+      {
+        \@@_error_num_args:ffff { item } { 2 } { 2 } { 1 }
+        \exp_after:wN \c_nan_fp
+      } {
```
josephwright: `expl3` style has these on separare lines

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,117 @@
...
+        \@@_item:ww #1 \@@_sep: #2 @
+      }
+  }
+\cs_new:Npn \@@_item:ww #1#2 \@@_sep: #3#4 \@@_sep: #5 @
```
josephwright: I assume you've avoided `\bool_lazy_or:nn(TF)` or similar here for performance reasons: perfectly reasonable, but I think deserves a comment.

l3kernel/l3fp-logic.dtx
```diff
@@ -110,6 +221,24 @@
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}[pTF]{\fp_if_tuple:n}
+%   Check if the result is a tuple.
+%    \begin{macrocode}
+\prg_new_conditional:Npnn \fp_if_tuple:n #1 { TF , T , F , p }
+  {
+    \exp_args:Ne \str_if_eq:nnTF
```
josephwright: We have `\str_if_eq:eeTF` available, and it will work here

l3kernel/l3fp.dtx
```diff
@@ -521,6 +521,14 @@
...
 %
+% \begin{function}[pTF]{\fp_if_tuple:n}
+%   \begin{syntax}
+%     \cs{fp_if_tuple_p:n} \Arg{fp expr}
+%     \cs{fp_if_tuple:nTF} \Arg{fp expr} \Arg{true code} \Arg{false code}
+%   \end{syntax}
+%   Evaluates the \meta{fp expr} and tests whether the result is a tuple.
+% \end{function}
+%
 % \section{Floating point expression loops}
```
josephwright, line 532: What's the case for adding this as a public function?

l3kernel/l3fp-convert.dtx
josephwright: Oh, and it's not covered in the docs!

l3kernel/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ this project uses date-based 'snapshot' version identifiers.
 ### Changed
 
 - Integrate `l3backend` files into `l3kernel` distribution
+- Add `\fp_if_tuple:n(TF)`, `\fp_item:nn` and `item()` to `l3fp`
```
josephwright: Doesn't need 'to `l3fp`' here - you could have '`item()` to fp expressions'

l3kernel/testfiles/m3fp-logic005.lvt
```diff
@@ -0,0 +1,42 @@
...
+\documentclass{minimal}
+\input{regression-test}
+
+\ExplSyntaxOn
+\debug_on:n { check-declarations , deprecation , log-functions }
+\ExplSyntaxOff
+
+\begin{document}
+\START
+\AUTHOR{Joseph Wright, Bruno Le Floch}
```
josephwright: Er, should be you :-)

josephwright (changes requested):
Some style points, some minor doc tweaks, one or two more general queries.

l3kernel/l3fp-convert.dtx
jasperhabicht: Sure, we can also just not add that. It was merely that I started with that function, but now it is only a wrapper, so might as well be removed.

I used the same approach like `\fp_max:nn`. Unsure what the benefit for having these is?

l3kernel/l3fp-logic.dtx
jasperhabicht: I actually just tried to use the same set of functions as the other `l3fp` modules. But if it is reasonable performance-wise I can add a comment here.

l3kernel/l3fp.dtx
jasperhabicht: There is none from my side. But the roadmap states 

> Function to count items in a tuple (and to determine if something is a tuple).

So maybe I add a `count()` function later (an internal function already exists anyways).

But maybe I read that wrong and should leave that to Bruno.

jasperhabicht:
It would probably good to replace `\int_compare:nNn` with something more low-level to improve parsing speed and adhere to the overall style in the module, but I am currently unable to get this working and therefore kept things for the time being.

For < 0 we might be able to just test for the first token of the int value being -. Or we could already check earlier and make use of the sign marker of the internal fp representation.

l3kernel/l3fp.dtx
jasperhabicht: I now removed `\fp_item:nn` but kept `\fp_if_tuple:n(TF)` and used this function for `item()` and `count()` (which I newly added).

jasperhabicht (commented):
@josephwright : Answering some of your input. Most of your comments have been resolved. Thanks!

jasperhabicht:
> Or we could already check earlier and make use of the sign marker of the internal fp representation.

What about this:

```
\cs_new:Npn \@@_item:nnn #1#2#3
  {
    \exp_after:wN \@@_item:Nnw
      \@@_item_aux:w #2 \@@_sep:
      {
        \@@_to_int:w #2 \@@_sep:
      }
      #1 \@@_sep:
  }
\cs_new:Npn \@@_item:Nnw #1#2 \s_@@_tuple \@@_tuple_chk:w #3 \@@_sep:
  {
    \token_if_eq_meaning:NNTF 2 #1
      {
        \@@_item:Nnw
          0
          {
            \@@_int_eval:w
            \@@_array_count:n {#3} + 1 + #2
            \@@_int_eval_end:
          }
          \s_@@_tuple \@@_tuple_chk:w {#3} \@@_sep:
      }
      {
          \@@_item_aux:nw {#2} #3 \s_@@_stop
      }
  }
\cs_new:Npn \@@_item_aux:w \s_@@ \@@_chk:w #1#2#3 \@@_sep:
  {
    #2
  }
```

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,175 @@
...
+      {
+        \@@_error_num_args:ffff { item } { 2 } { 2 } { 1 }
+        \exp_after:wN \c_nan_fp
+      }
+      {
+        \@@_item:ww #1 \@@_sep: #2 @
+      }
+  }
+%% For performance reasons, avoid \bool_lazy_or:nn(TF) or similar
+\cs_new:Npn \@@_item:ww #1 \@@_sep: #2 \@@_sep: #3 @
```
josephwright: Missing `\end{macrocode}`/`\begin{macrocode}` pair

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,175 @@
...
+                    \@@_item:nnn {#1} {#2}
+                  }
+                  {
+                    \@@_item_invalid_operation:nn {#1} {#2}
+                  }
+              }
+          }
+          {
+            \@@_item_invalid_operation:nn {#1} {#2}
+          }
```
josephwright: Rather than nesting, could you use `\bool_lazy_and:nnTF` here?

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,175 @@
...
+          \s_@@_tuple \@@_tuple_chk:w {#3} \@@_sep:
+      }
+      {
+          \@@_item_aux:nw {#2} #3 \s_@@_stop
+      }
+  }
+\cs_new:Npn \@@_item_aux:w \s_@@ \@@_chk:w #1#2#3 \@@_sep:
+  {
+    #2
+  }
```
josephwright: Normally we'd have this as a simple one-line ending `{#2}`

l3kernel/l3fp.dtx
```diff
@@ -1299,6 +1307,24 @@
...
 % \end{function}
 %
+% \begin{function}[tested = m3fp-logic005]{count}
+%   \begin{syntax}
+%     \cs{fp_eval:n} \{ |count(| \meta{fp expr} |)| \}
+%   \end{syntax}
+%   Evaluates \meta{fp expr} as tuple and couunts its items. If there is
+%   more than one operand or the operand is not a tuple,
+%   \enquote{invalid operation} occurs.
+% \end{function}
```
josephwright: Document the outcome for a single entry?

l3kernel/l3fp.dtx
```diff
@@ -1299,6 +1307,24 @@
...
+%
+% \begin{function}[tested = m3fp-logic005]{item}
+%   \begin{syntax}
+%     \cs{fp_eval:n} \{ |item(| \meta{fp expr_1} |,| \meta{fp expr_2} |)| \}
+%   \end{syntax}
+%   Evaluates $\meta{fp expr_1}=t$ and $\meta{fp expr_2}=i$ and extracts
+%   item $i$ from tuple $t$.  If the first operand is not a tuple or the 
+%   second operand is not an integer, \enquote{invalid operation} occurs.
+% \end{function}
```
josephwright, line 1332: I wonder what the expected result is for a simple scalar: one could argue `item(1,1)` should work.

josephwright (changes requested):
Minor style things plus a couple of new queries :-)

l3kernel/l3fp.dtx
jasperhabicht: This is in fact something I am still thinking about: should a non-tuple yield 1 with `count()` and itself with `item(..., 1)`? Probably yes ...

l3kernel/l3fp.dtx
josephwright: I think it likely should; if you look at other `item` functions, they don't require the list is long enough, only that it's defined (which here means there is an argument at all).

jasperhabicht: Right, this should be added. I also found that I probably should take care of indices larger than the tuple length and raise an error.

josephwright: Other `item` functions simply return nothing here - also, I guess we should deal with negative indicies (which go from the end of the tuple)

jasperhabicht: > also, I guess we should deal with negative indicies (which go from the end of the tuple)

This is already covered. =)

josephwright: > Other `item` functions simply return nothing here - also, I guess we should deal with negative indicies (which go from the end of the tuple)

May not here, as further operations would fail anyway

l3kernel/l3fp-logic.dtx
jasperhabicht: I might use `\bool_lazy_and:nnTF` and then remove the comment.

jasperhabicht:
I added checks for wrong indices and also updated the existing checks. ~I removed some of the error checks from the test file for the time being as I am unsure how to properly do the error handing.~

~⚠️ The current implementation uses raised flags, but this leads to errors if two commands are called after another and both raise flags. So, another approach to error handling is required.~

l3kernel/l3fp.dtx
jasperhabicht: I decided to throw an error for all "weird" cases:
- operator is not two
- first operator is not a tuple (but it can be an expression resulting in a tuple)
- second operator is not an int (and not inf or nan) (but it can be an expression resulting in an int)
- second operator is zero
- second operator is an int that is larger than the tuple count
- second operator is an int that is smaller than the negative tuple count

jasperhabicht:
Some more notes: 
- It turns out that there is actually already an internal check whether something is a tuple, namely if `\__fp_kind:w` returns 4. I used this for the public function (but now the question even more arises whether to keep this function).
- There is an internal function to test whether something in an integer, namely `\__fp_int:w(TF)` which I used instead of `\__fp_small_int:w(TF)` because the latter also returns true for `inf`.

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,190 @@
...
+  {
+    \exp_after:wN \@@_exp_after_o:w
+    #1 \@@_sep:
+  }
+\cs_new:Npn \@@_item_auxiii:nw #1#2 \@@_sep:
+  {
+    \exp_after:wN \@@_item_auxi:w
+      \int_value:w \@@_int_eval:w #1 - 1 \@@_int_eval_end: \@@_sep:
+  }
+\cs_new:Npn \@@_item_invalid_operation:nn #1#2
```
jasperhabicht, line 278: Maybe this and the following should go to `l3fp-traps`?

jasperhabicht (commented):
Some code improvements and better (but still seemingly buggy) error handling.

l3kernel/l3fp.dtx
jasperhabicht: > > Other `item` functions simply return nothing here - also, I guess we should deal with negative indicies (which go from the end of the tuple)
> 
> May not here, as further operations would fail anyway

Not necessarily. Currently, if the function errs, it returns `nan` which can be used in other functions.

jasperhabicht:
I replaced the raised-flag checks (which disable the use of two such expressions one after the other as we can't use groups for scoping) by a technique where I linearly do cheks using `\use_none_delimit_by_q_stop:w` to prevent `\fi` gobbling due to nesting etc.

jasperhabicht:
The implementation needs to be checked regarding use in a symbolic context. Currently, while this works

```tex
\fp_new_variable:n { A }
\fp_set:Nn \l_tmpa_fp { item ( A , 1 )  }
\fp_set_variable:nn { A } { ( 1 , 2 ) }
\fp_show:n { \l_tmpa_fp }
```

this does not

```tex
\fp_new_variable:n { A }
\fp_set:Nn \l_tmpa_fp { item ( ( A , 2 ) , 1 )  }
\fp_set_variable:nn { A } { 1 }
\fp_show:n { \l_tmpa_fp }
```

___

Works now, tests added.

jasperhabicht:
One nice little touch would be not to use `\__fp_parse:n` to convert the integer resulting from counting the tuple to an internal fp expression. 

But I did not find a better suited command and did not want to create one on my own. I wonder whether there is a simpler solution to this. It would then probably possible to remove the expansion control commands from the function.

---

Done using existing functions such as the internal packing algorithm to directly construct the internal fp representation.

josephwright (approved):

josephwright:
Looks good to me, but I'd like to get a second pair of eyes to take a look

jasperhabicht:
> Looks good to me, but I'd like to get a second pair of eyes to take a look

Of course! I'd also like to look somebody else over this!

I think, questions as to whether we want `item(2,1)` or even `item(((1,2),(3,4)),(1,2))` to work or not or maybe also allow `count(1,2,3)` (apart from `count((1,2,3))`) could be addressed in a second step. This really is meant to implement a basic functionality.

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,232 @@
...
+      }
+  }
+\cs_new:Npn \@@_count_aux:w \s_@@_tuple \@@_tuple_chk:w #1 \@@_sep:
+  {
+    \exp_after:wN \@@_count_int_pack:w
+      \exp:w \exp_end_continue_f:w
+      \@@_array_count:n {#1} \@@_sep:
+  }
+%    \end{macrocode}
+%  The following code converts the integer returned by \__fp_array_count:n
```
josephwright:
```suggestion
%  The following code converts the integer returned by \cs{@@_array_count:n}
```

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,232 @@
...
+        \if:w 4 \@@_kind:w #2 \@@_sep:
+          \@@_item_invalid_operation:nn {#1} {#2}
+          \use_none_delimit_by_s_stop:w
+        \fi:
+        \exp_after:wN \if_meaning:w
+        \int_value:w \@@_int_eval:w \@@_int_p:w #2 \@@_sep: \@@_int_eval_end: 0
+          \@@_item_invalid_operation:nn {#1} {#2}
+          \use_none_delimit_by_s_stop:w
+        \fi:
+        \if_true:
```
josephwright: I think we need a comment on the requirement for this

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,232 @@
...
+        #2
+        \@@_int_eval_end: \@@_sep:
+        #3 \s_@@_stop \s_stop
+    \fi:
+  }
+\cs_new:Npn \@@_item_head:Nw #1#2 \@@_sep: {#1}
+\cs_new:Npn \@@_item_auxi:w #1 \@@_sep: #2 \s_stop
+  {
+%    \end{macrocode}
+%  Return the relevant item if the current index is 1,
```
josephwright:
```suggestion
%  Return the relevant item if the current index is~1,
```

l3kernel/l3fp-logic.dtx
```diff
@@ -72,6 +72,232 @@
...
+\cs_new:Npn \@@_item_compare_one:w
+    #1 \s_@@ 1 \s_@@_mark #2 \s_@@_stop #3 \@@_sep: {#2}
+\cs_new:Npn \@@_item_auxii:w #1 \@@_sep: #2 \s_@@_stop
+  {
+    \@@_if_has_symbolic:nTF {#1} {
+        \@@_exp_after_symbolic_f:nw { \exp_after:wN \exp_stop_f: }
+    } {
+        \@@_exp_after_o:w
+    } 
+    #1 \@@_sep:
```
josephwright: Braces not formatted in our usual way

josephwright:
People seem to be busy - so I read over again and have added a few additional comments.

l3kernel/l3fp-logic.dtx
jasperhabicht: Do you mean the requirement for `\use_none_delimit_by_s_stop:w` or for `\@@_item_invalid_operation:nn`? If the former: maybe, after I rewrote the code to make it work for symbolic use, it might be working as well with a "traditional" if-else construct. Before, I often got errors due to ungobbled `\else` or `\fi`, which I think is somehow related to the `f` type expansion (I might be wrong). Thus, I came up with this mechanism. I will re-check and add a comment if applicable.

josephwright: I mean `\if_true: ... \fi:` rather than just using the function directly.

jasperhabicht: Hm, I thought that this would not work because there would then be an open `\if`, but it seems it does work. I am unsure whether `\use_none_delimit_by_s_stop:w` really gobbles the following `\fi` in this case. But if it does, I'd just remove the `\if_true:` and `\fi:` here (and in the other similar case at line 230).

jasperhabicht: I got it solved (thanks to David) by adding `\exp_after:wN` before `\use_none_delimit_by_s_stop:w`.

PR #1958, opened 2026-09-05
Fix function variant names in expl3.dtx

msyam:
The example describes arguments 2 and 4 as constructing control sequences using \csname, which requires c variants. Since c is a variant of an N argument, the base function signature and the generated variant have been updated accordingly.

### DIFF
--- a/l3kernel/expl3.dtx
+++ b/l3kernel/expl3.dtx
@@ -800,13 +800,13 @@
 % \end{verbatim}
 %
 % Another example: you may wish to declare a function
-% |\demo_cmd_b:ecece|, a variant of an existing function
-% |\demo_cmd_b:nNnNn|, that fully
-% expands arguments 1,~3 and~5, and produces commands to pass as
-% arguments 2 and~4 using~\tn{csname}.
+% |\demo_cmd_b:cnne|, a variant of an existing function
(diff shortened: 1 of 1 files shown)


muzimuzhi:
That signature was once changed from `ecece` to the current `enene` in 67621f42 ("enene" better than "ecece" as a variant, 2023-09-26) as part of
- https://github.com/latex3/latex3/pull/1271

Perhaps we want to adjust the description text to sync with `enene`, or use a better signature example, like `Nnnn` -> `cnne` or so.

@josephwright what's your thought?

FrankMittelbach:
> That signature was once changed from `ecece` to the current `enene` in [67621f4](https://github.com/latex3/latex3/commit/67621f42a33efad32c74cfaed94105962bb29fe3) ("enene" better than "ecece" as a variant, 2023-09-26) as part of
> 
>     * [Pivot toward e-type expansion #1271](https://github.com/latex3/latex3/pull/1271)
> 
> 
> Perhaps we want to adjust the description text to sync with `enene`, or use a better signature example, like `Nnnn` -> `cnne` or so.
> 
> @josephwright what's your thought?

I agree that the example seems rather odd like that I think `Nnnn -> cnne` is much better

msyam:
Updated the description as requested. Please take another look. Thanks!

PR #1962, opened 2026-09-11
docs: fix l3expan documentation

msyam:
## Summary

This PR fixes several documentation issues in `l3expan.dtx`.

### Changes

* ~~Correct the variant-generation example in the documentation.~~
* Correct the list of variable types supported by `V`-type expansion.
* Update argument names in several examples to match the documented argument structure.
* Clarify the documented argument structure for `\exp_args:N..x` functions.

All changes are documentation-only and do not modify the implementation.

### DIFF
--- a/l3kernel/l3expan.dtx
+++ b/l3kernel/l3expan.dtx
@@ -557,7 +557,7 @@
 %     \exp_args:Neee
 %   }
 %   \begin{syntax}
-%     \cs{exp_args:NNno} \meta{token_1} \meta{token_2} \Arg{token_3} \Arg{tokens}
+%     \cs{exp_args:NNno} \meta{token_1} \meta{token_2} \Arg{tokens_1} \Arg{tokens_2}
 %   \end{syntax}
 %   These functions absorb four arguments and expand the second, third
 %   and fourth as detailed by their argument specifier. The first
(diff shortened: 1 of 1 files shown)


msyam:
Could you help to clarify what is optimized in the “optimized functions” described in “Manipulating two/three arguments”?

I noticed that older versions of the documentation included the statement “These functions need slower processing.” for the other group of functions, but this statement has been removed in the current version.

Was this removal intentional?

msyam:
I have another question about the `\exp_last_unbraced:` variants.

`Nno` is listed among the functions that need slower processing, but `Nco` is not. Is this intentional? Is there a difference in their implementation or performance that explains why `Nco` does not require the slower processing?

l3kernel/l3expan.dtx
```diff
@@ -140,7 +140,7 @@
 %   to the \meta{parent control sequence}. So for example
 %   \begin{verbatim}
 %     \cs_set:Npn \foo:Nn #1#2 { code here }
-%     \cs_generate_variant:Nn \foo:Nn { c }
+%     \cs_generate_variant:Nn \foo:Nn { cn }
```
josephwright: This is wrong - only the changed arguments should be given.

josephwright (changes requested):
One place needs changing

l3kernel/l3expan.dtx
FrankMittelbach: or more exactly all up to the last changed one, because in case of, say,  `:NNN` -> `NcN` one would need to specify `Nc`.

msyam: Okay, I understand now. I had been under the impression that the full signature was required when defining a variant.

So, in a sense, it is essentially equivalent to:

```tex
\cs_new:Npn
  \foo:<changed arg-spec><unchanged arg-spec>
  { \exp_args:N<changed arg-spec> \foo:<original arg-spec> }
```
That is, the variant itself does not absorb any arguments (hence the use of \cs_new:Npn). Instead, \exp_args:N... absorbs and expands some of the arguments first, after which the base function absorbs all of the arguments.

msyam:
I have reverted the incorrect change. Could you please review it again?

l3kernel/l3expan.dtx
FrankMittelbach: yes that's about. Requiring the full signature might have looked a little nicer but then one would have to take it apart and use only some of it to keep the functions as performat as possible and even then it would generate additional (unnecessary) work  during the definition, so the current syntax is essentially guided by performance needs.

